PR #26, opened 2016-10-18
simplify the batch index logic

gongzhitaao:
Not a big deal.  But it looks much cleaner.


goodfeli:
I prefer the existing version, because it's more explicit about what is going on, and contains a comment explaining a possibly non-intuitive behavior.

npapernot:
Same here, I think the current version makes it easier to see how batches are indexed.

PR #28, opened 2016-10-27
added saliency map method

rfeinman:
It can compute one adversarial sample at a time. I've also included a new script in tutorials/ showing how to use my code. Right now the algorithm is rather slow for computing adversarial samples, but it seems to get the job done well.

### DIFF
--- a/tutorials/mnist_saliency_tutorial.py
+++ b/tutorials/mnist_saliency_tutorial.py
@@ -8,7 +8,8 @@ from keras.layers import Convolution2D, MaxPooling2D
 import tensorflow as tf
 from tensorflow.python.platform import flags
 
-from cleverhans.utils_mnist import data_mnist
+from cleverhans.utils_mnist import data_mnist, model_mnist
+from cleverhans.utils_keras import accuracy
 from cleverhans.attacks import saliency
 
 FLAGS = flags.FLAGS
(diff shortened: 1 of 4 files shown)


npapernot:
hi @rfeinman, thanks for the PR. @rms5643 is working on merging your implementation with one we have developed in parallel. We will also take care of writing up a tutorial markdown to go with the script. We will make additional commits to your PR's branch and merge it when all of that is ready. Thanks!

tutorials/mnist_saliency_tutorial.py
```diff
@@ -0,0 +1,71 @@
...
+    print "Created TensorFlow session and set Keras backend."
+
+    # Get MNIST test data
+    X_train, Y_train, X_test, Y_test = data_mnist()
+    print "Loaded MNIST test data."
+
+    # build the Keras model, and the
+    x = tf.placeholder(tf.float32, shape=(None, 1, 28, 28))
+    model, f_x = model_mnist(logits=True, input_ph=x)
+    model.compile(loss='categorical_crossentropy',
```
goodfeli: Do we know if this uses `from_logits = True` when computing the cross-entropy? The default when you call the `categorical_crossentropy` directly is `False` and that caused trouble for `fgsm` because it is not very numerically accurate.

rfeinman: It uses the default, which is `from_logits=False`. This loss function is only used during training, though, and should not affect `fgsm` - I noticed that you have written your own `tf_model_loss()` function for `fgsm`. I trained the MNIST model with this method and was able to achieve a 99.2% accuracy on the holdout. Then I applied `fgsm` to the holdout and the model accuracy dropped to ~5%

PR #50, opened 2016-12-06
FGSM code refactoring, label fix

rfeinman:
There are 3 important changes here:

1. Updated code style so that FGSM is consistent with JSMA. Both attacks should return a numpy array with the adversarial sample(s). Previously, FGSM returned a symbolic tensor that needed to be evaluated.
2. Improved general code style; user should not have to define TF placeholders when unnecessary.
3. Fixed FGSM algorithm; the adversary knows the _correct_ label for each sample they are trying to perturb, thus GT labels should be used to compute model loss, not model predictions. In the MNIST case this was a minor discrepancy (model achieves ~99% accuracy, so predictions and GT labels are roughly equivalent), but for other models this is more important.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -17,13 +17,14 @@ from tensorflow.python.platform import flags
 FLAGS = flags.FLAGS
 
 
-def fgsm(x, predictions, eps, back='tf', clip_min=None, clip_max=None):
+def fgsm(sess, model, X, eps, back='tf', clip_min=None, clip_max=None):
     """
     A wrapper for the Fast Gradient Sign Method.
     It calls the right function, depending on the
     user's backend.
-    :param x: the input
-    :param predictions: the model's output
+    :param sess: TF session
+    :param model: The model graph
+    :param X: The input feature matrix
     :param eps: the epsilon (input variation parameter)
     :param back: switch between TensorFlow ('tf') and
                 Theano ('th') implementation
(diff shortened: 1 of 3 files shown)


Hrant-Khachatrian:
importing attacks inside utils_tf does not seem to be a good idea..

rfeinman:
These are just my suggestions, and they don't need to be merged if they don't seem right.

Hrant-Khachatrian:
I think the first point is pretty good, attacks should have similar interfaces...

npapernot:
Hi Reuben and Hrant, thanks for your contributions. I received another proposal (by email) for a unique attack interface. Unfortunately, things are hectic at the moment since I'm at NIPS but I'll get to this soon. Regarding 1., we are also considering a full symbolic implementation of JSMA for efficiency. Regarding 3., the reason behind this is the recent discovery of label leaking. For consistency, I'll make that more clear and perhaps disable that enhancement as the default behavior.

rfeinman:
I just read through the paper about label leaking and I understand the phenomenon now - thanks. I think it's important to have both options implemented, because some people might not be doing adversarial training, but solely using FGSM to craft adversarial samples.

goodfeli:
Yes, attacks probably should have similar interfaces. We're about to make a change soon to accommodate iterative attacks.

goodfeli:
I don't want to do numeric and symbolic versions of every attack. Just symbolic with no `_symbolic` prefix. Some of the other things in this PR make sense but I wouldn't want to merge the part that turns existing symbolic attacks numeric.

rfeinman:
@goodfeli that's fine, I understand your concern. I would just say that if you are hoping to have a similar interface for every attack, and you're planning to keep FGSM symbolic only, you'd need to somehow alter JSMA so that it returns a symbolic tensor and computational efficiency is retained. I authored a portion of the JSMA code and I'm not sure I can see how that would be done at the moment.

npapernot:
I am not sure this PR makes much sense given the recent addition of the FGSM implementation in Theano (see #51), which changed a lot of the code. 

Two comments:

(a) Regarding your point 1., Alex Kurakin made some suggestions on the cleverhans-dev group on how to make the JSMA symbolic, and for a common attack interface. We should follow up on that.

(b) It would still be useful to have two variants of the FGSM with and without the label leaking fix. Maybe as an optional argument to attacks.fgsm 

If there are no objections, I will close this PR. It looks easier to recreate one specifically for (a) and (b).

rfeinman:
Yes, I think we can close this PR and follow up on (a) and (b) separately. I read through Alex's message on the dev group (including his linked Google doc) and, while he mentions that he would help rewrite JSMA, I didn't see any details. But I await eagerly to see what he comes up with.

PR #57, opened 2017-01-15
Python 3 compatibility

### LINKED ISSUE #18: Standardize whitespace
vim is showing me red marks on indentations in `test_mnist_accuracy.py`. We should set up a standard way to configure our editors.

t-vi:
Replace spurious tab by spaces and use range or xrange from six.moves.
In one place, I replaced xrange by range because the output was intended to be a list anyway.

Fixes #18 (but one might encourage to keep using spaces...)

Thank you for developing cleverhans.


goodfeli:
Thanks!

PR #72, opened 2017-01-22
Carlini L2 Algorithm

mahnerak:
Theano implementation of L2 attack from [Carlini et al, 2016]. This attack is slower than FGSM, but produces good-looking examples with high success rate.
Adam optimizer is also added for the algorithm.
See https://arxiv.org/abs/1608.04644 for other details.

### DIFF
--- a/cleverhans/utils_th.py
+++ b/cleverhans/utils_th.py
@@ -62,6 +62,39 @@ def adadelta(loss_or_grads, params, learning_rate=1.0, rho=0.95, epsilon=1e-6):
 
     return updates
 
+def adam(loss_or_grads, params, learning_rate=0.001, beta1=0.9,
+         beta2=0.999, epsilon=1e-8):
+    """ From Lasagne
+    """
+
+    all_grads = get_or_compute_grads(loss_or_grads, params)
+    t_prev = theano.shared(np.array(0., dtype=floatX))
+    updates = OrderedDict()
+
+    # Using theano constant to prevent upcasting of float32
+    one = T.constant(1)
+
+    t = t_prev + 1
+    a_t = learning_rate*T.sqrt(one-beta2**t)/(one-beta1**t)
+
+    for param, g_t in zip(params, all_grads):
+        value = param.get_value(borrow=True)
+        m_prev = theano.shared(np.zeros(value.shape, dtype=value.dtype),
+                               broadcastable=param.broadcastable)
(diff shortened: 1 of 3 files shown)


cleverhans/utils_th.py
```diff
@@ -62,6 +62,39 @@ def adadelta(loss_or_grads, params, learning_rate=1.0, rho=0.95, epsilon=1e-6):
 
     return updates
 
+def adam(loss_or_grads, params, learning_rate=0.001, beta1=0.9,
+         beta2=0.999, epsilon=1e-8):
+    """ From Lasagne
```
goodfeli: This should definitely have a better docstring

goodfeli:
Needs tests

cleverhans/utils_th.py
```diff
@@ -62,6 +62,39 @@ def adadelta(loss_or_grads, params, learning_rate=1.0, rho=0.95, epsilon=1e-6):
 
     return updates
 
+def adam(loss_or_grads, params, learning_rate=0.001, beta1=0.9,
```
goodfeli: Is this literally just a copy-paste from lasagne? If so, it's probably better to just have this piece of code depend on lasagne

goodfeli: is there a reason not to use the keras version of adam?

Hrant-Khachatrian: the reason is to keep the code without external dependencies, as discussed [here](https://groups.google.com/d/msg/cleverhans-dev/BfYcQghdkwY/UWy9DHYuCAAJ). I think we need to explicitly write in the comments that the code is copied from lasagne. Importing lasagne only for the optimizers... i think it's an overkill, especially because [Lasagne docs suggest](http://lasagne.readthedocs.io/en/latest/user/installation.html#stable-lasagne-release) using non-pip version of Theano, which can become confusing.

goodfeli: Your `carlini_l2_step` already imports Keras though, so you do have a Keras dependency either way.

Hrant-Khachatrian: well, i think that one should also be removed, maybe similar to the way [it's done in JSMA](https://github.com/openai/cleverhans/blob/master/cleverhans/attacks_tf.py#L148-L152). Or we can think of a more robust solution

npapernot:
I am closing this for now since we have deprecated Theano and a TF version is underway.

Issue #37, opened 2016-11-22
make the saliency maps more general


npapernot:
right now the map definitions are biased towards image data (because of row x col). We should clean up the code to define them for any input vector. We can do that once the tutorial is fully merged (branch saliency is closed)

rfeinman:
Agreed that this is important. FGSM works for any input shape, but JSMA is only for image data with shape (n_samples, 1, n_rows, n_cols). This is not very well generalized because it only works for black-and-white images. We should update to work with color images and with flat vector data. I can help.

npapernot:
Help would be welcomed yes, feel free to start a PR related to this

npapernot:
Solved by #74

PR #73, opened 2017-01-24
Add tests for pep8 compliance

goodfeli:

PR #74, opened 2017-01-31
Made TF Saliency Map Attack More General

rfeinman:
Addressing issue #37 by making jsma_tf attack generalized to accept data of any shape. We should eventually write a new tutorial for data of some other shape; maybe CIFAR-10. I have tested this on data of different shapes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,20 +5,44 @@
 This repository contains the source code for `cleverhans`, a Python library to
 benchmark machine learning systems' vulnerability to
 [adversarial examples](http://karpathy.github.io/2015/03/30/breaking-convnets/).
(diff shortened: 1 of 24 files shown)


npapernot:
@rfeinman thanks for the PR, but please make sure your code passes PEP8 tests next time (I fixed the errors for this time)

Will run a few more tests before merging

rfeinman:
Apologies - I will follow this new req in the future. The while statement in 246-247 should also be adjusted so that the 'and' comes before the line break.

npapernot:
No worries Reuben, we only recently implemented the pep8 compliance checks in #73

Issue #38, opened 2016-11-22
saliency map attack average perturbation print out


npapernot:
the saliency map attack reports the average distortion for all samples, would be nice to print a different number for samples that were successfully misclassified

PR #83, opened 2017-02-10
added new metric in jsma tutorial

npapernot:
reports average perturbation only for adv examples that were successful


npapernot:
pep8 tests passed manually on my machine

npapernot:
closes #38

Issue #53, opened 2016-12-24
JSMA API suggestion


gongzhitaao:
IMHO, there might be a couple of improvements over the JSMA API.

1. `increase` and `theta` are duplicates.
`increase=True` <-> `theta=+x`
`increase=False` <-> `theta=-x`
2. `jacobian_graph` is like a loose ends to me.
I think the gradients should be calculated internally in the JSMA function, it is, at least to me, counter-intuitive to called some random function before calling the `jsma`.

npapernot:
regardin 2. this is not possible otherwise everytime we make a new call to jsma in order to craft a new adversarial example, we will create new TF graph elements and make the operations slower and slower at each call of jsma

gongzhitaao:
Actually TF provides a [tf.while_loop](https://www.tensorflow.org/api_docs/python/control_flow_ops/control_flow_operations#while_loop) to achieve this.  I have an example implementation of JSMA with tf.while_loop, please see [here](https://github.com/gongzhitaao/tensorflow-adversarial/blob/master/attacks/jsma.py#L59).

npapernot:
This will get addressed by the new API in preparation.

PR #84, opened 2017-02-10
deprecate increase argument in jsma

npapernot:
one argument (increase) can be deduced from the sign of the perturbation introduced to each pixel (theta)


npapernot:
closes #53 

pep8 test passes on my machine locally

npapernot:
@goodfeli not sure this is very clean, I'm OK with another solution if you have a better one

Issue #85, opened 2017-02-10
remove flags from tutorials so they don't raise warnings


npapernot:
following changes made by #79

PR #87, opened 2017-02-18
Set up Travis tests, take two

esigler:
Howdy! I was looking at taking care of #85, and while getting set up noticed there was a dependency needed.  This PR builds on #80, specifically it:

1. Adds a `pycodestyle` dependency to requirements.txt
2. Adjusts the TravisCI config to use Conda for the build environment ([instructions](https://conda.io/docs/travis.html)).  Using apt & building directly, tests sometimes took 10+ minutes, vs ~90 seconds with conda.
3. Adjusts Python versions tested.  See details in this [Travis test run](https://travis-ci.org/esigler/cleverhans/builds/202761441) for which ones work 'out of the box' vs. which would need more digging in.
4. Adds a tests passing badge to the README

Here's an example of TravisCI passing tests with this config: https://travis-ci.org/esigler/cleverhans/builds/202823158

Open questions include:

1. Thoughts on the build matrix?
2. Did you want a pep8 check as part of CI as well?


goodfeli:
Awesome, thanks! No particular thoughts on the build matrix for now. The PEP8 check is already included because that's one of the tests inside `devtools` that `nosetests` is running right now.

PR #93, opened 2017-02-18
shrink mnist accuracy test and make it run on nosetests / cpu

goodfeli:


goodfeli:
OK, I confirm that on cirrascale, I get the same drop in accuracy if I use `CUDA_VISIBLE_DEVICES=-1`. So

1) It seems like tensorflow is using GPU on cirrascale even when `CUDA_VISIBLE_DEVICES` is an empty string
2) It seems like this is likely to be an issue with tensorflow CPU accuracy not matching tensorflow GPU accuracy, rather than anything we do in cleverhans. Somewhat hard to tell if it's clearly a "bug" or just catastrophic changes in compounding rounding error.

PR #97, opened 2017-02-18
adding new norms to FGSM

npapernot:
added L1 and L2 norms for FGSM in TF

credit to @ftramer for this

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -1,7 +1,8 @@
 import numpy as np
 
 
-def fgsm(x, predictions, eps, back='tf', clip_min=None, clip_max=None):
+def fgsm(x, predictions, eps, back='tf', clip_min=None, clip_max=None,
+         ord=np.inf):
     """
     A wrapper for the Fast Gradient Sign Method.
     It calls the right function, depending on the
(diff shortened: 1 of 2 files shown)


cleverhans/attacks.py
```diff
@@ -1,7 +1,8 @@
 import numpy as np
 
 
-def fgsm(x, predictions, eps, back='tf', clip_min=None, clip_max=None):
+def fgsm(x, predictions, eps, back='tf', clip_min=None, clip_max=None,
+         ord=np.inf):
```
goodfeli: should add `ord` to the docstring

goodfeli:
I think it would be good to add a test of each value of `ord`. A simple enough test might pass on travis, since we don't know what it is making the mnist accuracy test get low accuracy.

goodfeli:
I won't insist on changing the name / interface, but have you given some thought as to whether it makes sense to still call this `fgsm` when `ord` is anything other than `np.inf`? The "s" in `fgsm` is for "sign" and we take the sign only if we are using the max norm.

npapernot:
Closing this given that this will be added through the new API

PR #101, opened 2017-02-20
New Feature: Ability to visualize adversarial examples (#6)

### LINKED ISSUE #6: new feature: ability to visualize adversarial examples
also, in particular, confirm that eps of 0.3 in the tutorial results in images that are easy for humans to read. this is higher than i've used before

rms5643:
Adds two visualizations: side-by-side, which shows the original image, the peterbations, then the adversarial example, and grid, which gives a 10x10 grid showing originals down the diagonal (same viz. used in Papernot's DL limitations paper). Also added proper useage of both visualizations in mnist-jsma tutorial. To make it easier for new users, I also added a couple flags in the tutorial to control if visualizations are rendered or not (also, for the best visualization, I made the default number of test set examples to 10, so that the grid is nearly complete. I only say nearly because the first 10 samples are not unique.)

 Committer: Ryan Sheatsley <rms5643@psu.edu>

closes #6


npapernot:
LGTM, closes #6

PR #103, opened 2017-02-21
Fix for adversarial visualizations (and fixed num of adversarial examples to be crafted in print statement)

rms5643:
wrote two conditions to actually check the viz_enabled flag as well as adjusted the first print statement that shows how many adversarial examples will be crafted to the correct value (we don't create an adversarial example for the source class, so just had to subtract 1)

### DIFF
--- a/tutorials/mnist_tutorial_jsma.py
+++ b/tutorials/mnist_tutorial_jsma.py
@@ -191,7 +191,7 @@ def main(argv=None):
     sess.close()
 
     # Finally, block & display a grid of all the adversarial examples
-    if viz_enabled:
+    if FLAGS.viz_enabled:
         _ = grid_visual(grid_viz_data)
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


goodfeli:
I'll leave this for someone who uses JSMA to review

tutorials/mnist_tutorial_jsma.py
```diff
@@ -190,7 +191,8 @@ def main(argv=None):
     sess.close()
 
     # Finally, block & display a grid of all the adversarial examples
-    _ = grid_visual(grid_viz_data)
+    if viz_enabled:
```
npapernot: This looks like a typo, shouldn't it be `FLAGS.viz_enabled`?

npapernot: @rms5643

npapernot:
LGTM

Issue #82, opened 2017-02-10
profile JSMA to identify bottleneck 


npapernot:
tf implementation seems somewhat slower than theano code written 2 years ago

npapernot:
Bottleneck is itertools.combinations so no trivial tweak without making changes to the underlying search heuristic

PR #106, opened 2017-03-08
Compute saliency map an order of magnitude more efficiently.

carlini:
Instead of computing saliency map in python with Pool.map,
use efficient numpy operations. This gives a speedup of ~50x
on MNIST/CIFAR tests using only a single thread.

### DIFF
--- a/cleverhans/attacks_tf.py
+++ b/cleverhans/attacks_tf.py
@@ -84,42 +84,45 @@ def apply_perturbations(i, j, X, increase, theta, clip_min, clip_max):
 
 def saliency_map(grads_target, grads_other, search_domain, increase):
     """
-    TensorFlow implementation for computing salency maps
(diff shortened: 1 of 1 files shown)


carlini:
Not sure why that test failed.

Also, you may want to hold off on merging this in. This commit (and the existing code before it) both assume that the 2-dimensional matrix can fit in memory, which is not the case on ImageNet (299x299x3 Inception would require ~300GB of RAM).

npapernot:
Thanks for the PR, this will help with #82.

Looks like the test failed because PEP8 checks did not pass, can you fix the errors?

carlini:
I'll make the changes early next week (Oakland camera-ready is Monday, hence the PRs).

npapernot:
I'll merge this for now, but feel free to open a second PR with the code that helps deal with large matrices

PR #107, opened 2017-03-09
Fix JSMA bug for odd-length inputs.

carlini:
When running JSMA, fix a bug where if the set of changable features
is odd, then it will crash if no solution is found instead of
returning failure.

This is triggered because JSMA assumes there are at least two pixels
that can be changed at any point in time. If there is only one, things
go bad.


goodfeli:
I think this might not get reviewed until next week. I don't use this particular part of the codebase. @npapernot would be a good reviewer but he is on vacation until Monday.

npapernot:
Thanks for spotting this @carlini, there is indeed a bug in this implementation. The search domain should have at least two pixels left for the while loop to iterate.

Merging now

PR #110, opened 2017-03-23
added mention of label leaking tweak to FGSM docstring

### LINKED ISSUE #104: FGSM loss computation
According to the paper [Adversarial examples in the physical world](https://arxiv.org/abs/1607.02533), FGSM uses `y_true`, the true label, to generate the adversarial images, instead of model predictions.  Besides, it does not really make sense to me to use `y_pred` and logits to compute the loss since `y_pred = softmax(logits)`.  Instead of 

```python
def fgsm(x, predictions, eps, clip_min=None, clip_max=None)
```

The second parameter should be
```python
def fgsm(x, y_true, eps, clip_min=None, clip_max=None)
```

npapernot:
Closes #104

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -8,6 +8,14 @@ def fgsm(x, predictions, eps, back='tf', clip_min=None, clip_max=None):
     user's backend.
     :param x: the input
     :param predictions: the model's output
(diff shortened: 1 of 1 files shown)


npapernot:
@goodfeli can you take a quick look and merge?

PR #114, opened 2017-03-25
Fix #56

### LINKED ISSUE #56: unused learning_phase in batch_eval
`batch_eval` in `utils_th` (https://github.com/openai/cleverhans/blob/master/cleverhans/utils_th.py#L167) throws an error if computation graph doesn't use `learning_phase`. Way to fix that would be to add `on_unused_input='ignore'` to arguments of `theano.function` call.

kudkudak:
This now works https://gist.github.com/kudkudak/aa1eff9f9c0f40b3cfcc945e453d83c4 (note removed Dropout layers in cnn_model)

A bit random note: while doing this, it got me thinking about the way cleverhans is now coupled with keras, wrote few lines about it in https://github.com/openai/cleverhans/issues/109


npapernot:
Thank you for taking care of this!

Issue #109, opened 2017-03-23
Tutorials not working with new keras 2.0


rva5120:
I get this when I run the unmodified version of the tutorial on CPU, with the latest keras.
The error does not show up with keras 1.2

```
(tensorflow) rva5120@cse-p336mcro05:~/tensorflow/cleverhans/tutorials$ python mnist_tutorial_tf.py
Using TensorFlow backend.
W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use SSE3 instructions, but these are available on your machine and could speed up CPU computations.
W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use SSE4.1 instructions, but these are available on your machine and could speed up CPU computations.
W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use SSE4.2 instructions, but these are available on your machine and could speed up CPU computations.
W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use AVX instructions, but these are available on your machine and could speed up CPU computations.
X_train shape: (60000, 28, 28, 1)
60000 train samples
10000 test samples
/home/grads/rva5120/tensorflow/cleverhans/cleverhans/utils.py:143: UserWarning: Update your `Conv2D` call to the Keras 2 API: `Conv2D(64, (8, 8), padding="same", strides=(2, 2))`
  border_mode="same"
/home/grads/rva5120/tensorflow/cleverhans/cleverhans/utils.py:147: UserWarning: Update your `Conv2D` call to the Keras 2 API: `Conv2D(128, (6, 6), padding="valid", strides=(2, 2))`
  border_mode="valid"),
/home/grads/rva5120/tensorflow/cleverhans/cleverhans/utils.py:149: UserWarning: Update your `Conv2D` call to the Keras 2 API: `Conv2D(128, (5, 5), strides=(1, 1))`
  Convolution2D(nb_filters * 2, 5, 5, subsample=(1, 1)),
Defined TensorFlow model graph.
Epoch 0
	Epoch took 52.5618000031 seconds
Traceback (most recent call last):
  File "mnist_tutorial_tf.py", line 122, in <module>
    app.run()
  File "/home/grads/rva5120/tensorflow/local/lib/python2.7/site-packages/tensorflow/python/platform/app.py", line 44, in run
[...]

PR #118, opened 2017-03-29
Bug fixes for Keras version 2.0

rafaelpossas:
- Changes related to the new Convolution Layers in Keras
- Fixed a bug caused by this change on Keras (https://github.com/fchollet/keras/pull/5866) where the outcome of the metrics function is no longer a scalar, therefore the mean should be calculated manually.

### DIFF
--- a/cleverhans/utils.py
+++ b/cleverhans/utils.py
@@ -134,11 +134,11 @@ def conv_2d(filters, kernel_shape, strides, padding):
     :return: the Keras layer
     """
     if LooseVersion(keras.__version__) >= LooseVersion('2.0.0'):
-        return Conv2D(filters=filters, kernel_size=kernel, strides=strides,
-                      padding=padding)
+        return Conv2D(filters=filters, kernel_size=kernel_shape,
+                      strides=strides, padding=padding)
     else:
-        return Convolution2D(filters, kernel[0], kernel[1], subsample=strides,
-                             border_mode=padding)
+        return Convolution2D(filters, kernel_shape[0], kernel_shape[1],
+                             subsample=strides, border_mode=padding)
 
 
 def cnn_model(logits=False, input_ph=None, img_rows=28, img_cols=28,
--- a/cleverhans/utils.py
+++ b/cleverhans/utils.py
@@ -3,15 +3,16 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import os
+from distutils.version import LooseVersion
 import keras
-import matplotlib.pyplot as plt
 from keras.models import Sequential
 from keras.layers import Dense, Dropout, Activation, Flatten
-if int(keras.__version__[0]) >= 2:
+if LooseVersion(keras.__version__[0]) >= LooseVersion('2.0.0'):
     from keras.layers import Conv2D
 else:
     from keras.layers import Convolution2D
+import matplotlib.pyplot as plt
+import os
 
 
 class _ArgsWrapper(object):
@@ -116,6 +117,21 @@ def other_classes(nb_classes, class_ind):
     return other_classes_list
 
 
+def conv_wrap(filters, kernel, strides, padding):
+    """
+    A wrapper that defines the right convolutional layer according to the
+    version of Keras that is installed.
+    :return: the Keras layer
+    """
+    if LooseVersion(keras.__version__[0]) >= LooseVersion('2.0.0'):
+        return Conv2D(filters=filters, kernel_size=kernel, strides=strides,
+                      padding=padding)
+    else:
+        return Convolution2D(filters, kernel[0], kernel[1], subsample=strides,
+                             border_mode=padding)
+
+
+
 def cnn_model(logits=False, input_ph=None, img_rows=28, img_cols=28,
               channels=1, nb_filters=64, nb_classes=10):
     """
@@ -133,7 +149,6 @@ def cnn_model(logits=False, input_ph=None, img_rows=28, img_cols=28,
     :param nb_classes: the number of output classes
     :return:
     """
-    keras_v = int(keras.__version__[0])
     model = Sequential()
 
     # Define the layers successively (convolution layers are version dependent)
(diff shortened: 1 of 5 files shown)


npapernot:
Thank you @rafaelpossas, I will get back to this after #75 is merged. Two things that can be improved:
* can you make sure PEP8 tests pass?
* is the change going to break the keras model definition for clients with keras < 2.0 ?

npapernot:
Addresses #109

rafaelpossas:
Thanks @npapernot.. As you have noticed is not backward compatible with previous versions.. I think it looks good now..

npapernot:
Yes, can you check that the proposed fixed still works on your machine?

rafaelpossas:
Sue thing...

cleverhans/utils.py
```diff
@@ -8,7 +8,10 @@
 import matplotlib.pyplot as plt
 from keras.models import Sequential
 from keras.layers import Dense, Dropout, Activation, Flatten
-from keras.layers import Convolution2D
+if int(keras.__version__[0]) >= 2:
```
goodfeli: This check is brittle.

Should use distutils.version or possibly packaging.version.parse for a more robust check.

http://stackoverflow.com/questions/11887762/compare-version-strings-in-python

cleverhans/utils.py
```diff
@@ -130,29 +133,43 @@ def cnn_model(logits=False, input_ph=None, img_rows=28, img_cols=28,
...
-              Convolution2D(nb_filters * 2, 6, 6, subsample=(2, 2),
-                            border_mode="valid"),
-              Activation('relu'),
-              Convolution2D(nb_filters * 2, 5, 5, subsample=(1, 1)),
-              Activation('relu'),
-              Dropout(0.5),
-              Flatten(),
-              Dense(nb_classes)]
+    layers = [Dropout(0.2, input_shape=input_shape)]
+    if keras_v >= 2:
```
goodfeli: Rather than if statements everywhere, I'd suggest one wrapper class containing the if statement, and then we'd just call the wrapper class everywhere.

cleverhans/utils.py
npapernot: Thanks for the pointer, these packages will be useful yes. See e2cebeb

cleverhans/utils.py
npapernot: I used a simple helper function but happy to make it into a class if you think that is a better approach. See e2cebeb

cleverhans/utils.py
```diff
@@ -3,12 +3,17 @@
...
+from distutils.version import LooseVersion
 import keras
-import matplotlib.pyplot as plt
 from keras.models import Sequential
 from keras.layers import Dense, Dropout, Activation, Flatten
-from keras.layers import Convolution2D
+import matplotlib.pyplot as plt
+import os
+
+if LooseVersion(keras.__version__[0]) >= LooseVersion('2.0.0'):
```
goodfeli: What the `[0]`? Shouldn't it work without that?
The `[0]` is what makes it brittle (imagine Keras version 10)

npapernot: My bad I left it by mistake, I'll fix it

cleverhans/utils.py
```diff
@@ -113,6 +118,20 @@ def other_classes(nb_classes, class_ind):
     return other_classes_list
 
 
+def conv_wrap(filters, kernel, strides, padding):
```
goodfeli: I'd just name it `conv_2d`, the fact that is a wrapper is an implementation detail that could go in the notes of the docstring but doesn't need to be exposed to the user so prominently.

cleverhans/utils_tf.py
```diff
@@ -200,10 +199,12 @@ def model_eval(sess, x, y, model, X_test, Y_test, args=None):
 
             # The last batch may be smaller than all others, so we need to
             # account for variable batch size here
-            accuracy += cur_batch_size * acc_value.eval(
+            cur_acc = acc_value.eval(
                 feed_dict={x: X_test[start:end],
                            y: Y_test[start:end],
                            keras.backend.learning_phase(): 0})
+
+            accuracy += (cur_batch_size * cur_acc).mean()
```
goodfeli: why did we not need a `mean` call before but need it now?

rafaelpossas: @goodfeli 

1.2 Version of keras had mean being called inside categorical_accuracy:

```
def categorical_accuracy(y_true, y_pred):
    '''Calculates the mean accuracy rate across all predictions for
    multiclass classification problems.
    '''
    return K.mean(K.equal(K.argmax(y_true, axis=-1),
                  K.argmax(y_pred, axis=-1)))
```


Which returned a scalar that that was then being aggregated by your code for each batch, before calculating the total avg.

However, 2.0 version removed the Mean calculation from inside the method, and added a K.cast wrapping the call, which results in categorical_accuracy returning an array and, therefore, messing with the calculation when the batch size was smaller than the pre-defined value.
```
def categorical_accuracy(y_true, y_pred):
    return K.cast(K.equal(K.argmax(y_true, axis=-1),
                          K.argmax(y_pred, axis=-1)),
                  K.floatx())
```
Let me know your thoughts on my approach. 
Thanks

cleverhans/utils.py
rafaelpossas: @npapernot After changing the small bug on version check it works on my machine.

cleverhans/utils_tf.py
goodfeli: I think it would make more sense to put a `tf.reduce_mean` at line 179. That way the shape of `acc_value` is the same regardless of keras version, so there is less variability in the behavior of our `cleverhans` code. The way it is right now, we need to keep track of two different possible shapes for two different possible `keras` versions, so it's a little more confusing.

I'd also suggest putting a comment at line 179 explaining why the `reduce_mean` is needed for new keras but not needed for old keras.

rafaelpossas: Thanks! Sounds better..Will do that.

npapernot:
I changed the name of the wrapper to `conv_2d`

Thanks @rafaelpossas, let me know when you pushed the changes

rafaelpossas:
@npapernot @goodfeli 

Changes have been pushed, tested the code on both 1.2 and 2.0 versions of Keras and it works fine.. Thanks for the feedback guys.

cleverhans/utils.py
```diff
@@ -113,6 +118,20 @@ def other_classes(nb_classes, class_ind):
     return other_classes_list
 
 
+def conv_2d(filters, kernel, strides, padding):
```
goodfeli: Maybe `kernel_shape` instead of `kernel` to be clear the user isn't supposed to pass in a 4D tensor of parameters

cleverhans/utils.py
```diff
@@ -113,6 +118,20 @@ def other_classes(nb_classes, class_ind):
...
+    """
+    A wrapper that defines the right convolutional layer according to the
```
goodfeli: Doesn't need to say it's a wrapper

cleverhans/utils.py
```diff
@@ -113,6 +118,20 @@ def other_classes(nb_classes, class_ind):
...
+    version of Keras that is installed.
```
goodfeli: Document the parameters

goodfeli:
The latest commit broke the build because it still uses the old `kernel` parameter name in the function body. After that is fixed and the tests pass, I think this is good to merge.

goodfeli:
Also, should we change setup.py to run keras 2.0 on travis?

goodfeli:
Travis build failed

  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/nose-1.3.7-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/openai/cleverhans/tests_tf/test_mnist_accuracy.py", line 81, in test
    main()
  File "/home/travis/build/openai/cleverhans/tests_tf/test_mnist_accuracy.py", line 61, in main
    model = cnn_model()
  File "/home/travis/build/openai/cleverhans/cleverhans/utils.py", line 181, in cnn_model
    model.add(layer)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/Keras-2.0.2-py2.7.egg/keras/models.py", line 455, in add
    output_tensor = layer(self.outputs[0])
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/Keras-2.0.2-py2.7.egg/keras/engine/topology.py", line 554, in __call__
    output = self.call(inputs, **kwargs)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/Keras-2.0.2-py2.7.egg/keras/layers/convolutional.py", line 164, in call
    dilation_rate=self.dilation_rate)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/Keras-2.0.2-py2.7.egg/keras/backend/tensorflow_backend.py", line 2856, in conv2d
    x = tf.nn.convolution(
AttributeError: 'module' object has no attribute 'convolution'

npapernot:
yeah, it's because tf was too old in the travis config, I updated it now

npapernot:
I also have to fix the keras scalar output issue for theano

PR #119, opened 2017-03-29
adding tests for tutorials

npapernot:
Trying out Travis on the tutorials


cleverhans/utils_mnist.py
```diff
@@ -11,7 +11,7 @@
 from . import utils
 
 
-def data_mnist():
+def data_mnist(testing=False):
```
goodfeli: I'd suggest `train_start=0, train_end=60000, test_start=0, test_end=10000` for max flexibility.

tests_tf/test_mnist_tutorial_tf.py
```diff
@@ -0,0 +1,21 @@
+from inspect import getsourcefile
+import os
+import sys
+
+# Import the tutorial after evaluating its relative path
```
goodfeli: Is the issue here that the tutorials aren't a submodule of `cleverhans` and there is probably something else called `tutorials` on the `PYTHONPATH`? Would it be easier to just rename them `cleverhans_tutorials` so they can be imported without changing the path?

tests_tf/test_mnist_tutorial_tf.py
```diff
@@ -0,0 +1,21 @@
...
+current_dir = os.path.dirname(os.path.abspath(getsourcefile(lambda: 0)))
+current_dir = current_dir[:current_dir.rfind(os.path.sep)]
+tutorial_dir = os.path.join(current_dir, 'tutorials')
+sys.path.insert(0, tutorial_dir)
+import mnist_tutorial_tf
+sys.path.pop(0)
```
goodfeli: if we keep the path modification code, it might be somewhat tricky to make it behave well if something fails. For example, if `import mnist_tutorial_tf` fails, maybe this results in `nosetests` being stuck in the wrong directory. This may be something we should handle with a `try` block and `finally` block. Even then, it could hard to tell whether or not to run the `pop` inside the `finally` block.

tutorials/mnist_tutorial_tf.py
```diff
@@ -24,11 +24,18 @@
 flags.DEFINE_float('learning_rate', 0.1, 'Learning rate for training')
 
 
-def main(argv=None):
+def mnist_tutorial(testing=False):
```
goodfeli: aesthetically, I think it would be nicest not to have `testing` flags, but just have general problem size flags. That way the test can call the tutorial with very small problem sizes, and the user is also free to shrink the problem size if they want it to run faster for whatever reason.

goodfeli: (and for tutorials, it's nice if we hide as much of the internals as possible)

tutorials/mnist_tutorial_tf.py
```diff
@@ -24,11 +24,18 @@
...
     MNIST cleverhans tutorial
+    :param testing: only used internally for cleverhans testing purposes
     :return:
     """
+    # Used for cleverhans testing purposes only. Content returned must be:
+    # [0] accuracy of model on clean data
+    # [1] accuracy of model on adversarial examples
+    # [2] accuracy of adversarially trained model on clean data
+    # [3] accuracy of adversarially trained model on adversarial examples
+    test = np.zeros(4, dtype=np.float32)
```
goodfeli: how about something like

    class AccuracyReport(object):
        """
        An object summarizing the accuracy results for experiments involving training on clean examples or adversarial examples, then evaluating on clean or adversarial examples.
        """
        def __init__(self, clean_train_clean_eval, clean_train_adv_eval, adv_train_clean_eval, adv_train_adv_eval):
           self.__dict__.update(locals())
           del self.self

PR #122, opened 2017-04-19
Jsma fixes

### LINKED ISSUE #121: Occasional `KeyError` with JSMA depending on input
I was using the JSMA attack (with the `increase=False` option) on "tutorial" MNIST classifier from the TensorFlow website.  However, it *occasionally* dies when trying to remove a key that does not exist:
```
Current iteration: 5 - Current Prediction: 1
Current iteration: 10 - Current Prediction: 1
Current iteration: 15 - Current Prediction: 1
Current iteration: 20 - Current Prediction: 1
Current iteration: 25 - Current Prediction: 3
  File "mnist_test_updated.py", line 245, in <module>
    max_distortion/100, False, 0, 1)
  File "/net/af11/apn4za/AdversarialMachineLearning/fjsma/cleverhans/cleverhans/attacks_tf.py", line 246, in jsma_tf
    grads_target, grads_others, search_domain, increase)
  File "/net/af11/apn4za/AdversarialMachineLearning/fjsma/cleverhans/cleverhans/attacks_tf.py", line 124, in saliency_map
    search_domain.remove(p1)
KeyError: 0
```

[Link to offending lines.][badlines]

This occurs when there are *no* pixels that satisfy the saliency map condition of `alpha < 0` and `beta > 0` (using the terms from Papernot, et al), so then "best" is 0.  However, for MNIST, the 0 pixel is *rarely* set, so it is not in the `search_domain` set and the removal fails.

I did a temporary fix on my side by using `discard` instead of `remove`, which fails silently.  The downside of this is that the saliency map will continue to delete the `0, 0` pixels until the `max_iters` value is exceeded in the main `jsma_tf` loop; this wastes resources, but is a simple fix.  

[badlines]:https://github.com/openai/cleverhans/blob/master/cleverhans/attacks_tf.py#L124-L125

npapernot:
Closes #121 

When no pixels satisfy the saliency map, the JSMA will try to `remove` the same pair of pixels multiple times from the `search_domain` list. Switched to `discard` (performance overhead is acceptable given that this case is not common afaik)

Some changes made in #106 created a bug by setting the gradients of pixels outside of the search domain to `0` (the value is not sufficiently extreme in order to guarantee that the pixel will not be selected as part of a pair when computing the saliency score)

### DIFF
--- a/cleverhans/attacks_tf.py
+++ b/cleverhans/attacks_tf.py
@@ -97,8 +97,8 @@ def saliency_map(grads_target, grads_other, search_domain, increase):
 
     # Remove the already-used input features from the search space
     invalid = list(set(range(nf)) - search_domain)
-    grads_target[invalid] = - int(increase) * np.max(grads_target)
-    grads_other[invalid] = - int(increase) * np.min(grads_other)
+    grads_target[invalid] = - (2 * int(increase) - 1) * np.max(grads_target)
+    grads_other[invalid] = - (2 * int(increase) - 1)* np.min(grads_other)
 
     # Create a 2D numpy array of the sum of grads_target and grads_other
     target_sum = grads_target.reshape((1, nf)) + grads_target.reshape((nf, 1))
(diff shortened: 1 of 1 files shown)


npapernot:
Tests won't fail anymore once #75 is merged (`Qt` bug in Travis config will be fixed)

PR #124, opened 2017-04-23
Minor fixes in tutorials

frankyjuang:


npapernot:
Tests are failing because of Qt/Matplotlib but Travis bug is fixed in #75

PR #137, opened 2017-05-11
remove hard dependency on matplotlib in utils.py

### LINKED ISSUE #136: Make matplotlib dependency soft
Currently it's not possible to run mnist_tutorial_tf.py unless matplotlib imports successfully. matplotlib is not used for any feature in the tutorial. matplotlib fails to import if python-tk is absent, so this forces installing python-tk on Cloud machines that otherwise would have no GUI related libraries at all.

npapernot:
Closes #136 

Moves the `matplotlib` import inside the only two functions of `utils.py` that use it


goodfeli:
LGTM, thanks. I'll merge once the tests pass.

PR #138, opened 2017-05-12
model_eval bug fix, wrongly use of tf.argmax

leviswind:
tf.argmax can't assign axis with -1, here is the official tutorial
```
axis: A Tensor. Must be one of the following types: int32, int64. int32, 0 <= axis < rank(input).
 Describes which axis of the input Tensor to reduce across. For vectors, use axis = 0.
```

### DIFF
--- a/cleverhans/utils_tf.py
+++ b/cleverhans/utils_tf.py
@@ -3,6 +3,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+from distutils.version import LooseVersion
 import math
 import numpy as np
 import os
@@ -207,8 +208,12 @@ def model_eval(sess, x, y, model, X_test, Y_test, args=None):
     assert args.batch_size, "Batch size was not given in args dict"
 
     # Define accuracy symbolically
-    correct_preds = tf.equal(tf.argmax(y, axis=tf.rank(y) - 1),
-                             tf.argmax(model, axis=tf.rank(model) - 1))
+    if LooseVersion(tf.__version__) >= LooseVersion('1.0.0'):
+        correct_preds = tf.equal(tf.argmax(y, axis=-1),
+                                 tf.argmax(model, axis=-1))
+    else:
+        correct_preds = tf.equal(tf.argmax(y, axis=tf.rank(y) - 1),
+                                 tf.argmax(model, axis=tf.rank(model) - 1))
     acc_value = tf.reduce_mean(tf.to_float(correct_preds))
 
     # Init result var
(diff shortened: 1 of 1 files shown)


npapernot:
Hi, 

I am not sure if that is true. Here is a PR implementing this functionality in TF: https://github.com/tensorflow/tensorflow/pull/6042

Did you run into any issues running cleverhans?

leviswind:
Hi, 
I updated the version of tensorflow to the latest and restart my computer. The issue still exists, here is the error information
```
ValueError: Dimension (-1) must be in the range [0, 2), where 2 is the number 
of dimensions in the input. for 'ArgMax' (op: 'ArgMax') with input shapes: [?,10], []
```
Can you run the mnist_blackbox.py with the latest version of tensorflow to see if it works? Maybe something is wrong with my version of tensorflow

goodfeli:
The black box tutorial works for me.

60000 train samples
10000 test samples
Preparing the black-box model.
Defined TensorFlow model graph.
Completed model training.
Test accuracy of black-box on legitimate test examples: 0.971212121212
Training the substitute model.
Defined TensorFlow model graph for the substitute.
Substitute training epoch #0
Completed model training.
Augmenting substitute training data.
Labeling substitute training data.
Substitute training epoch #1
Completed model training.
Augmenting substitute training data.
Labeling substitute training data.
Substitute training epoch #2
Completed model training.
Augmenting substitute training data.
Labeling substitute training data.
Substitute training epoch #3
Completed model training.
Augmenting substitute training data.
Labeling substitute training data.
Substitute training epoch #4
Completed model training.
Augmenting substitute training data.
Labeling substitute training data.
Substitute training epoch #5
Completed model training.
Test accuracy of oracle on adversarial examples generated using the substitute: 0.0966666666727

goodfeli:
I'm using tf 1.1.0. What version are you using?

    goodfellow@instance-2:~/cleverhans/tutorials$ python -c "import tensorflow as tf; print tf.__version__"
1.1.0

leviswind:
I'm sorry for my mistake. I followed the tensorflow tutorial to update my tensorflow. However it seems the version of my tensorflow is r0.12. Maybe it's the problem. I'll update it to the latest version to check the result. Thank you for your answer.

leviswind:
Maybe it's the problem. tensorflow/tensorflow#9872

goodfeli:
For the PR, do you want to change it to do a version check? The current code makes a slightly smaller graph and should be better for the latest version of tf, but it could be good to support version 0.12.

This topic is orthogonal to the PR, but at some point we should make a formal policy for which versions of tf we support.

npapernot:
+1 for specifying which versions of tf we support. 

The code currently mentions a potential deprecation for versions that don't support `global_variables_initializer()` here: https://github.com/openai/cleverhans/blob/master/cleverhans/utils_tf.py#L145 

So we may want to support >= 0.12

PR #147, opened 2017-05-27
Refine initialize_uninitialized_global_variables

frankyjuang:
Use built-in function `tf.report_uninitialized_variables` to get uninitialized variables.

### DIFF
--- a/cleverhans/utils_tf.py
+++ b/cleverhans/utils_tf.py
@@ -66,20 +66,10 @@ def initialize_uninitialized_global_variables(sess):
     :param sess: the TensorFlow session
     :return:
     """
-    # List all global variables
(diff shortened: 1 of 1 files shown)


frankyjuang:
Besides, would it be better if we set `init_all=False` of `model_train` as default?

goodfeli:
@frankyjuang Why did you close this? Did you decide it wasn't a good idea after looking at the output of the tests?

npapernot:
The tests were failing because of the `pep8` checker

frankyjuang:
No, I found that it's not a good idea to use `tf.report_uninitialized_variables` since it requires a VariableScope with `reuse` set to `True`.

PR #151, opened 2017-06-05
Pure tensorflow version of tutorial

goodfeli:


cleverhans/utils.py
```diff
@@ -253,3 +165,17 @@ def grid_visual(data):
     # Draw the plot and return
     plt.show()
     return figure
+
+
+def conv_2d(*args, **kwargs):
+    from cleverhans.utils_keras import conv_2d
+    warnings.warn("utils.conv_2d is deprecated and may be removed on or after"
```
npapernot: It is more of a structure question for what should be included in the `cleverhans/` folder, since we will have to keep keras as a dependency for our Travis setup because of the tests related to the tutorials.

I don't have a strong opinion either way.

tutorials/mnist_tutorial_pure_tf.py
```diff
@@ -0,0 +1,297 @@
...
+    self.layers = layers
+    self.input_shape = input_shape
+    for layer in self.layers:
+      layer.set_input_shape(input_shape)
+      input_shape = layer.get_output_shape()
+
+  def get_params(self):
+    out = []
+    for layer in self.layers:
+      out = ordered_union(out, layer.get_params())
```
npapernot: `ordered_union` is not defined anywhere, is it missing an import?

goodfeli: Oh, I meant to remove `get_params` for the tutorial.

cleverhans/utils.py
goodfeli: I don't have a strong opinion either way either.

tutorials/mnist_tutorial_pure_tf.py
npapernot: Sounds good!

cleverhans/utils.py
npapernot: I guess we can keep it for now. To be coherent, we would probably need to remove the MNIST dataset loader if we decided to remove the keras model definition.

PR #154, opened 2017-06-07
A model abstraction that exposes just the methods needed by attacks

fartashf:


fartashf:
I signed it!

cleverhans/model_abs.py
```diff
@@ -0,0 +1,73 @@
+from abc import ABCMeta
+from keras.models import Model
+
+
+class ModelAbstraction:
+    """
+    An abstract interface for a model that exposes the required functionalities
```
npapernot: Could you make this comment a little more explicit (on the "required functionalities")?

cleverhans/model_abs.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        :return: A symbolic representation of the output logits, values before
+        softmax
+        """
+        error = 'get_logits not implemented'
+        raise NotImplementedError(error)
+
+    def get_probs(self):
+        """
+        :return: A symbolic representation of the output logits, values before
```
npapernot: It looks like this comment is not correct

cleverhans/model_abs.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        :param model: A function that takes a symbolic input and returns the
+                      symbolic output for the model's predictions.
+        """
+        raise NotImplementedError('Model abstraction not implementated.')
+
+    def get_hidden(self, layer):
+        """
+        :return: A symbolic representation of the hidden features
+        :param layer: The name of the hidden layer to return features at.
```
npapernot: This requires that all classes that inherit `ModelAbstraction` keep track of the name of layers when defining the model. It would be nice to formalize this in the documentation of `__init__`

npapernot:
Could you add some tests to `test_tf/` to make sure that the right exceptions are thrown and that a dummy Keras model correctly instantiates when using KerasModelWrapper?

cleverhans/model_abs.py
```diff
@@ -0,0 +1,73 @@
+from abc import ABCMeta
+from keras.models import Model
+
+
+class ModelAbstraction:
```
goodfeli: Should inherit from `object`

cleverhans/model_abs.py
```diff
@@ -0,0 +1,73 @@
+from abc import ABCMeta
+from keras.models import Model
```
goodfeli: Shouldn't depend on `keras`

goodfeli: OK, I see why this is needed for `KerasModelWrapper`. I think either `KerasModelWrapper` should go in a separate module that imports `keras`, or `KerasModelWrapper` should import `keras` only when the `KerasModelWrapper` class is constructed. Users shouldn't need to install `keras` to be able to import `ModelAbstraction`.

cleverhans/model_abs.py
goodfeli: I'd say maybe just `Model` to keep it shorter

cleverhans/model_abs.py
npapernot: +1 for avoiding to reintroduce a dependency on Keras

cleverhans/model_abs.py
```diff
@@ -0,0 +1,73 @@
...
+    An abstract interface for a model that exposes the required functionalities
+    """
+    __metaclass__ = ABCMeta
+
+    def __init__(self, model):
+        """
+        :param model: A function that takes a symbolic input and returns the
+                      symbolic output for the model's predictions.
+        """
+        raise NotImplementedError('Model abstraction not implementated.')
```
goodfeli: I think we actually want this to just be `pass`. We probably want subclasses to call this constructor. In the future, we might have the base class do some generic work in the constructor (register the model with some name server, or something like that---just speculating). If we put an exception in the constructor, no one will call it, and we'll have to change all the subclass code if we ever depend on the constructor being called.

cleverhans/model_abs.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def get_hidden(self, layer):
```
goodfeli: Shouldn't the features be associated with a particular input tensor?

I would expect something like `def fprop(self, x)` that returns a dictionary mapping layer names to activations computed on `x`.

cleverhans/model_abs.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        :return: A symbolic representation of the hidden features
+        :param layer: The name of the hidden layer to return features at.
+        """
+        error = 'Feature extraction for hidden layers not implemented'
+        raise NotImplementedError(error)
+
+    def get_logits(self):
```
goodfeli: is there a strong reason to special-case the logits? Why not just `get_hidden('logits')`?

cleverhans/model_abs.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        :return: A symbolic representation of the output logits, values before
+        softmax
+        """
+        error = 'get_logits not implemented'
+        raise NotImplementedError(error)
+
+    def get_probs(self):
```
goodfeli: is there a strong reason to special-case the probs? Why not just `get_hidden('probs')`?

cleverhans/model_abs.py
fartashf: They don't necessarily need to keep track of the name as is the case for `Keras` that provides a specific function. I made that clear in the comments.

cleverhans/model_abs.py
fartashf: I moved the import inside the `get_hidden`, where it is used.

npapernot: Thanks!

cleverhans/model_abs.py
fartashf: Are you suggesting that the methods do not return a symbolic representation but rather compute and return the output given an input?

cleverhans/model_abs.py
fartashf: Here `get_hidden` depends on the model-specific layer names. `get_logits` and `get_probs` are supposed to be independent of the layer names. Possibly, `get_logits` and `get_probs` can use `get_hidden` but after they find the appropriate layer. Your suggestion can be done but it would be by having multiple special cases inside `get_hidden` that would complicate the method. 

`get_logits` and `get_probs` are special cases because they are currently the only methods needed by the core code. Here are the usages:
https://github.com/tensorflow/cleverhans/blob/fa17e16051f60221926b7fb884e352c15e484134/cleverhans/utils_tf.py#L52-L56
https://github.com/tensorflow/cleverhans/blob/fa17e16051f60221926b7fb884e352c15e484134/cleverhans/utils_tf.py#L213-L219

In the future, we might want to have something like `get_embed` for the final feature representation before logits, eg. FC7.

cleverhans/model_abs.py
goodfeli: No. I'm suggesting that we make it possible to build the model's fprop graph on top of more than one symbolic tensor. For adversarial training, we will usually want to have one fprop graph built on a symbolic tensor representing the clean training examples, and another graph built on a symbolic tensor representing the adversarial examples. Does that make sense?

fartashf: I see. Makes sense to pass `x`. The PR is actually returning a new Model for Keras rather than a symbolic representation which is not good.

But as for not passing `layer`, there would be an overhead of creating too many new Model classes to expose the features. Also, as for in-place operations, this would increase the computation time and memory usage.

How about `def fprop(self, x, layer)`?

goodfeli: I think it should be possible to run the model once and get the state of all layers. tf.slim does something similar by returning a dictionary of `endpoints` that are basically what we want.
It may not be possible to expose hidden layers using Keras models. That's one of the reasons we moved away from using Keras as the basis for CleverHans. It's OK if your Keras implementation of the Model interface is only able to support exposing the probabilities. That way Keras models could work with most attacks like FGSM, but to use feature adversaries it'd be necessary to use something like tf.slim, Sonnet, or your own model code like in the pure_tf tutorial.

cleverhans/model_abs.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class Model(object):
+    """
+    An abstract interface for model wrappers that exposes model symbols
+    needed for making an attack. This abstraction removes the dependency on
+    any specific neural network package (e.g. Keras) from the core
+    code of CleverHans. It can also simplify exposing the hidden features of a
+    model when a specific package does not directly expose them (needed by
+    "features adversaries").
```
npapernot: Could you add a pointer to the paper on this topic for users who are not aware of this line of work? I'm assuming you are referring to this: https://arxiv.org/abs/1511.05122

cleverhans/model_abs.py
```diff
@@ -0,0 +1,138 @@
...
+    __metaclass__ = ABCMeta
+
+    def __init__(self, model):
+        """
+        Init a wrapper. If `fprop` is implemented, `__init__`
+        should keep track of the name of the layers or `self.model` should
+        provide a method for retrieving a layer.
+
+        :param model: A function that takes a symbolic input and returns the
+                      symbolic output for the model's predictions.
```
npapernot: Should we specify here that this are post-softmax predictions (probabilities) to make it clear what the user should pass? 

To put into context, there is a lot of on-going ambiguity (#160) with the layers that need to be passed so the model abstraction you are contributing is a good opportunity to specify things like that.

cleverhans/model_abs.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        Create a wrapper for a Keras model
+
+        :param model: A Keras model
+        """
+        super(KerasModelWrapper, self).__init__(model)
+
+        # Initialize attributes
+        self.model = model
+        self.model_dict = {None: model}
```
npapernot: From what I understand, this `model_dict` attribute mapping layers to models should be implemented by all model abstractions? If that is indeed the case, would it make sense to move it to the abstract `Model` class?

cleverhans/model_abs.py
fartashf: Right. I had only Keras in mind when thinking about the overhead. But still I'm not sure returning all hidden layers at the same time is needed. Feature adversaries attack one layer at a time.

Let's make sure we agree on how the model abstraction is going to be used in the group. I send an email.

abhibhav14:
There is an error in the KerasModelWrappper fprop function, in the bit that checks if the model being requested is already cached. The current code attempts to call a dictionary. It should be 
return self.model_dict[layer](x)
and not
return self.model_dict(x)

cleverhans/model_abs.py
fartashf: Not all wrappers need to have caching. Keras models need it because they don't expose the hidden features themselves and we have to create a new model to expose a specific layer. But other models such as pure tf models don't necessarily need it because no new object is created.

fartashf:
Thanks Nicolas for the comments and approving. I made the new changes you suggested. Let's agree in the group if the current implementation is what we really want.

goodfeli:
This is worth taking some time to get right, so I suggest we write some code with it to make sure it really does what we want. This doesn't need to all be done by @fartashf ; some of the rest of us can also push to this branch.
I'd suggest:
- Convert the tutorials to use the new interface, and make sure everything all works
- Convert all the attacks to assume that the model includes a definition of the probabilities, and make sure the attacks all fetch either logits or probabilities as needed
- Maybe make an example that uses L1 regularization of the activities, or something like that that exercises the model interface more

fartashf:
I agree with converting the rest of the code to use the model abstraction.
We might need to change the model abstraction based on specific use-cases.

Let me make the change to `fprop` and `get_layer`. Then I can proceed with
converting attacks and others can help with the rest of your suggested list.

abhibhav14:
I can work on converting the tutorials that depend on Keras to use new new wrapper.

abhibhav14:
I have updated the attacks and made a PR to fartashfs branch over here https://github.com/fartashf/cleverhans/pull/1

npapernot:
Let us resolve the discussion on the `cleverhans-dev` mailing list before continuing to work on this PR: https://groups.google.com/forum/#!topic/cleverhans-dev/5F6XDlDB2S4

npapernot:
This will likely break the Travis build but it's probably best if we make sure we all agree before continuing to edit

fartashf:
I confirm the commits.

tests_tf/test_model.py
```diff
@@ -0,0 +1,188 @@
...
+        self.assertEqual(int(out_dict['l1'].shape[1]), 20)
+        self.assertEqual(int(out_dict['l2'].shape[1]), 10)
+
+        # Test the caching
+        x2 = tf.placeholder(tf.float32, shape=(None, 100))
+        out_dict2 = model.fprop(x2)
+        self.assertEqual(set(out_dict2.keys()), set(['l1', 'l2', 'softmax']))
+        self.assertEqual(int(out_dict2['l1'].shape[1]), 20)
+
+    def test_get_loss(self):
```
npapernot: If we decide to keep `get_loss`, this test needs to be fixed

cleverhans/utils_tf.py
```diff
@@ -49,13 +50,18 @@ def model_loss(y, model, mean=True):
              sample loss
     """
 
-    op = model.op
```
npapernot: Depending on what we decide to do with `Model.get_loss`, this edit should be removed or the docstring improved if we decide to keep `get_loss`

cleverhans/model.py
```diff
@@ -0,0 +1,253 @@
...
+        :return: Softmax layer name
+        """
+        for i, layer in enumerate(self.model.layers):
+            cfg = layer.get_config()
+            if 'activation' in cfg and cfg['activation'] == 'softmax':
+                return layer.name
+
+        raise Exception("No softmax layers found")
+
+    def _get_logits_name(self):
```
npapernot: This is not tested

cleverhans/model.py
```diff
@@ -0,0 +1,253 @@
...
+        Keras has its own way to keep track of the state of the graph so we'll
+        throw an exception here.
+        """
+        raise NotImplementedError('Keras keeps track of the graph state with'
+                                  'the learning phase variable. Use the'
+                                  'learning phase variable rather than this'
+                                  'method to change the behavior of the'
+                                  'graph.')
+
+    def _get_softmax_name(self):
```
npapernot: This is not tested

cleverhans/model.py
```diff
@@ -0,0 +1,221 @@
...
+
+class Model(object):
+    """
+    An abstract interface for model wrappers that exposes model symbols
+    needed for making an attack. This abstraction removes the dependency on
+    any specific neural network package (e.g. Keras) from the core
+    code of CleverHans. It can also simplify exposing the hidden features of a
+    model when a specific package does not directly expose them. For an
+    example of why this is useful, see the paper "Adversarial Manipulation
+    of Deep Representations" (https://arxiv.org/abs/1511.05122).
```
goodfeli: The last sentence is maybe not especially useful. We can definitely include a ref to this paper somewhere in the library---ideally when we add the attack from the paper.

cleverhans/model.py
```diff
@@ -0,0 +1,221 @@
...
+    """
+    __metaclass__ = ABCMeta
+
+    def __init__(self, model):
+        """
+        Init a wrapper. If `fprop_layer` is implemented, `__init__`
```
goodfeli: No need to say "Init a wrapper." That's just want an __init__ method does

cleverhans/model.py
```diff
@@ -0,0 +1,221 @@
...
+    needed for making an attack. This abstraction removes the dependency on
+    any specific neural network package (e.g. Keras) from the core
+    code of CleverHans. It can also simplify exposing the hidden features of a
+    model when a specific package does not directly expose them. For an
+    example of why this is useful, see the paper "Adversarial Manipulation
+    of Deep Representations" (https://arxiv.org/abs/1511.05122).
+    """
+    __metaclass__ = ABCMeta
+
+    def __init__(self, model):
```
goodfeli: This is an extremely weird interface. Why do we pass a callable into the __init__ method of the base class? Why not do something more standard, like implement a subclass that implements some method?

cleverhans/model.py
```diff
@@ -0,0 +1,221 @@
...
+        should keep track of the name of the layers or `self.model` should
+        provide a method for retrieving a layer.
+        :param model: A function that takes a symbolic input and returns the
+                      symbolic output for the model's post-softmax predictions
+                      (probabilities).
+        """
+        self.model = model
+
+        # The following is a cache to prevent the construction of increasingly
+        # large graphs after multiple calls of the fprop methods. The cache is
```
goodfeli: Wording cleanup, the new graphs aren't larger.

cleverhans/model.py
```diff
@@ -0,0 +1,221 @@
...
+        # implemented as a dictionary of the form (input, train): output_dict
+        # The key is a pair of input (the symbolic representation of the input)
+        # and train (a boolean indicating whether the graph is in training or
+        # inference mode---which changes the behavior of layers like dropout).
```
goodfeli: I'm not a big fan of the boolean for controlling behavior.

There are times when you have more than two behaviors. There are times when you want Monte Carlo dropout at test time.

It's probably better to get different behaviors by requesting different endpoints or by feeding different values to placeholders that control behavior at different layers.

cleverhans/model.py
```diff
@@ -0,0 +1,221 @@
...
+    def __call__(self, *args, **kwargs):
+        """
+        For compatibilty with a standard model.
+        """
+        return self.fprop_probs(*args, **kwargs)
+
+    def set_state(self, state):
+        """
+        Set which state the model is currently being used in. This can be any
+        string and is used to by the cache to keep different output tensors
```
goodfeli: Wait, is the state a string or a bool? It says it's a bool up above.

cleverhans/model.py
```diff
@@ -0,0 +1,221 @@
...
+        # By default, we assume the model is being used for inference (i.e.,
+        # 'test' time). If the model is being trained, a call to set_train()
+        # should be made first.
+        self.state = 'test'
+
+        pass
+
+    def __call__(self, *args, **kwargs):
+        """
+        For compatibilty with a standard model.
```
goodfeli: What is "a standard model"?

cleverhans/model.py
```diff
@@ -0,0 +1,221 @@
...
+        # implemented as a dictionary of the form (input, train): output_dict
+        # The key is a pair of input (the symbolic representation of the input)
+        # and train (a boolean indicating whether the graph is in training or
+        # inference mode---which changes the behavior of layers like dropout).
+        # The values output_dict are also a dictionary mapping layer names
+        # to symbolic representation of the output of that layer.
+        self.fprop_cache = {}
+
+        # By default, we assume the model is being used for inference (i.e.,
+        # 'test' time). If the model is being trained, a call to set_train()
```
goodfeli: Since every model is first trained and then tested, shouldn't it start out in train mode? (Assuming we keep the binary state system)

cleverhans/model.py
```diff
@@ -0,0 +1,221 @@
...
+        as a reference to the inference phase, but you could set it as 'train'
+        (which would for instance mean that the model uses dropout layers).
+        :param state: (boolean) True if the model should be in training state
+                      or False for inference.
+        """
+        self.state = state
+        print('The model state was set to: ' + state)
+        return
+
+    def fprop_layer(self, x, layer):
```
goodfeli: I think these names would be better as `get_layer`, `get_logits`, etc.
`fprop_layer` sounds like you're going to propagate the layer itself, `fprop_logits` sounds like you're going to propagate the logits somewhere, etc.

cleverhans/model.py
goodfeli: I actually don't see an implementation of `set_train`, where is it?

goodfeli:
Did we convert the tutorials to use the new interface?

cleverhans/model.py
npapernot: Agreed, I updating the names in 9ea2a3c

cleverhans/model.py
npapernot: That is true, unless the model is pre-trained. But in our tutorials we usually start by training the model so what you are suggesting makes more sense. 

Changed in e6b85de

cleverhans/model.py
npapernot: I think what @fartashf meant is the previous `model` argument that attacks where expecting (functions taking an input tensor and returning the tensor giving the output of the model on that input).

let me know if 6bffaab makes things more clear

cleverhans/model.py
npapernot: Fixed in 34112c0 and e35b2ef

cleverhans/model.py
npapernot: This was an outdated comment, there can be many states (the `self.state` attribute can be any string). Fixed in 34112c0

cleverhans/model.py
npapernot: Fixed in 34112c0

cleverhans/model.py
npapernot: Fixed in e2a7b19

cleverhans/model.py
npapernot: Fixed in 30ff8af

npapernot:
We haven't converted the tutorials yet, it's probably a good idea to do so before merging

cleverhans/model.py
npapernot: This is a good point, in most cases this argument won't be needed. Should we remove the argument to the init method? 

We'll just have to think of a way to give the Keras model function to the `KerasModelWrapper` subclass of `Model`

fartashf: It is true that this is a weird interface but that's only because as far as I know python does not have a nice way of casting an object to a subclass. @npapernot's commit 45bcf6dd3c8de22e747a6efefd6ead2ddbf6a7d3, address this issue by adding a separate `set_model` method. That way we can have models that implement `Model` without inheriting from any other class, e.g. Ian's pure TF model. But I think it still makes sense to have `__init__(self, model)` for `KerasModelWrapper` in addition to `__init(self)`. It can simply make a call to `set_model`.

npapernot: For the main class, I would prefer going with Ian's suggestion. Making design choices based on Keras has been a source of problem in the past for us. 

However, Nicholas is in the process of putting the model argument in the KerasModelWrapper's init method as an optional argument.

carlini:
Ian, Nicolas, are there any other changes that are needed to merge this?

npapernot:
Looks good on my end, Ian is traveling so he might need more time to review the PR

cleverhans/model.py
```diff
@@ -0,0 +1,107 @@
+from abc import ABCMeta
+
+
+class Model(object):
```
AlexeyKurakin: What's the decision regarding adding `loss` function to the model class?

cleverhans/model.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __call__(self, *args, **kwargs):
+        """
+        For compatibility with functions used as model definitions (taking
+        an input tensor and returning the tensor giving the output
+        of the model on that input).
+        """
+        return self.get_probs(*args, **kwargs)
+
+    def set_state(self, state):
```
carlini: I think making it a property makes sense.

cleverhans/model.py
```diff
@@ -0,0 +1,107 @@
...
+        raise NotImplementedError('`get_probs` not implemented')
+
+    def get_layer_names(self):
+        """
+        :return: a list of names for the layers that can be exposed by this
+        model abstraction.
+        """
+        raise NotImplementedError('`get_layer_names` not implemented')
+
+    def fprop(self, x):
```
carlini: This sounds like a good idea. The only concern is it might be slightly more work to make sure the logits layer is named 'logits' and not something else.

cleverhans/model.py
npapernot: Alex and I just talked in real life and we agreed to not include the model loss per the discussion we had on cleverhans-dev

cleverhans/model.py
npapernot: Good idea, see 824fa7d

cleverhans/model.py
goodfeli: Is there a particular use case where it would be hard to make sure the logits layer is named "logits"? It seems trivial to me but maybe I'm not being creative enough thinking of corner cases.

carlini: I'm just worried about other high-level libraries (keras/slim/etc) where you might not be able to control the name easily. I don't think implementing it this way is ever harmful, but sometimes might not work.

npapernot: Done, see 46e2f77

.travis.yml
```diff
@@ -1,4 +1,5 @@
 # adapted from https://github.com/fchollet/keras/blob/master/.travis.yml
+group: deprecated-2017Q2
```
goodfeli: This should be removed since the problem was the TensorFlow import, right?

cleverhans/model.py
```diff
@@ -0,0 +1,105 @@
...
+    An abstract interface for model wrappers that exposes model symbols
+    needed for making an attack. This abstraction removes the dependency on
+    any specific neural network package (e.g. Keras) from the core
+    code of CleverHans. It can also simplify exposing the hidden features of a
+    model when a specific package does not directly expose them.
+    """
+    __metaclass__ = ABCMeta
+
+    def __init__(self):
+        """
```
goodfeli: I don't understand this comment.
Does this mean that if a subclass implements `get_layer`, the subclass should also implement `__init__`? What needs to be done to keep track of the name of the layers?

cleverhans/model.py
```diff
@@ -0,0 +1,105 @@
...
+        :param x: A symbolic representation of the network input
+        :return: A dictionary mapping layer names to the symbolic
+                 representation of their output.
+        """
+        # In case of cache hit, return cached dictionary of output tensors.
+        if (x, self.state) in self.fprop_cache.keys():
+            return self.fprop_cache[(x, self.state)]
+        else:
+            result = self._fprop(x)
+            self.fprop_cache[(x, self.state)] = result
```
goodfeli: I'm confused about why we need the cache now. I thought the point of the cache was to avoid rebuilding TensorFlow graphs when people made more calls to generate_np. But this is for symbolic calls, not numpy calls. Is there a use case where we expect people to repeatedly request the same graph explicitly?

cleverhans/model.py
```diff
@@ -0,0 +1,105 @@
...
+    def __init__(self):
+        """
+        If `get_layer` is implemented, `__init__`
+        should keep track of the name of the layers.
+        """
+        # The following is a cache to prevent the re-construction of identical
+        # graphs after multiple calls of the fprop methods. The cache is
+        # implemented as a dictionary of the form (input, train): output_dict
+        # The key is a pair of input (the symbolic representation of the input)
+        # and train (a boolean indicating whether the graph is in training or
```
goodfeli: Is it actually a boolean? Elsewhere it looks like a string ("train")

tutorials/mnist_tutorial_pure_tf.py
```diff
@@ -31,32 +32,56 @@
...
   """
 
   def __init__(self, layers, input_shape):
+    super(MLP, self).__init__()
+
+    self.layer_names = []
     self.layers = layers
     self.input_shape = input_shape
-    for layer in self.layers:
+    for i,layer in enumerate(self.layers):
```
goodfeli: This looks like a PEP8 error, because there's no whitespace after the ",". Is the PEP8 check failing to run on this file?

tutorials/mnist_tutorial_pure_tf.py
```diff
@@ -31,32 +32,56 @@
...
+      name = layer.__class__.__name__
+      self.layer_names.append(name+"_"+str(i))
```
goodfeli: more PEP8 errors

tutorials/mnist_tutorial_pure_tf.py
```diff
@@ -31,32 +32,56 @@
...
+
       layer.set_input_shape(input_shape)
       input_shape = layer.get_output_shape()
 
-  def fprop(self, x, return_all=False, set_ref=False):
+  def get_layer_names(self):
+    return self.layer_names
+
+  def fprop(self, x, set_ref=False):
+    if (x, self.state) in self.fprop_cache.keys():
```
goodfeli: OK, this is concerning. I don't think everyone who implements a symbolic fprop should have to implement a caching mechanism. I thought we were going to put a caching mechanism in a base class's generate_np and then no one else would need to implement a new caching mechanism.

tutorials/mnist_tutorial_pure_tf.py
```diff
@@ -31,32 +32,56 @@
...
       if set_ref:
         layer.ref = x
       x = layer.fprop(x)
       assert x is not None
       states.append(x)
-    if return_all:
-      return states
-    return x
+    self.fprop_cache[x, self.state] = states
+    return states
```
goodfeli: I thought the Model class API specified that fprop returns a dictionary. This is returning a list.

tutorials/mnist_tutorial_pure_tf.py
carlini: I think Alex's recent suggestion to have a _fprop that does the heavy lifting, and make fprop just wrap _fprop with caching makes sense (if we want to keep the caching even).

tutorials/mnist_tutorial_pure_tf.py
carlini: I don't think PEP8 is set up to run on the tutorials. It seems to just run on cleverhans/ proper.

cleverhans/model.py
carlini: I'm not entirely sure why we have a cache, I agree. If you call model(xs) with the same xs multiple times, then I'm fine returning multiple graphs and letting the users shoot themselves in the foot, since that would be expected behavior.

.travis.yml
npapernot: Correct, I removed it

Issue #150, opened 2017-06-05
Fix generate_np bug


goodfeli:
When generate_np is called more than once with different arguments, it just replays the first graph.

I think it would be best to fix this by just removing generate_np, because some keyword arguments need to  trigger generation of new graphs and others don't. This seems like too subtle of a point for most Attack implementers to handle.

carlini:
This would have fairly large consequences for attacks that are not easy to (or perhaps even impossible to) implement fully symbolically.

For example, if you wanted to generate adversarial examples on a few thousand iterations using the attack algorithm I have, or JSMA as currently implemented, you would have to construct thousands of tf.gradient() operations, instead of caching it and just doing one.

goodfeli:
You can use the while loop to avoid making a large unrolled graph:
https://www.tensorflow.org/api_docs/python/tf/while_loop

Or you can wrap python code up to make a tensorflow op:
https://www.tensorflow.org/api_docs/python/tf/py_func

carlini:
Oh, I think I was unclear what I meant by "iterations" -- I didn't mean iterations of the attack algorithm, I meant iterations of calling attack.generate(). If we have to construct a new graph each time, that would be expensive.

goodfeli:
Oh, for that, it's straightforward, you just feed a different value each
time.

carlini:
This requires re-generating the entire graph each time though, right? You have to pay the cost of doing that every time you call generate(), when ideally you'd only construct the graph once and use it every time after that.

goodfeli:
No, you generate the graph once and feed many different values through it.
The tutorials already do this:
https://github.com/openai/cleverhans/blob/master/tutorials/mnist_tutorial_tf.py

carlini:
Yeah, but that only works because the FGS attack is a fully TF graph. Could you do this same thing with the JSMA code without having to call tf.gradients() each time you want a new adversarial example?

goodfeli:
Yes, by wrapping a python function that does the JSMA code in a tf op using
tf.py_func

carlini:
Oh, I didn't realize that would do that, okay, now I understand what you're saying. Then yes, I agree just removing generate_np would be the easiest solution.

npapernot:
The JSMA code is already being wrapped with `py_func` in the generate() method of `SaliencyMapMethod` https://github.com/openai/cleverhans/blob/master/cleverhans/attacks.py#L356

npapernot:
[...]

Issue #145, opened 2017-05-25
decide which version(s) of TF we support and add to README


npapernot:

npapernot:
see #138

npapernot:
Closed by #155

PR #155, opened 2017-06-07
Update Travis configuration matrix and README to reflect changes made to support specific TF versions and deprecate Theano support

npapernot:


carlini:
Should this wait until bug 150 is decided and resolved?

goodfeli:
Yeah, let's resolve #150 before we increment the tag.

npapernot:
We should also close #145 at the same time

.travis.yml
```diff
@@ -5,9 +5,21 @@ language: python
 matrix:
     include:
         - python: 2.7
-          env: KERAS_BACKEND=tensorflow
+          env:
+            - KERAS_BACKEND=tensorflow
+            - TENSORFLOW_V=1.0.1
```
goodfeli: We should also test tf versions 1.1 and 1.2, right? We don't need to get all combinations of python version and tensorflow version, but I think we should get all supported minor versions of tensorflow.

npapernot: I thought we were going for the last stable release (at the moment 1.1), and all past major versions (only 1.0 at the moment) ? This is what I meant here: https://groups.google.com/forum/#!topic/cleverhans-dev/TlyFojSkepU

Did you have something else in mind like all past major versions and all minor versions after the last major?

goodfeli: I think this is OK. I had made two mistakes:
- I didn't see that we do test v1.1 already; it just doesn't show up in the diff for this PR.
- I saw that a TensorFlow v1.2 is available, but it is not considered stable yet so we don't need to support it yet

README.md
```diff
@@ -68,26 +64,28 @@ export PYTHONPATH="/path/to/cleverhans":$PYTHONPATH
 
 You may want to make that change permanent through your shell's profile.
 
+### Currently supported setups
+
+Although CleverHans is likely to work on many other machine configurations, we
+currently [test it](https://travis-ci.org/tensorflow/cleverhans) with Python
+{2.7, 3.5} and TensorFlow {1.0, 1.2} on Ubuntu 14.04.5 LTS (Trusty Tahr).
```
goodfeli: Typo, 1.2 instead of 1.1

README.md
```diff
@@ -97,35 +95,44 @@ carefully by the adversary. The adversary then uses the substitute
...
 When reporting benchmarks, please:
-* Use a versioned release of `CleverHans`. You can find a list of released versions [here](https://github.com/tensorflow/cleverhans/releases).
+* Use a versioned release of CleverHans. You can find a list of released versions [here](https://github.com/tensorflow/cleverhans/releases).
 * Either use the latest version, or, if comparing to an earlier publication, use the same version as the earlier publication.
 * Report which attack method was used.
 * Report any configuration variables used to determine the behavior of the attack.
 
 For example, you might report "We benchmarked the robustness of our method to
-adversarial attack using v1.0.0 of `CleverHans`. On a test set modified by the
+adversarial attack using v2.0.0 of CleverHans. On a test set modified by the
```
goodfeli: This part still increments version number

README.md
npapernot: Fixed

README.md
npapernot: Fixed

PR #156, opened 2017-06-07
New L2 Attack Algorithm, generate_np rewrite

carlini:
I've implemented the attack algorithm I describe in "Towards Evaluating the Robustness of Neural Networks". This should not be merged until Bug 150 is decided for a proper implementation of how we want to do handle multiple calls to generate_np. Right now it is correct but not efficient.


cleverhans/attacks.py
```diff
@@ -486,6 +486,102 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
         return True
 
 
+class CarliniWagnerL2(Attack):
```
npapernot: This is simply a suggestion: most other attacks use a name that reflects at the high-level how they work. On the other hand, the attack is referred to as Carlini and Wagner in the literature. So I'm not sure what naming convention is most appropriate for here.

cleverhans/attacks.py
```diff
@@ -486,6 +486,102 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+            raise NotImplementedError('Theano version not implemented.')
+        self.attack_objects = {}
+
+    def generate(self, x, y=None, nb_classes=10,
+                 batch_size=1, confidence=0,
+                 targeted=True, learning_rate=1e-3,
+                 binary_search_steps=10, max_iterations=100,
+                 abort_early=True, initial_const=1e-2,
+                 clip_min=0, clip_max=1):
+
```
npapernot: The way arguments are parsed is not entirely consistent with other classes, which have a `parse_params` method to make sure that parameter values being fed have reasonable values. Do you think you could refactor the code to adopt a similar functionality?

cleverhans/attacks.py
```diff
@@ -486,6 +486,102 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+                      abort_early, initial_const, clip_min, clip_max,
+                      nb_classes, x.get_shape().as_list()[1:])
+
+        def cw_wrap(x_val, y_val):
+            return np.array(attack.attack(x_val, y_val), dtype=np.float32)
+
+        wrap = tf.py_func(cw_wrap, [x, y], tf.float32)
+        return wrap
+
+    def generate_np(self, x_val, y_val=None, nb_classes=10,
```
npapernot: Same comment about the arguments being parsed outside of the `parse_params` method.

cleverhans/attacks.py
```diff
@@ -486,6 +486,102 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+                    targeted=True, learning_rate=5e-3,
+                    binary_search_steps=5, max_iterations=1000,
+                    abort_early=True, initial_const=1e-2,
+                    clip_min=0, clip_max=1):
+
+        """
+        Generate adversarial samples and return them in a Numpy array.
+
+        :param x_val: (required) A Numpy array with the original inputs.
+        :param y_val: (required) A Numpy array with the labels that we either
```
npapernot: Does that mean that the label vector should always be provided?

tests_tf/test_attacks.py
```diff
@@ -40,7 +39,7 @@ def test_generate_np(self):
         perturbation_norm = np.sqrt(np.sum(perturbation**2, axis=1))
         # test perturbation norm
         self.assertTrue(np.allclose(perturbation_norm, self.attack.eps))
-
+
```
npapernot: looks like the test was not checked in? Also might be easier to manage if different attacks using different files for their test. (could rename the test_attacks.py to be test_vat.py ---sorry didn't catch this earlier in the PR about VAT)

npapernot:
Thanks for the PR! I took a first pass and left a few comments. I'll have to take a closer look again later

tests_tf/test_attacks.py
npapernot: Nevermind about creating separate files, as long as the tests are placed in the corresponding test file (i.e., `test_attacks.py` or `test_attacks_tf.py` (which does not exist yet)

goodfeli:
I've been waiting to review this until we settle #150, which will hopefully be this morning.

npapernot:
+1 on doing things in that order, so we don't have to rebase the PR

cleverhans/attacks.py
carlini: No longer required.

cleverhans/attacks.py
carlini: Yeah ... don't have a good name.

carlini:
My rebase messed up the comments.

I think I'm going to avoid rewriting JSMA and VAT generate_np in this PR and make another one for those after this one gets merged in.

carlini:
Given the recent discussion regarding PR #154, I think this code can be merged mostly independently (that is, attacks will still work the same way by using __call__, they will just be called differently). The only conflicts will be in the tutorials.

carlini:
I rebased again and it seems this has confused googlebot.

cleverhans/attacks.py
```diff
@@ -48,48 +58,97 @@ def generate(self, x, **kwargs):
...
+            x = tf.placeholder(tf.float32, shape=[None]+list(x_val.shape)[1:])
+
+            # now we generate the graph that we want
+            x_adv = self.generate(x, **new_kwargs)
+
+            if hash_key is not None:
+                # only save the graph if every fixed element is hashable
+                self.graphs[hash_key] = (x, new_kwargs, x_adv)
+
+            if len(self.graphs) == 10:
```
AlexeyKurakin: Maybe `len(self.graphs) >= 10` ?
Also consider graph cache with expiration of old values (maybe in separate pull request after this one)

cleverhans/attacks.py
```diff
@@ -48,48 +58,97 @@ def generate(self, x, **kwargs):
...
             raise ValueError("Cannot use `generate_np` when no `sess` was"
                              " provided")
-        return self.sess.run(self._x_adv, feed_dict={self._x: x_val})
+
+        if 'y' in kwargs and kwargs['y'] is None:
+            del kwargs['y']
+
+        # the set of arguments that are structural properties of the attack
+        # if these arguments are different, we must construct a new graph
+        fixed = dict((k, v) for k, v in kwargs.items()
```
AlexeyKurakin: It would be nice if each attack have a list of allowed structural arguments, and here only allowed structural arguments will be included in `fixed`

It is possible that user of Cleverhans will use attacks interchangeably, so will provide list of structural arguments for all attacks combined. If this is the case then `fixed` will included arguments which are actually not used by current attack (and provided for support of other attack). Then these arguments will be used as a part of hash key for the graph which might result in having identical graphs with different keys.

cleverhans/attacks.py
carlini: Sure, no harm doing that.

cleverhans/attacks.py
carlini: To clarify: you're pointing out that users may pass extra arguments which aren't important, and then we will key off those even though we shouldn't be?

Then yes, I could fix this by making the attack define what structural arguments it cares about.

AlexeyKurakin: yes. user may pass extra argument which attack ignores, however key will included these arguments

carlini: This is done now in fad875b.

goodfeli:
OK, model abstraction merged, ready to rebase this

carlini:
Working on that now.

carlini:
This is an ugly rebase. Will pick up tomorrow.

npapernot:
OK, let us know if you need some help with the rebase

npapernot:
I just realized this PR already has +4,000 lines of delta so it may be hard to review the changes effectively. Should we try to create a new one or even two new ones (one for generate_np and one for the new attack)?

carlini:
I'm basically working on doing that. I'm replaying my commits off the current master and removing everything from 665a5b4 to 3a155f0 and from d221450 to d3416eb that got smashed in here, and making a new branch with just the changes I've made.

carlini:
Regarding making two PRs, one for the attack and one for generate_np: I can do that, sure. The best thing to do for that would be to merge generate_np first, then the L2 attack (and make it use the new generate_np functionality).

This shouldn't take more than few hours to split everything apart and replay the commits, I should have it done by tonight.

carlini:
Per discussion, closing this PR and will open two new ones.

PR #157, opened 2017-06-07
Update the JSMA docstring to clarify that it can take both logits and softmax outputs

carlini:
JSMA is implemented at the logit layer, pre-softmax. However, the tutorial implements it after the softmax. Add some (keras-specific) code to remove the last softmax layer.

### DIFF
--- a/tutorials/mnist_tutorial_jsma.py
+++ b/tutorials/mnist_tutorial_jsma.py
@@ -112,6 +112,14 @@ def mnist_tutorial_jsma(train_start=0, train_end=60000, test_start=0,
     grid_shape = (nb_classes, nb_classes, img_rows, img_cols, channels)
     grid_viz_data = np.zeros(grid_shape, dtype='f')
 
+    # by default, we have softmax after a CNN, remove it here
+    # JSMA takes the the logits, pre-softmax
+    model.layers.pop()
+    last = model.layers[-1]
+    last.outbound_nodes = []
+    model.outputs = [last.output]
+    model.built = False
+
     # Instantiate a SaliencyMapMethod attack object
     jsma = SaliencyMapMethod(model, back='tf', sess=sess)
     jsma_params = {'theta': 1., 'gamma': 0.1,


carlini:
Hmm ... it looks like the accuracy test has failed. Nicolas, should JSMA operate post-softmax on these networks? If so, the documentation should probably change.

goodfeli:
I haven't looked at it, but the JSMA implementation might be crawling the
TensorFlow graph to obtain the logits after you pass it probabilities.

If that's what it's doing, and if it runs without raising an exception when
you pass it the logits, then it should probably add a check that the
probabilities it receives are the output of the Softmax op.

carlini:
It looks like it's not doing this (even if it was, then removing the softmax should be a no-op).

npapernot:
The JSMA should be able to run with both the logits and softmax outputs. Did you make this PR because the implementation failed on some network?

npapernot:
I talked to Nicholas in real life :) There is no specific need to remove the softmax layer at the moment. However, we should update the documentation to say that the model can be pre or post softmax. 

I will also open a new issue, because we could build on #154, and additional code, to clarify which attacks need the logits and which ones need the softmax outputs. In addition to improving the documentation, this may require adding some sanity checks on the `model` argument when initializing the different attacks.

goodfeli:
I think ideally we should make them all take probabilities and they can
recover the logits if needed. Maybe we should pull out the code for doing
that and turn it into a utility function.

npapernot:
Would it make sense to require that the model argument be an instance of the new model abstraction class? This would allow the attack to have more guarantees about the structure of the model. It would also make it easier for attacks to have a robust access to the logits layer or softmax layer.

tutorials/mnist_tutorial_jsma.py
```diff
@@ -112,6 +112,14 @@ def mnist_tutorial_jsma(train_start=0, train_end=60000, test_start=0,
     grid_shape = (nb_classes, nb_classes, img_rows, img_cols, channels)
     grid_viz_data = np.zeros(grid_shape, dtype='f')
 
+    # by default, we have softmax after a CNN, remove it here
```
goodfeli: It's probably better to change JSMA to take probabilities rather than to modify the tutorial

npapernot: The JSMA implementation would not change whether it is being fed probabilities or logits, it can handle both. That's why we were thinking of updating the documentation to make that more clear rather than modifying the tutorial (the few lines added to the tutorial in this PR would be removed). Does that make sense or do you prefer an alternative?

goodfeli:
Didn't Nicolas see a difference in performance when he changed the tutorial
from passing probabilities to passing logits?

npapernot:
I talked to Nicholas about the performance drop he observed when changing the tutorial.

Basically, users should first try to run the JSMA attack using probabilities, which is likely to work better in most cases. If they encounter numerical issues when computing the Jacobian, they can try feeding in the logits directly, which can work better if the numerical instabilities are due to the softmax. This is why we used the logits in the original paper, and then the probabilities in subsequent papers.

goodfeli:
I'd suggest making JSMA use the new Model abstraction, and then have the
attack explicitly request either logits or probabilities from the Model
based on a flag in the __init__ method for JSMA. That way people don't need
to redefine their model to change the attack behavior.

PR #164, opened 2017-06-08
Rename test_attacks.py -> test_vat.py

carlini:
As suggested in PR 156.


carlini:
Looks like the build failed due to a HTTP read timeout, so everything should be fine here.

npapernot:
I restarted the test

aam-at:
The idea is to have all attack tests in one file. It follows a standard python convention where tests for module file are located in `test_<file_name>`. Imho, I don't think that having tests in one will make things unmanageable.

npapernot:
That's fair, I'll close it in that case and delete my comment from #156

Issue #170, opened 2017-06-13
Try Adam for the tutorials


goodfeli:
In a private experiment, I found that Adam with a learning rate of .001 works much better than the tutorial default. It's worth checking if this holds true for the tutorial or if it was specific to the model I was using in my private experiments.

abhibhav14:
The model we get is quite a bit more accurate with Adam, reaching 0.9933 compared to the 0.9774 of the current model, but at the cost of accuracy on the adversarial samples, which goes up from 0.0124 to 0.0277, all on MNIST. The loss function itself probably becomes zero for some inputs, which causes the gradient sign method to fail more.

goodfeli:
After how many epochs? What about the accuracy of the adversarially trained model?
If you run it multiple times, I suspect you'll see a lot of fluctuation in the accuracy of the adversarial examples.
I doubt the loss is hitting machine zero. It sounds like the model is just better.

abhibhav14:
The above was after 6 epochs.
I am getting variation over run. In the end, after the adversarial training I get 0.9614 accuracy on the adversarial samples.
And yeah, the model is not hitting machine zero in this case.

goodfeli:
What was the accuracy on adversarial examples before the switch to Adam?

abhibhav14:
using the original Adadelta: before training - 0.013 and after training 0.56
using adam: before training - 0.028 and after training 0.96
Adam seems overall better.

goodfeli:
I'm assuming "before training" = "normal training" and "after training" =
"adversarial training". That does look like a big improvement.

Could you send a PR?

abhibhav14:
I will run everything a couple of more times to make sure everything runs smoothly, then send one in.

abhibhav14:
@goodfeli Everything checks out fine, I have made a PR #171 where I made the change to the utils_tf and also all the relevant tutorials.

npapernot:
#171 was merged, so it looks like we can close this

abhibhav14:
Should we increase the values in the tests to match? For example, right now the test_mnist_tutorial_tf just expects adv_train_adv_eval to be greater than 0.06, but we can now increase this by a bit.

goodfeli:
Yes, we should raise it because failing to get the usual numbers for Adam
could indicate a bug.

npapernot:
Good catch, I opened #176 to not forget about this

PR #171, opened 2017-06-16
Switch to the adam optimizer

abhibhav14:
Changed optimizer to Adam and default learning rates to 1e-3 for tf as discussed in #170 .


npapernot:
@abhibhav14, I confirm your observations, the accuracies are also greatly improved upon by the switch to adam. Thanks for running the experiments and putting together the PR!

Closes #170

PR #187, opened 2017-07-06
Fix Pyhton 2/3 compatibility issues to fix #185

### LINKED ISSUE #185: Python 3 compatibility issues in download_images.py
Following lines result in Python 3 compatibility issues

examples/nips17_adversarial_competition/dataset/download_images.py
```python
import cStringIO
...
import urllib2
```
[Python 3 changelog](https://docs.python.org/3.0/whatsnew/3.0.html#text-vs-data-instead-of-unicode-vs-8-bit) says
> The StringIO and cStringIO modules are gone. Instead, import the io module and use io.StringIO or io.BytesIO for text and data respectively.

Since the usage of `cStringIO` is `image = Image.open(cStringIO.StringIO(image_buffer)).convert('RGB')` I think it makes sense to replace it with `io.BytesIO`, which is present in both Python 2.7 (introduced in 2.6) and Python 3.X.

`urllib2` was split into several modules in Python 3 and I think the correct solution would be to try importing `urllib.request` for Python 3 and fall back to `urllib2` for Python 2.

The following diff makes me able to download the dataset and I'd be more than happy to submit it if it proves to be useful, but I'm not sure what the default policy about Python 2/Python 3 compatibility is.

```
diff --git a/examples/nips17_adversarial_competition/dataset/download_images.py b/examples/nips17_adversarial_competition/dataset/download_images.py                                                        
index 20f5564..20bb579 100644                                                                                                                                                                               
--- a/examples/nips17_adversarial_competition/dataset/download_images.py                                                                                                                                    
+++ b/examples/nips17_adversarial_competition/dataset/download_images.py                                                                                                                                    
@@ -24,14 +24,17 @@ from __future__ import division                                                                                                                                                         
 from __future__ import print_function                                                                                                                                                                      
                                                                                                                                                                                                            
 import argparse                                                                                                                                                                                            
-import cStringIO                                                                                                                                                                                           
 import csv                                                                                                                                                                                                 
 import os                                                                                                                                                                                                  
 import sys                                                                                                                                                                                                 
-import urllib2
 
 from PIL import Image
+from io import BytesIO
 
+try:
+    from urllib.request import urlopen
+except ImportError:
+    from urllib2 import urlopen
 
 def parse_args():
   """Parses command line arguments."""
@@ -53,12 +56,12 @@ def download_image(image_id, url, x1, y1, x2, y2, output_dir):
     return True
   try:
     # Download image
-    url_file = urllib2.urlopen(url)
+    url_file = urlopen(url)
     if url_file.getcode() != 200:
       return False
     image_buffer = url_file.read()
     # Crop, resize and save image
-    image = Image.open(cStringIO.StringIO(image_buffer)).convert('RGB')
+    image = Image.open(BytesIO(image_buffer)).convert('RGB')
     w = image.size[0]
     h = image.size[1]
     image = image.crop((int(x1*w), int(y1*h), int(x2*w), int(y2*h)))
```

omtcyfz:
Submitting the diff I attached to address #185, as advised by @goodfeli.

### DIFF
--- a/examples/nips17_adversarial_competition/dataset/download_images.py
+++ b/examples/nips17_adversarial_competition/dataset/download_images.py
@@ -24,14 +24,17 @@ from __future__ import division
 from __future__ import print_function
 
 import argparse
-import cStringIO
 import csv
 import os
 import sys
-import urllib2
 
 from PIL import Image
+from io import BytesIO
 
+try:
+    from urllib.request import urlopen
+except ImportError:
+    from urllib2 import urlopen
 
 def parse_args():
   """Parses command line arguments."""
@@ -53,12 +56,12 @@ def download_image(image_id, url, x1, y1, x2, y2, output_dir):
     return True
   try:
     # Download image
-    url_file = urllib2.urlopen(url)
+    url_file = urlopen(url)
     if url_file.getcode() != 200:
       return False
     image_buffer = url_file.read()
     # Crop, resize and save image
-    image = Image.open(cStringIO.StringIO(image_buffer)).convert('RGB')
+    image = Image.open(BytesIO(image_buffer)).convert('RGB')
     w = image.size[0]
     h = image.size[1]
     image = image.crop((int(x1*w), int(y1*h), int(x2*w), int(y2*h)))


omtcyfz:
I signed it!

Issue #189, opened 2017-07-07
error in mnist_tutorial_jsma.py


kirk86:
There is the following error when one defines a different number of classes as and argument.
```
Attacking input 1/10
Traceback (most recent call last):
  File "mnist_tutorial_jsma.py", line 226, in <module>
    app.run()
  File "/miniconda2/lib/python2.7/site-packages/tensorflow/python/platform/app.py", line 48, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "mnist_tutorial_jsma.py", line 215, in main
    learning_rate=FLAGS.learning_rate)
  File "mnist_tutorial_jsma.py", line 141, in mnist_tutorial_jsma
    target_classes = other_classes(nb_classes, current_class)
    other_classes_list.remove(class_ind)
ValueError: list.remove(x): x not in list
```

npapernot:
It looks from the error that you are trying to use a class index `class_ind` that is too large (i.e., larger than `nb_classes`). Otherwise, could you give us a bit more context so we can understand what went wrong?

kirk86:
@npapernot yeah sure, apologies, my bad. So what I did is I just run the tutorial example without changing anything. The only thing that I changed is the user defined argument `nb_classes`. The default value is set to 10 I just used 5 in this case.

npapernot:
If you only changed the `nb_classes` to 5, it is normal to get this error because some of the inputs have labels between 5 and 9 (in `Y_test`).

If you want to reduce `nb_classes` to 5, you should also filter the data so it only includes examples from classes `0` to `4` included.

PR #191, opened 2017-07-08
Add argument checker and tests for `utils.other_classes`, improve docstring

npapernot:
This follows issue #189 and should make it easier to understand the error encountered

### DIFF
--- a/tests_tf/test_utils_tf.py
+++ b/tests_tf/test_utils_tf.py
@@ -12,7 +12,7 @@ from cleverhans import utils_tf
 
 def numpy_kl_with_logits(p_logits, q_logits):
     def numpy_softmax(logits):
-        logits = (logits.transpose() - np.max(logits, axis=1)).transpose()
+        logits -= np.max(logits, axis=1, keepdims=True)
         exp_logits = np.exp(logits)
         return exp_logits / np.sum(exp_logits, axis=1, keepdims=True)
 
--- a/tests_tf/test_utils_tf.py
+++ b/tests_tf/test_utils_tf.py
@@ -17,8 +17,9 @@ def numpy_kl_with_logits(p_logits, q_logits):
         return exp_logits / np.sum(exp_logits, axis=1, keepdims=True)
 
     p = numpy_softmax(p_logits)
-    q = numpy_softmax(q_logits)
-    return (p * (np.log(p) - np.log(q))).sum(axis=1).mean()
+    log_p = p_logits - np.log(np.sum(np.exp(p_logits), axis=1, keepdims=True))
+    log_q = q_logits - np.log(np.sum(np.exp(q_logits), axis=1, keepdims=True))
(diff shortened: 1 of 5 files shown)


npapernot:
This also cleans up the `tests_tf` folder (some of the tests were not in the right files)

tests_tf/test_utils_tf.py
```diff
@@ -0,0 +1,45 @@
...
+
+import numpy as np
+import unittest
+import tensorflow as tf
+
+from cleverhans import utils_tf
+
+
+def numpy_kl_with_logits(q_logits, p_logits):
+    def numpy_softmax(logits):
```
goodfeli: This version of the softmax is numerically unstable. Should subtract off the max of logits before applying the exp.

tests_tf/test_utils_tf.py
```diff
@@ -0,0 +1,45 @@
...
+        exp_logits = np.exp(logits)
+        return exp_logits / np.sum(exp_logits, axis=1, keepdims=True)
+
+    q = numpy_softmax(q_logits)
+    p = numpy_softmax(p_logits)
+    return (q * (np.log(q) - np.log(p))).sum(axis=1).mean()
```
goodfeli: This is not the numerically stable implementation of cross-entropy.
You need to expand
log softmax(x)_i = x_i - log sum exp(x_i)
or you can run into saturation issues.

tests_tf/test_utils_tf.py
```diff
@@ -0,0 +1,45 @@
...
+from __future__ import unicode_literals
+
+import numpy as np
+import unittest
+import tensorflow as tf
+
+from cleverhans import utils_tf
+
+
+def numpy_kl_with_logits(q_logits, p_logits):
```
goodfeli: This function is already defined in tests_tf/test_utils.py, with the same mistakes. Maybe we should put one version of it in a shared location, and write some tests for it, using extreme values of logits that would break the current version without the numerical stabilization tricks.

goodfeli: It's probably a bit confusing to make q be the left argument, because definitions of the KL divergence almost always use p as the left argument. https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence

tests_tf/test_utils_tf.py
```diff
@@ -0,0 +1,47 @@
...
+import numpy as np
+import unittest
+import tensorflow as tf
+
+from cleverhans import utils_tf
+
+
+def numpy_kl_with_logits(p_logits, q_logits):
+    def numpy_softmax(logits):
+        logits = (logits.transpose() - np.max(logits, axis=1)).transpose()
```
goodfeli: instead of two transposes, it's better to do keepdims=True in the max

npapernot: Thanks :)

PR #192, opened 2017-07-08
Make improvements to utils.random_target

npapernot:
This fixes the functionality of random_target when multiple labels are given to it at once. It also adds tests for this function.

Should only be merged once #191 is merged. (A small re-base will likely be needed---I can take care of it)

### DIFF
--- a/cleverhans/utils.py
+++ b/cleverhans/utils.py
@@ -75,6 +75,24 @@ def other_classes(nb_classes, class_ind):
     return other_classes_list
 
 
+def to_categorical(y, num_classes=None):
+    """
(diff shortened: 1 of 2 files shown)


goodfeli:
This will need a rebase

cleverhans/utils.py
```diff
@@ -77,24 +77,62 @@ def other_classes(nb_classes, class_ind):
     return other_classes_list
 
 
+def to_categorical(y, num_classes=None):
```
goodfeli: I think the tf.one_hot op does this already

Issue #190, opened 2017-07-07
Run PEP8 checker on tutorials


goodfeli:

goodfeli:
Closes #179

Issue #179, opened 2017-06-26
change spacing to 4 spaces in pure tf tutorial


npapernot:
all other code has 2 spaces

goodfeli:
I think we should standardize on 2 spaces for consistency with TensorFlow, especially now that this repository lives in the TensorFlow organization.

carlini:
I don't have a strong opinion, but PEP8 says python should have 4 spaces (and all non-google python code I'm aware of uses 4). Do you see any benefits of going to 2 spaces?

goodfeli:
actually I'm confused by @npapernot 's earlier comment; it looks like most of CleverHans has 4 spaces

npapernot:
My bad. This was a mistake, I meant to say that most of the code base has 4 spaces, but the tutorial has 2 spaces. 

I don't have a preference towards 2 or 4 spaces and am happy to use either.

PR #194, opened 2017-07-09
Pep8

goodfeli:


npapernot:
Closes #190 

Closes #179

This also closes 181 right?

Issue #176, opened 2017-06-22
reassess tutorial tests min / max accuracies and make tutorial tests less flaky


npapernot:

npapernot:
see #170

abhibhav14:
I am currently working on it right now, For the MNIST tutorials, we can increase the adversarial examples after adversarial training from 0.07 to 0.70.
However, I just realized that switching to Adam messed up the blackbox tutorial. Originally, test accuracy of oracle on samples generated by substitute was roughly 0.5, but switching to Adam has actually made that worse, it is now a little over 0.8.

npapernot:
@abhibhav14  Did you make any progress on this? 

If not, I'll probably make a PR soon to address this issue because some of the tests are more frequently flaky because of the new optimizers

abhibhav14:
For each of the tests, I have reassessed the accuracies and made a pull request at #197 . For the blackbox attack, the accuracy of the blackbox on the adversarial examples in the test is quite low, since the blackbox network is trained on a limited set on a few epochs. However, I still have not been able to fix the tutorial. I tried doubling the holdout, but this barely decreases the accuracy by about 0.05 while taking up double the time as before.

npapernot:
Running the test on the black-box tutorial is hard because it is expensive to do the data augmentation. One way to make the test more representative might be to make the black-box and substitute models more similar in the tutorial itself.

npapernot:
I am closing this given that the problem not addressed yet in this issue is also described in #183. Tests are running more smoothly now.

PR #197, opened 2017-07-09
Reassessed tutorial accuracies

abhibhav14:
Fix for #176

### DIFF
--- a/cleverhans/utils.py
+++ b/cleverhans/utils.py
@@ -37,6 +37,11 @@ class AccuracyReport(object):
         self.adv_train_clean_eval = 0.
         self.adv_train_adv_eval = 0.
 
+        # Training data accuracy results to be used by tutorials
+        self.train_clean_train_clean_eval = 0.
+        self.train_clean_train_adv_eval = 0.
+        self.train_adv_train_clean_eval = 0.
+        self.train_adv_train_adv_eval = 0.
 
 def batch_indices(batch_nb, data_length, batch_size):
     """
--- a/cleverhans/utils.py
+++ b/cleverhans/utils.py
@@ -43,6 +43,7 @@ class AccuracyReport(object):
         self.train_adv_train_clean_eval = 0.
         self.train_adv_train_adv_eval = 0.
 
+
 def batch_indices(batch_nb, data_length, batch_size):
     """
     This helper function computes a batch start and end index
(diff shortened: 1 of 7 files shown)


npapernot:
Thanks @abhibhav14, I'll run the Travis tests a couple more times to make sure they are not flaky anymore.

We might want to reassess the JSMA tutorial tests, I think that the threshold on the accuracy on adversarial examples is too low now that the optimizer yields more accurate models.

abhibhav14:
@npapernot I have commented the same thing on the Issue thread, but please note that the Blackbox tutorial is still broken. The tests pass because of the reduced input and epochs and data size, but when the blackbox is trained properly, the the adversarial attack accuracy is still as high as 80%.

npapernot:
I'll take a look at the tutorial once we've merge the `generate_np` method and the new attack 

Could we also add numpy and tensorflow seeds to the tests, to minimize the flakyness of tests?

abhibhav14:
I added the random seeds and also changed the accuracy to that on the training samples for the two FGSM attacks. I will soon add the same for the JSMA tutorial too.

goodfeli:
OK, thanks. Now that it's seeded, could you post the results of 3-5 runs for us to see what it does on your machine?

abhibhav14:
I unfortunately do not get determinism, but the differences are small enough to not matter.
For the keras tutorial:
0.973, 0.9736, 0.9732, 0.9736
0.0096, 0.0106, 0.0110, 0.009
0.9496, 0.9512, 0.9498, 0.951
0.4308, 0.4288, 0.4270, 0.432
where the four rows are four runs of clean_trian_clean eval, clean_train_adv_eval, adv_train_clean_eval and adv_train_adv eval

Similar results for the pure tf tutorial are:
0.9618, 0.9642, 0.9602, 0.9638
0.0220, 0.0202, 0.0220, 0.1980
0.9410, 0.9410, 0.9436, 0.9406
0.4052, 0.4296, 0.4422, 0.4224

goodfeli:
Seems like something is still not seeded right.
- Is it deterministic if you turn off dropout?
- Is the loss right after initialization deterministic? Or do you need to do some learning before it becomes non-deterministic?

What are the accuracies on the smaller versions of the tutorials used in the tests?

abhibhav14:
The above results are on the test versions of the tutorials, and the results on the actual tutorials are quite similar, except the results of the adversarial training, which are much better. However, there is slightly more variation, despite the seeding.
As for the second point, there is a small difference in both the loss after initialization for the first batch, and also the prediction on the first sample before training, however this difference is beyond the third decimal place.

goodfeli:
OK. It's a bit weird that it's still stochastic, but this is definitely an improvement.

PR #198, opened 2017-07-09
Fix generate_np issues

carlini:
This commit fixes generate_np as discussed in bug #150. Calling generate is now the default expected behavior; generate_np is meant as a simple way for people getting started with the code. The Attack class now implements generate_np by wrapping generate for all attacks.

These commits are cherry-picked from PR #156.

PR #199, opened 2017-07-10
Use new model in attacks code

carlini:
After doing everything required to create a unified Model interface, the attacks don't actually use it. Make the existing attacks use it.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -221,7 +221,7 @@ class FastGradientMethod(Attack):
         else:
             from .attacks_th import fgm
 
-        return fgm(x, self.model.get_logits(x), y=self.y, eps=self.eps,
+        return fgm(x, self.model.get_probs(x), y=self.y, eps=self.eps,
                    ord=self.ord, clip_min=self.clip_min,
                    clip_max=self.clip_max)
 
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -18,10 +18,7 @@ class Attack(object):
 
     def __init__(self, model, back='tf', sess=None):
         """
-        :param model: An instance of the cleverhans.Model class, or optionally
-                      (but not recommended) a function that takes a symbolic
-                      input and returns the symbolic output for the model's
-                      predictions.
+        :param model: An instance of the Model class.
         :param back: The backend to use. Either 'tf' (default) or 'th'.
         :param sess: The tf session to run graphs in (use None for Theano)
         """
@@ -32,8 +29,8 @@ class Attack(object):
         if not isinstance(model, Model):
             if hasattr(model, '__call__'):
                 warnings.warn("The model argument should be an instance of the"
-                              " cleverhans.Model class; otherwise the oytput"
-                              " layer will be inferred by the attack.")
+                              " Model class; otherwise the output layer will"
+                              " be inferred by the attack.")
             else:
                 raise ValueError("model argument must be a function that "
                                  "returns the symbolic output when given an "
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -6,7 +6,7 @@ import collections
 
 import cleverhans.utils as utils
 
-from cleverhans.model import Model, WrapCallable
+from cleverhans.model import Model, CallableModelWrapper
 
 
 class Attack(object):
@@ -198,7 +198,7 @@ class FastGradientMethod(Attack):
         self.structural_kwargs = ['ord']
 
         if not isinstance(self.model, Model):
-            self.model = WrapCallable(self.model, 'probs')
+            self.model = CallableModelWrapper(self.model, 'probs')
 
     def generate(self, x, **kwargs):
         """
@@ -283,7 +283,7 @@ class BasicIterativeMethod(Attack):
         self.structural_kwargs = ['ord', 'nb_iter']
 
         if not isinstance(self.model, Model):
-            self.model = WrapCallable(self.model, 'probs')
+            self.model = CallableModelWrapper(self.model, 'probs')
 
     def generate(self, x, **kwargs):
         import tensorflow as tf
@@ -379,7 +379,7 @@ class SaliencyMapMethod(Attack):
         super(SaliencyMapMethod, self).__init__(model, back, sess)
 
         if not isinstance(self.model, Model):
-            self.model = WrapCallable(self.model, 'probs')
+            self.model = CallableModelWrapper(self.model, 'probs')
 
         if self.back == 'th':
             error = "Theano version of SaliencyMapMethod not implemented."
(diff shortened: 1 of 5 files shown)


carlini:
(This should be merged before my L2 attack.)

carlini:
There is a flaky test causing this to fail that I fix in #198.

npapernot:
Thanks, I restarted the test

cleverhans/attacks.py
```diff
@@ -13,18 +15,26 @@ class Attack(object):
...
-        if not hasattr(model, '__call__'):
-            raise ValueError("model argument must be a function that returns "
-                             "the symbolic output when given an input tensor.")
+        if not isinstance(model, Model):
+            if hasattr(model, '__call__'):
+                warnings.warn("The model argument should be an instance of the"
+                              " cleverhans.Model class; otherwise the oytput"
+                              " layer will be inferred by the attack.")
+            else:
+                raise ValueError("model argument must be a function that "
```
npapernot: Should we also recommend that people use the model class in this ValueError?

cleverhans/attacks.py
```diff
@@ -13,18 +15,26 @@ class Attack(object):
 
     def __init__(self, model, back='tf', sess=None):
         """
-        :param model: A function that takes a symbolic input and returns the
-                      symbolic output for the model's predictions.
+        :param model: An instance of the cleverhans.Model class, or optionally
```
goodfeli: I think it's OK to just call it `Model`. There actually isn't a `cleverhans.Model` at the moment, though I think it would be fine to add an import of the `Model` class at the base module

cleverhans/attacks.py
```diff
@@ -13,18 +15,26 @@ class Attack(object):
...
             raise ValueError("Backend argument must either be 'tf' or 'th'.")
         if back == 'th' and sess is not None:
             raise Exception("A session should not be provided when using th.")
-        if not hasattr(model, '__call__'):
-            raise ValueError("model argument must be a function that returns "
-                             "the symbolic output when given an input tensor.")
+        if not isinstance(model, Model):
+            if hasattr(model, '__call__'):
+                warnings.warn("The model argument should be an instance of the"
+                              " cleverhans.Model class; otherwise the oytput"
```
goodfeli: typo "oytput"

cleverhans/attacks.py
goodfeli: Yes

cleverhans/attacks.py
```diff
@@ -13,18 +15,26 @@ class Attack(object):
 
     def __init__(self, model, back='tf', sess=None):
         """
-        :param model: A function that takes a symbolic input and returns the
-                      symbolic output for the model's predictions.
+        :param model: An instance of the cleverhans.Model class, or optionally
+                      (but not recommended) a function that takes a symbolic
```
goodfeli: I would consider not documenting the less preferred option

cleverhans/attacks.py
```diff
@@ -13,18 +15,26 @@ class Attack(object):
...
         if not(back == 'tf' or back == 'th'):
             raise ValueError("Backend argument must either be 'tf' or 'th'.")
         if back == 'th' and sess is not None:
             raise Exception("A session should not be provided when using th.")
-        if not hasattr(model, '__call__'):
-            raise ValueError("model argument must be a function that returns "
-                             "the symbolic output when given an input tensor.")
+        if not isinstance(model, Model):
+            if hasattr(model, '__call__'):
+                warnings.warn("The model argument should be an instance of the"
```
goodfeli: What would you think about just declaring this interface deprecated and issuing a warning that it will be gone in 6 months?

cleverhans/model.py
```diff
@@ -67,3 +67,28 @@ def fprop(self, x):
...
+class CallableModelWrapper(Model):
+    def __init__(self, callable_fn, output_layer):
+        """
+        Wrap a callable function that takes a tensor as input and returns
+        a tensor as output with the given layer name.
+        :param callable_fn: The callable function taking a tensor and
+                            returning a given layer as output.
+        :param output_layer: A string of either "probs" or "logits",
+                             corresponding to the output of the callable.
+        """
```
goodfeli: Do we need to enforce this? We could imagine maybe making a CallableModelWrapper for producing the features for a Sara Sabour-style attack, where we'd want the output to have a different name. If we don't enforce this here, there will still be an error message later saying that the Model doesn't provide the desired layer.

cleverhans/attacks.py
carlini: Sure.

cleverhans/attacks.py
carlini: Sure. If we do 4 months then it would line up with removing theano.

cleverhans/model.py
carlini: Done.

cleverhans/attacks.py
npapernot: My opinion is slightly leaning towards deprecating the interface (6 months should be fine). I see an argument in favor and against:

* in favor: the attacks can now rely on the methods provided by the Model abstraction
* against: the users have to make an extra call to `CallableModelWrapper`

carlini: I'll put a notice we'll deprecate in 6 months. I think that at the end of the day, we want attacks to be able to use any method of Model, some of which might not be implemented if people just give a callable.

cleverhans/model.py
```diff
@@ -67,3 +67,28 @@ def fprop(self, x):
...
+    def __init__(self, callable_fn, output_layer):
+        """
+        Wrap a callable function that takes a tensor as input and returns
+        a tensor as output with the given layer name.
+        :param callable_fn: The callable function taking a tensor and
+                            returning a given layer as output.
+        :param output_layer: A string of the output layer returned by the
+                             function. (Usually either "probs" or "logits".)
+        """
+        if output_layer not in ['probs', 'logits']:
```
npapernot: We should remove this check if we don't enforce that

tests_tf/test_attacks.py
```diff
@@ -45,5 +46,19 @@ def test_generate_np(self):
...
+class TestCallableModelWrapperInitArguments(unittest.TestCase):
+    def test_output_layer(self):
+        def model():
+            return True
+
+        # The following two calls should not raise Exceptions
+        wrap = CallableModelWrapper(model, 'probs')
+        wrap = CallableModelWrapper(model, 'logits')
+
+        with self.assertRaises(Exception) as context:
```
npapernot: This should go away too

npapernot:
LGTM

carlini:
I've rebased this branch, so to push I had to --force it. If you want I can just submit a new PR that's clean.

npapernot:
There is a problem with the rebase (one of the tests should not be there anymore). I pointed this out to Nicholas in real life and he says he'll take care of it later today.

carlini:
Okay, tests are good now.

PR #204, opened 2017-07-14
Fix bug in attack_random_noise.py in adversarial competition toolkit.

AlexeyKurakin:


npapernot:
This is outside of the cleverhans/ or tutorials/ folders, LGTM

PR #205, opened 2017-07-15
Implement new L2 attack

carlini:
New L2 attack implementation from "Towards Evaluating the Robustness of Neural Networks".

Commits are cherry-picked from #156.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -601,19 +601,18 @@ class CarliniWagnerL2(Attack):
 
         if not isinstance(self.model, Model):
             self.model = CallableModelWrapper(self.model, 'logits')
-        
+
     def generate(self, x, **kwargs):
         """
-        Return a tensor that consturcts adversarial examples for the given
+        Return a tensor that constructs adversarial examples for the given
         input. Generate uses tf.py_func in order to operate over tensors.
 
-        :param x: (required) A tensor with the original inputs.
+        :param x: (required) A tensor with the inputs.
         :param y: (optional) A tensor with the true labels for an untargeted
                   attack. If None (and y_target is None) then use the
                   original labels the classifier assigns.
-        :param y_target: (optional) A tensor with the true labels for an 
-                  untargeted attack. If None (and y_target is None) then use 
-                  the original labels the classifier assigns.
+        :param y_target: (optional) A tensor with the target labels for a
+                  targeted attack.
         :param nb_classes: The number of classes the model has.
         :param confidence: Confidence of adversarial examples: higher produces
                            examples with larger l2 distortion, but more
@@ -655,13 +654,12 @@ class CarliniWagnerL2(Attack):
                       self.nb_classes, x.get_shape().as_list()[1:])
 
         if 'y' in kwargs and 'y_target' in kwargs:
-            raise ValueError("Can not send both 'y' and 'y_target'.")
+            raise ValueError("Can not set both 'y' and 'y_target'.")
         elif 'y' in kwargs:
             labels = kwargs['y']
         elif 'y_target' in kwargs:
             labels = kwargs['y_target']
         else:
-            # TODO abstract this out for other classes too
             preds = self.model.get_probs(x)
             preds_max = tf.reduce_max(preds, 1, keep_dims=True)
             original_predictions = tf.to_float(tf.equal(preds,
@@ -681,7 +679,7 @@ class CarliniWagnerL2(Attack):
                      abort_early=True, initial_const=1e-2,
                      clip_min=0, clip_max=1):
 
-        # ignore the y and y_val argument
+        # ignore the y and y_target argument
         self.nb_classes = nb_classes
         self.batch_size = batch_size
         self.confidence = confidence
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -641,7 +641,7 @@ class CarliniWagnerL2(Attack):
 
         from .attacks_tf import CarliniWagnerL2 as CWL2
         self.parse_params(**kwargs)
-        
+
         params = (self.batch_size,
                   self.confidence, self.targeted, self.learning_rate,
                   self.binary_search_steps, self.max_iterations,
@@ -663,11 +663,11 @@ class CarliniWagnerL2(Attack):
         return res
 
     def parse_params(self, y_val=None, nb_classes=10,
-                    batch_size=1, confidence=0,
-                    targeted=True, learning_rate=5e-3,
-                    binary_search_steps=5, max_iterations=1000,
-                    abort_early=True, initial_const=1e-2,
-                    clip_min=0, clip_max=1):
+                     batch_size=1, confidence=0,
+                     targeted=True, learning_rate=5e-3,
+                     binary_search_steps=5, max_iterations=1000,
+                     abort_early=True, initial_const=1e-2,
+                     clip_min=0, clip_max=1):
 
         # ignore the y_val argument
         self.nb_classes = nb_classes
@@ -681,7 +681,7 @@ class CarliniWagnerL2(Attack):
         self.initial_const = initial_const
         self.clip_min = clip_min
         self.clip_max = clip_max
-    
+
 
 def fgsm(x, predictions, eps, back='tf', clip_min=None, clip_max=None):
     """
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -651,18 +651,19 @@ class CarliniWagnerL2(Attack):
         from .attacks_tf import CarliniWagnerL2 as CWL2
 
         params = (batch_size, confidence, targeted,
-                      learning_rate, binary_search_steps, max_iterations,
-                      abort_early, initial_const, clip_min, clip_max,
-                      nb_classes, tuple(x_val.shape[1:]))
+                  learning_rate, binary_search_steps, max_iterations,
+                  abort_early, initial_const, clip_min, clip_max,
+                  nb_classes, tuple(x_val.shape[1:]))
         if params in self.attack_objects:
             attack = self.attack_objects[params]
         else:
-            attack = CWL2(self.sess, self.model, batch_size, confidence, targeted,
-                          learning_rate, binary_search_steps, max_iterations,
-                          abort_early, initial_const, clip_min, clip_max,
-                          nb_classes, tuple(x_val.shape[1:]))
+            attack = CWL2(self.sess, self.model, batch_size, confidence,
+                          targeted, learning_rate, binary_search_steps,
+                          max_iterations, abort_early, initial_const,
+                          clip_min, clip_max, nb_classes,
+                          tuple(x_val.shape[1:]))
             self.attack_objects[params] = attack
-        
+
         res = attack.attack(x_val, y_val)
         return res
 
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -110,12 +110,12 @@ class Attack(object):
 
     def generate_np(self, x_val, **kwargs):
         """
-        Generate adversarial examples and return them as a Numpy array.
+        Generate adversarial examples and return them as a NumPy array.
         Sub-classes *should not* implement this method unless they must
         perform special handling of arguments.
-        :param x_val: A Numpy array with the original inputs.
+        :param x_val: A NumPy array with the original inputs.
         :param **kwargs: optional parameters used by child classes.
-        :return: A Numpy array holding the adversarial examples.
+        :return: A NumPy array holding the adversarial examples.
         """
         if self.back == 'th':
             raise NotImplementedError('Theano version not implemented.')
@@ -200,7 +200,7 @@ class FastGradientMethod(Attack):
         Generate symbolic graph for adversarial examples and return.
         :param x: The model's symbolic inputs.
         :param eps: (optional float) attack step size (input variation)
-        :param ord: (optional) Order of the norm (mimics Numpy).
+        :param ord: (optional) Order of the norm (mimics NumPy).
                     Possible values: np.inf, 1 or 2.
         :param y: (optional) A tensor with the model labels. Only provide
                   this parameter if you'd like to use true labels when crafting
@@ -240,7 +240,7 @@ class FastGradientMethod(Attack):
 
         Attack-specific parameters:
         :param eps: (optional float) attack step size (input variation)
-        :param ord: (optional) Order of the norm (mimics Numpy).
+        :param ord: (optional) Order of the norm (mimics NumPy).
                     Possible values: np.inf, 1 or 2.
         :param y: (optional) A tensor with the model labels. Only provide
                   this parameter if you'd like to use true labels when crafting
@@ -380,7 +380,7 @@ class BasicIterativeMethod(Attack):
                     compared to original input
         :param eps_iter: (required float) step size for each attack iteration
         :param nb_iter: (required int) Number of attack iterations.
-        :param y: (required) A tensor with the model labels.
+        :param y: (optional) A tensor with the model labels.
         :param y_target: (optional) A tensor with the labels to target. Leave
                          y_target=None if y is also set. Labels should be
                          one-hot-encoded.
@@ -578,6 +578,10 @@ class CarliniWagnerL2(Attack):
     are robust to other attacks.
     Paper link: https://arxiv.org/abs/1608.04644
 
+    At a high level, this attack is an iterative attack using Adam and
+    a specially-chosen loss function to find adversarial examples with
+    lower distortion than other attacks. This comes at the cost of speed,
+    as this attack is often much slower than others.
     """
     def __init__(self, model, back='tf', sess=None):
         super(CarliniWagnerL2, self).__init__(model, back, sess)
@@ -597,35 +601,41 @@ class CarliniWagnerL2(Attack):
 
     def generate(self, x, **kwargs):
         """
-        Generate adversarial samples and return them in a Numpy array.
+        Return a tensor that consturcts adversarial examples for the given
+        input. Generate uses tf.py_func in order to operate over tensors.
 
-        :param x: (required) A Numpy array with the original inputs.
-        :param y: (optional) A Numpy array with the labels that we either
+        :param x: (required) A tensor with the original inputs.
+        :param y: (optional) A tensor with the labels that we either
                   should target (if targeted=True) or avoid (if
-                  target=False). If None, use labels the classifier assigns.
+                  targeted=False). If None, use labels the classifier assigns.
         :param nb_classes: The number of classes the model has.
         :param confidence: Confidence of adversarial examples: higher produces
-                           examples that are farther away, but more strongly
-                           classified as adversarial.
+                           examples with larger l2 distortion, but more 
+                           strongly classified as adversarial.
         :param batch_size: Number of attacks to run simultaneously.
-        :param targeted: True if we should perform a targetted attack, False
+        :param targeted: True if we should perform a targeted attack, False
                          otherwise.
         :param learning_rate: The learning rate for the attack algorithm.
                               Smaller values produce better results but are
                               slower to converge.
         :param binary_search_steps: The number of times we perform binary
                                     search to find the optimal tradeoff-
-                                    constant between distance and confidence.
-        :param max_iterations: The maximum number of iterations. Larger values
-                               are more accurate; setting too small will
-                               require a large learning rate and will produce
-                               poor results.
+                                    constant between norm of the purturbation
+                                    and confidence of the classification.
+        :param max_iterations: The maximum number of iterations. Setting this
+                               to a larger value will produce lower distortion
+                               results. Using only a few iterations requires
+                               a larger learning rate, and will produce larger
+                               distortion results.
         :param abort_early: If true, allows early aborts if gradient descent
-                            gets stuck.
+                            is unable to make progress (i.e., gets stuck in
+                            a local minimum).
         :param initial_const: The initial tradeoff-constant to use to tune the
-                              relative importance of distance and confidence.
+                              relative importance of size of the pururbation 
+                              and confidence of classification.
                               If binary_search_steps is large, the initial
-                              constant is not important.
+                              constant is not important. A smaller value of
+                              this constant gives lower distortion results.
         :param clip_min: (optional float) Minimum input component value
         :param clip_max: (optional float) Maximum input component value
         """
@@ -647,7 +657,7 @@ class CarliniWagnerL2(Attack):
                 raise ValueError("Must supply target labels 'y'"
                                  " in targeted attack.")
             # TODO abstract this out for other classes too
-            preds = self.model(x)
+            preds = self.model.get_probs(x)
             preds_max = tf.reduce_max(preds, 1, keep_dims=True)
             original_predictions = tf.to_float(tf.equal(preds,
                                                         preds_max))
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -587,7 +587,7 @@ class CarliniWagnerL2(Attack):
 
         import tensorflow as tf
         self.feedable_kwargs = {'y': tf.float32}
-        
+
         self.structural_kwargs = ['nb_classes',
                                   'batch_size', 'confidence',
                                   'targeted', 'learning_rate',
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -636,15 +636,16 @@ class CarliniWagnerL2(Attack):
         if 'y' in kwargs:
             labels = kwargs['y']
         else:
-            if self.targeted == True:
-                raise ValueError("Must supply target labels in targeted attack.")
+            if self.targeted:
+                raise ValueError("Must supply target labels 'y'"
+                                 " in targeted attack.")
             # TODO abstract this out for other classes too
             preds = self.model(x)
             preds_max = tf.reduce_max(preds, 1, keep_dims=True)
             original_predictions = tf.to_float(tf.equal(preds,
                                                         preds_max))
             labels = original_predictions
-            
+
         def cw_wrap(x_val, y_val):
             return np.array(attack.attack(x_val, y_val), dtype=np.float32)
         wrap = tf.py_func(cw_wrap, [x, labels], tf.float32)
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -633,10 +633,22 @@ class CarliniWagnerL2(Attack):
                       self.clip_min, self.clip_max, self.nb_classes,
                       x.get_shape().as_list()[1:])
 
+        if 'y' in kwargs:
+            labels = kwargs['y']
+        else:
+            if self.targeted == True:
+                raise ValueError("Must supply target labels in targeted attack.")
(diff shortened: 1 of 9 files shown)


cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+        self.structural_kwargs = ['nb_classes',
+                                  'batch_size', 'confidence',
+                                  'targeted', 'learning_rate',
+                                  'binary_search_steps', 'max_iterations',
+                                  'abort_early', 'initial_const',
+                                  'clip_min', 'clip_max']
+
+    def generate(self, x, **kwargs):
+        """
+        Generate adversarial samples and return them in a Numpy array.
```
npapernot: The `generate` method should return a tensor, this comment is likely to be misleading for users (even if this implementation is wrapped).

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+
+        :param x: (required) A Numpy array with the original inputs.
```
npapernot: See above

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+        :param y: (optional) A Numpy array with the labels that we either
```
npapernot: See above

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+                  should target (if targeted=True) or avoid (if
+                  target=False). If None, use labels the classifier assigns.
+        :param nb_classes: The number of classes the model has.
```
npapernot: Could this be inferred from the model object? (to remove the number of arguments that users need to pass.

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+        :param confidence: Confidence of adversarial examples: higher produces
+                           examples that are farther away, but more strongly
```
npapernot: "farther away" is somewhat vague. Could we instead say something along the lines of "perturbations with a higher L2 norm"

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
         return True
 
 
+class CarliniWagnerL2(Attack):
```
npapernot: Given that you will make PRs for the L0 and Linf attacks, is it possible to make the class general enough, that it can take a norm argument and call the right underlying algorithm?  Or is that not possible because the attacks have very different interfaces for different norms?

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+        :param x: (required) A Numpy array with the original inputs.
+        :param y: (optional) A Numpy array with the labels that we either
+                  should target (if targeted=True) or avoid (if
+                  target=False). If None, use labels the classifier assigns.
+        :param nb_classes: The number of classes the model has.
+        :param confidence: Confidence of adversarial examples: higher produces
+                           examples that are farther away, but more strongly
+                           classified as adversarial.
+        :param batch_size: Number of attacks to run simultaneously.
+        :param targeted: True if we should perform a targetted attack, False
```
npapernot: Is the spelling for targetted, targeted or targetted? (I'm actually not sure :) )

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+                         otherwise.
+        :param learning_rate: The learning rate for the attack algorithm.
+                              Smaller values produce better results but are
+                              slower to converge.
+        :param binary_search_steps: The number of times we perform binary
+                                    search to find the optimal tradeoff-
+                                    constant between distance and confidence.
```
npapernot: Would "norm of the perturbation" and "confidence of the misclassification" make that comment easier to understand for newcomers?

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+        :param max_iterations: The maximum number of iterations. Larger values
+                               are more accurate; setting too small will
+                               require a large learning rate and will produce
```
npapernot: It sounds like something is missing from the comparison. Do you mean that to compensate with the smaller number of iterations, the optimization has to take large steps at each iteration (and hence a larger learning rate) to be as likely to succeed (even though in practice it doesn't)?

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+                  should target (if targeted=True) or avoid (if
+                  target=False). If None, use labels the classifier assigns.
+        :param nb_classes: The number of classes the model has.
+        :param confidence: Confidence of adversarial examples: higher produces
+                           examples that are farther away, but more strongly
+                           classified as adversarial.
+        :param batch_size: Number of attacks to run simultaneously.
+        :param targeted: True if we should perform a targetted attack, False
+                         otherwise.
+        :param learning_rate: The learning rate for the attack algorithm.
```
npapernot: Is learning rate the best name to call this parameter? I am wondering if this will be confusing for new users (because of the learning rate used during training)? 

I don't have a strong opinion however.

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+                              slower to converge.
+        :param binary_search_steps: The number of times we perform binary
+                                    search to find the optimal tradeoff-
+                                    constant between distance and confidence.
+        :param max_iterations: The maximum number of iterations. Larger values
+                               are more accurate; setting too small will
+                               require a large learning rate and will produce
+                               poor results.
+        :param abort_early: If true, allows early aborts if gradient descent
+                            gets stuck.
```
npapernot: Can you define a little more precisely what you mean by "get stuck". I suppose it means that the gradients are vanishing for several iterations?

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+        :param initial_const: The initial tradeoff-constant to use to tune the
+                              relative importance of distance and confidence.
```
npapernot: See above (regarding "distance" and "confidence")

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+                      x.get_shape().as_list()[1:])
+
+        if 'y' in kwargs:
+            labels = kwargs['y']
+        else:
+            if self.targeted:
+                raise ValueError("Must supply target labels 'y'"
+                                 " in targeted attack.")
+            # TODO abstract this out for other classes too
+            preds = self.model(x)
```
npapernot: Should this be `model.get_probs()` with the new interface?

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+        return wrap
+
+    def parse_params(self, y=None, y_val=None, nb_classes=10,
+                     batch_size=1, confidence=0,
+                     targeted=True, learning_rate=5e-3,
+                     binary_search_steps=5, max_iterations=1000,
+                     abort_early=True, initial_const=1e-2,
+                     clip_min=0, clip_max=1):
+
+        # ignore the y and y_val argument
```
npapernot: Could you elaborate on why these two arguments are ignored? Not necessarily in the code, but in this thread at least.

cleverhans/attacks.py
goodfeli: Not important enough to delay merge, but preferred capitalization seems to be NumPy

cleverhans/attacks_tf.py
```diff
@@ -417,3 +417,245 @@ def jacobian_augmentation(sess, x, X_sub_prev, Y_sub, grads, lmbda,
 
     # Return augmented training data (needs to be labeled afterwards)
     return X_sub
+
+
+class CarliniWagnerL2:
```
npapernot: I think we already talked about that, but I'm wondering how we should organize the code here. This is the first attack that has an underlying class object. My question is do we want to eventually merge this `CarliniWagnerL2` class in `attacks_tf.py` w/ the wrapper class `CarliniWagnerL2` in `attacks.py`.

The reason this might make sense is that once we drop support for Theano in a few months, the wrapper classes in `attacks.py` will be pretty bare bones, so putting the code all in one place might make things easier to maintain (we could also apply the same logic to the other attacks).

Not sure what is the best way to go, @goodfeli do you have any thoughts?

cleverhans/attacks_tf.py
```diff
@@ -417,3 +417,245 @@ def jacobian_augmentation(sess, x, X_sub_prev, Y_sub, grads, lmbda,
...
+    def __init__(self, sess, model, batch_size, confidence,
+                 targeted, learning_rate,
+                 binary_search_steps, max_iterations,
+                 abort_early, initial_const,
+                 clip_min, clip_max, num_labels, shape):
+        """
+        This attack was originally proposed by Carlini and Wagner. It is an
+        iterative attack that finds adversarial examples on many defenses that
```
npapernot: See above

cleverhans/attacks.py
goodfeli: It seems kind of dangerous to have this option take on almost opposite meanings based on the values of other arguments.
Would it make sense to have target be either None or a tensor containing the target? Then raise an error if both y and target are specified.

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
         return True
 
 
+class CarliniWagnerL2(Attack):
+    """
+    This attack was originally proposed by Carlini and Wagner. It is an
+    iterative attack that finds adversarial examples on many defenses that
```
npapernot: Could you add a high-level TL;DR of what the attack is doing? Something like running Adam on customized loss functions

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+                                  'clip_min', 'clip_max']
+
+    def generate(self, x, **kwargs):
+        """
+        Generate adversarial samples and return them in a Numpy array.
+
+        :param x: (required) A Numpy array with the original inputs.
+        :param y: (optional) A Numpy array with the labels that we either
+                  should target (if targeted=True) or avoid (if
+                  target=False). If None, use labels the classifier assigns.
```
goodfeli: should be "targeted" not "target" if keeping the current names

cleverhans/attacks_tf.py
```diff
@@ -417,3 +417,245 @@ def jacobian_augmentation(sess, x, X_sub_prev, Y_sub, grads, lmbda,
...
+        are robust to other attacks.
+        Paper link: https://arxiv.org/abs/1608.04644
+
+        :param confidence: Confidence of adversarial examples: higher produces
+                           examples that are farther away, but more strongly
+                           classified as adversarial.
+        :param batch_size: Number of attacks to run simultaneously.
+        :param targeted: True if we should perform a targetted attack, False
```
npapernot: Typo? See above

cleverhans/attacks_tf.py
```diff
@@ -417,3 +417,245 @@ def jacobian_augmentation(sess, x, X_sub_prev, Y_sub, grads, lmbda,
...
+                         otherwise.
+        :param learning_rate: The learning rate for the attack algorithm.
```
npapernot: See above, and this probably applies to most of the docstring since it is shared w/ the wrapper class in `attacks.py`

cleverhans/attacks.py
```diff
@@ -514,6 +514,115 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+        :param nb_classes: The number of classes the model has.
+        :param confidence: Confidence of adversarial examples: higher produces
```
goodfeli: Is this a minimum confidence level after which the optimization stops? Or a target confidence level the optimization aims at?

cleverhans/attacks.py
goodfeli: typo "targetted"

tests_tf/test_mnist_tutorial_jsma.py
```diff
@@ -16,7 +16,7 @@ def test_mnist_tutorial_jsma(self):
         jsma_tutorial_args = {'train_start': 0,
                               'train_end': 1000,
                               'test_start': 0,
-                              'test_end': 1666,
+                              'test_end': 166,
```
npapernot: Why was this changed?

npapernot:
Thanks a lot for the PR Nicholas, I left a few comments for the first part of the code and will take a look again when you got a chance to make the edits (but I suspect the other files are more straightforward from the quick glance I took).

cleverhans/attacks_tf.py
goodfeli: What's the purpose of having two different classes?

Also, if we keep the two-class version, I suspect this one should inherit from `object` to make it a "new-style class".

tests_tf/test_attacks.py
```diff
@@ -112,7 +113,7 @@ def test_generate_np(self):
         perturbation_norm = np.sqrt(np.sum(perturbation**2, axis=1))
         # test perturbation norm
         self.assertTrue(np.allclose(perturbation_norm, self.attack.eps))
-
+
```
goodfeli: probably better to not add this but not a big deal

tutorials/mnist_tutorial_cw.py
```diff
@@ -0,0 +1,226 @@
...
+    # Instantiate a CW attack object
+    cw = CarliniWagnerL2(model, back='tf', sess=sess)
+
+    idxs = [np.where(np.argmax(Y_test, axis=1) == i)[0][0] for i in range(10)]
+    if targeted:
+        # Initialize our array for grid visualization
+        grid_shape = (nb_classes, nb_classes, img_rows, img_cols, channels)
+        grid_viz_data = np.zeros(grid_shape, dtype='f')
+
+        onehot = np.zeros((10, 10))
```
goodfeli: maybe `one_hot` just for consistent pep8 naming scheme / matching tf.one_hot

cleverhans/attacks.py
carlini: Okay, makes sense. Should we standardize on the name of the y values we supply? JSMA calls it "targets", FGM calls it "y". Should I use "targets" for the targeted mode and "y" for untargeted?

cleverhans/attacks.py
carlini: Yeah, there are different interfaces (e.g., there's no binary search in l0/li), but those have other constants l2 doesn't.

cleverhans/attacks.py
carlini: Ah sorry, yeah, I copied the docstring and didn't update this.

cleverhans/attacks.py
carlini: Apparently neither was I. But targeted is right.

cleverhans/attacks.py
carlini: I don't have a better word, do you? I use learning_rate because that's what it's used for. step_size isn't entirely accurate since adam does adaptive step sizes for a given learning rate.

cleverhans/attacks.py
carlini: What I mean to say: If the number of iterations is small, then the learning rate must be high. This in turn makes the distance to adversarial examples higher than it must be.

I'll try to edit the docs.

cleverhans/attacks.py
carlini: Will do. That's what I mean.

cleverhans/attacks.py
carlini: Because I specifically handle them in the generate() method, and don't do anything with it in the parse_params. I might be able to restructure the code to do it in parse_params() now that generate_np() is implemented in the Attack class.

tests_tf/test_mnist_tutorial_jsma.py
carlini: By accident -- cherrypicked it it seems.

cleverhans/attacks_tf.py
carlini: Yeah, I should remove the second class. It was a class so that I could easily deal with graph caching, but now that generate_np is rewritten, I should clean this up.

cleverhans/attacks.py
carlini: It targets specific the confidence that you set. So if you set 0, it makes it just cross the decision boundary. If you set 5 then it wants exactly 5 (and reports "failure" if it's less than 5).

cleverhans/attacks_tf.py
carlini: I will kill this class. It's only a class because it was convenient when I was doing my own caching in the attack, but with the generate_np rewrite it should just be its own method.

cleverhans/attacks.py
carlini: Done in f8feb2e

cleverhans/attacks.py
npapernot: Ok, let's keep it that way in that case

cleverhans/attacks.py
npapernot: I don't have a better suggestion, let's just keep it that way then. The rest of the sentence should make it clear enough.

cleverhans/attacks.py
npapernot: Ok thanks!

cleverhans/attacks_tf.py
npapernot: That sounds good!

cleverhans/attacks.py
```diff
@@ -540,35 +544,41 @@ def __init__(self, model, back='tf', sess=None):
...
+                               results. Using only a few iterations requires
+                               a larger learning rate, and will produce larger
+                               distortion results.
         :param abort_early: If true, allows early aborts if gradient descent
-                            gets stuck.
+                            is unable to make progress (i.e., gets stuck in
+                            a local minimum).
         :param initial_const: The initial tradeoff-constant to use to tune the
-                              relative importance of distance and confidence.
+                              relative importance of size of the pururbation
```
npapernot: There is a typo

npapernot:
The tests are failing because of the two following `pep8` errors:

```
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:556:72: W291 trailing whitespace
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:577:77: W291 trailing whitespace
```

carlini:
I have a fix for those pep8 errors, along with removing the attacks_tf.py extra class to make the code cleaner. If not done tonight, will be done tomorrow.

goodfeli:
Re standardizing on "targeted", "y", etc., do we maybe want to make targeted and non-targeted attacks official part of the Attack interface?

cleverhans/attacks_tf.py
carlini: Okay, so I think that actually having the second class is useful. Since the attack is implemented in python, and generate() is implemented with a py_func, if the py_func created a second instance of the class, then the pseudo-code

    args = {...}
    xs = CarliniWagnerL2().generate(...)
    a = sess.run(xs, args)
    b = sess.run(xs, args)

would result in constructing the graph twice. JSMA has this problem too, but gets around it by placing a call to

    grads = jacobian_graph(preds, x, self.nb_classes)  

outside of the py_func. This is equivelant to what I do now, which is instantiate the second class outside of the py_func. In my case, it's easier to have a second class instead of a helper function because there are so many elements of the graph that are re-used and just saved as instance variables.

carlini: See discussion under Ian's comment.

carlini:
Rebase seems to have slightly confused googlebot.

Unless there's anything else, I'm happy with the current state of the commit. The final possible change I have in mind would be to remove nb_classes and just figure it out based on the model, but I think I will do that in a separate PR so I can fix JSMA too.

npapernot:
Yes, let's keep the nb_classes removal for another PR. 

I'll take another pass at this one

npapernot:
And yes the CLA bot bugs when someone rebases through the online interface, but we can still merge so it's OK

cleverhans/attacks.py
```diff
@@ -571,6 +570,130 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+                                  'binary_search_steps', 'max_iterations',
+                                  'abort_early', 'initial_const',
+                                  'clip_min', 'clip_max']
+
+        if not isinstance(self.model, Model):
+            self.model = CallableModelWrapper(self.model, 'logits')
+        
+    def generate(self, x, **kwargs):
+        """
+        Return a tensor that consturcts adversarial examples for the given
```
npapernot: There is a typo

cleverhans/attacks.py
```diff
@@ -571,6 +570,130 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+        input. Generate uses tf.py_func in order to operate over tensors.
+
+        :param x: (required) A tensor with the original inputs.
+        :param y: (optional) A tensor with the true labels for an untargeted
+                  attack. If None (and y_target is None) then use the
+                  original labels the classifier assigns.
```
npapernot: Could we remove "original", it sounds like we are talking about the labels from the dataset?

cleverhans/attacks.py
```diff
@@ -571,6 +570,130 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+        :param y_target: (optional) A tensor with the true labels for an
```
npapernot: This seems to have been copied over

cleverhans/attacks.py
```diff
@@ -571,6 +570,130 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+
+        attack = CWL2(self.sess, self.model, self.batch_size,
+                      self.confidence, 'y_target' in kwargs,
+                      self.learning_rate, self.binary_search_steps,
+                      self.max_iterations, self.abort_early,
+                      self.initial_const, self.clip_min, self.clip_max,
+                      self.nb_classes, x.get_shape().as_list()[1:])
+
+        if 'y' in kwargs and 'y_target' in kwargs:
+            raise ValueError("Can not send both 'y' and 'y_target'.")
```
npapernot: send -> set

cleverhans/attacks.py
```diff
@@ -571,6 +570,130 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+        elif 'y' in kwargs:
+            labels = kwargs['y']
+        elif 'y_target' in kwargs:
+            labels = kwargs['y_target']
+        else:
+            # TODO abstract this out for other classes too
```
npapernot: Could you raise an issue instead and remove the comment? It's easier to track what's open in the issue tracker

cleverhans/attacks.py
```diff
@@ -571,6 +570,130 @@ def parse_params(self, eps=2.0, num_iterations=1, xi=1e-6, clip_min=None,
...
+        return wrap
+
+    def parse_params(self, y=None, y_target=None, nb_classes=10,
+                     batch_size=1, confidence=0,
+                     learning_rate=5e-3,
+                     binary_search_steps=5, max_iterations=1000,
+                     abort_early=True, initial_const=1e-2,
+                     clip_min=0, clip_max=1):
+
+        # ignore the y and y_val argument
```
npapernot: Anything new related to this?

cleverhans/attacks_tf.py
```diff
@@ -437,3 +437,245 @@ def jacobian_augmentation(sess, x, X_sub_prev, Y_sub, grads, lmbda,
 
     # Return augmented training data (needs to be labeled afterwards)
     return X_sub
+
+
+class CarliniWagnerL2:
```
npapernot: Anything new related to @goodfeli 's comment on inheriting from `object`?

cleverhans/attacks_tf.py
```diff
@@ -437,3 +437,245 @@ def jacobian_augmentation(sess, x, X_sub_prev, Y_sub, grads, lmbda,
...
+        This attack was originally proposed by Carlini and Wagner. It is an
+        iterative attack that finds adversarial examples on many defenses that
+        are robust to other attacks.
+        Paper link: https://arxiv.org/abs/1608.04644
+
+        :param confidence: Confidence of adversarial examples: higher produces
+                           examples that are farther away, but more strongly
+                           classified as adversarial.
+        :param batch_size: Number of attacks to run simultaneously.
+        :param targeted: True if we should perform a targetted attack, False
```
npapernot: The typo is still here

cleverhans/attacks_tf.py
```diff
@@ -437,3 +437,245 @@ def jacobian_augmentation(sess, x, X_sub_prev, Y_sub, grads, lmbda,
...
+                         otherwise.
+        :param learning_rate: The learning rate for the attack algorithm.
```
npapernot: Was the docstring updated to match the Attacks class?

tests_tf/test_attacks.py
```diff
@@ -275,6 +276,157 @@ def fn(*x, **y):
         self.assertTrue(ok[0])
 
 
+class TestCarliniWagnerL2(CleverHansTest):
+    def setUp(self):
```
npapernot: The np seed seems to be missing

npapernot:
I left a few comments, but looks like it's ready to merge afterwards

npapernot:
There are also some pep8 errors left:

```
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:604:1: W293 blank line contains whitespace
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:614:73: W291 trailing whitespace
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:615:77: W291 trailing whitespace
```

cleverhans/attacks.py
carlini: I commented above about this.

tests_tf/test_attacks.py
carlini: It's done in the super class.

PR #207, opened 2017-07-18
added func arguments description

IvanKush:
added method arguments description

PR #210, opened 2017-07-21
added func arguments description

IvanKush:
added method arguments description and changed description from `placeholder` to `tensor` in label according to comments from this closed uncommited pull https://github.com/tensorflow/cleverhans/pull/207


IvanKush:
I signed it!

PR #214, opened 2017-07-29
Targeted attack kwargs

carlini:
As discussed on the mailing list, use better keyword arguments for targeted attacks. This also implements targeted attacks for FGM and BIM.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -119,9 +119,6 @@ class Attack(object):
         """
         if self.back == 'th':
             raise NotImplementedError('Theano version not implemented.')
-
-        import tensorflow as tf
-
         if self.sess is None:
             raise ValueError("Cannot use `generate_np` when no `sess` was"
                              " provided")
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -190,6 +190,7 @@ class FastGradientMethod(Attack):
         super(FastGradientMethod, self).__init__(model, back, sess)
         self.feedable_kwargs = {'eps': np.float32,
                                 'y': np.float32,
+                                'y_target': np.float32,
                                 'clip_min': np.float32,
                                 'clip_max': np.float32}
         self.structural_kwargs = ['ord']
@@ -221,12 +222,14 @@ class FastGradientMethod(Attack):
         else:
             from .attacks_th import fgm
 
-        return fgm(x, self.model.get_probs(x), y=self.y, eps=self.eps,
-                   ord=self.ord, clip_min=self.clip_min,
-                   clip_max=self.clip_max)
+        y = self.y or self.y_target
 
-    def parse_params(self, eps=0.3, ord=np.inf, y=None, clip_min=None,
-                     clip_max=None, **kwargs):
+        return fgm(x, self.model.get_probs(x), y=y, eps=self.eps,
+                   ord=self.ord, clip_min=self.clip_min,
+                   clip_max=self.clip_max, targeted=(self.y_target != None))
+
+    def parse_params(self, eps=0.3, ord=np.inf, y=None, y_target=None,
+                     clip_min=None, clip_max=None, **kwargs):
         """
         Take in a dictionary of parameters and applies attack-specific checks
         before saving them as attributes.
@@ -241,16 +244,23 @@ class FastGradientMethod(Attack):
                   labels to avoid the "label leaking" effect (explained in this
                   paper: https://arxiv.org/abs/1611.01236). Default is None.
                   Labels should be one-hot-encoded.
+        :param y_target: (optional) A tensor with the labels to target. Do not
+                         set y_target if y is also set. Labels should be 
+                         one-hot-encoded.
         :param clip_min: (optional float) Minimum input component value
         :param clip_max: (optional float) Maximum input component value
         """
         # Save attack-specific parameters
+            
         self.eps = eps
         self.ord = ord
         self.y = y
+        self.y_target = y_target
         self.clip_min = clip_min
         self.clip_max = clip_max
 
+        if self.y != None and self.y_target != None:
+            raise ValueError("Must not set both y and y_target")
         # Check if order of the norm is acceptable given current implementation
         if self.ord not in [np.inf, int(1), int(2)]:
             raise ValueError("Norm order must be either np.inf, 1, or 2.")
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -397,7 +397,7 @@ class SaliencyMapMethod(Attack):
             raise NotImplementedError(error)
 
         import tensorflow as tf
-        self.feedable_kwargs = {'targets': tf.float32}
+        self.feedable_kwargs = {'y_target': tf.float32}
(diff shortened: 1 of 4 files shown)


cleverhans/attacks.py
```diff
@@ -308,14 +328,25 @@ def generate(self, x, **kwargs):
         # Fix labels to the first model predictions for loss computation
         model_preds = self.model.get_probs(x)
         preds_max = tf.reduce_max(model_preds, 1, keep_dims=True)
-        y = tf.to_float(tf.equal(model_preds, preds_max))
```
carlini: This was actually a bug in the old code -- it would never read the self.y params.

cleverhans/attacks.py
```diff
@@ -210,6 +211,9 @@ def generate(self, x, **kwargs):
                   labels to avoid the "label leaking" effect (explained in this
                   paper: https://arxiv.org/abs/1611.01236). Default is None.
                   Labels should be one-hot-encoded.
+        :param y_target: (optional) A tensor with the labels to target. Do not
+                         set y_target if y is also set. Labels should be
```
npapernot: Maybe we can make the comment more useful by saying: If you want to run an untargeted (i.e., misclassification) attack while avoiding the label leaking effect, set y to the true labels and y_target to None, if you want to run a targeted (i.e., source-target misclassification attack), set y to None and y_target to the adversary's targets

cleverhans/attacks.py
```diff
@@ -241,16 +250,23 @@ def parse_params(self, eps=0.3, ord=np.inf, y=None, clip_min=None,
...
+                         one-hot-encoded.
```
npapernot: Can we include an assert to check that labels are one-hot encoded in the code?

cleverhans/attacks.py
```diff
@@ -291,7 +308,10 @@ def generate(self, x, **kwargs):
                     compared to original input
         :param eps_iter: (required float) step size for each attack iteration
         :param nb_iter: (required int) Number of attack iterations.
-        :param y: (required) A tensor with the model labels.
+        :param y: (optional) A tensor with the model labels.
+        :param y_target: (optional) A tensor with the labels to target. Do not
+                         set y_target if y is also set. Labels should be
```
npapernot: same than comment above on the docstring, could be more informative

cleverhans/attacks.py
```diff
@@ -351,6 +383,9 @@ def parse_params(self, eps=0.3, eps_iter=0.05, nb_iter=10, y=None,
         :param eps_iter: (required float) step size for each attack iteration
         :param nb_iter: (required int) Number of attack iterations.
         :param y: (required) A tensor with the model labels.
+        :param y_target: (optional) A tensor with the labels to target. Do not
+                         set y_target if y is also set. Labels should be
```
npapernot: same than above regarding the docstring

cleverhans/attacks_tf.py
```diff
@@ -21,12 +21,15 @@ def fgsm(x, predictions, eps=0.3, clip_min=None, clip_max=None):
...
-def fgm(x, preds, y=None, eps=0.3, ord=np.inf, clip_min=None, clip_max=None):
+def fgm(x, preds, y=None, eps=0.3, ord=np.inf,
+        clip_min=None, clip_max=None,
+        targeted=False):
     """
     TensorFlow implementation of the Fast Gradient Method.
     :param x: the input placeholder
     :param preds: the model's output tensor
-    :param y: (optional) A placeholder for the model labels. Only provide
+    :param y: (optional) A placeholder for the model labels. If targeted
```
npapernot: I am wondering whether we should rename this parameter `y_target` and deprecate the `y` and `targeted` parameters for consistence w/ the class in `attacks.py`.

tests_tf/test_attacks.py
```diff
@@ -158,6 +158,25 @@ def test_generate_np_gives_adversarial_example_l1(self):
     def test_generate_np_gives_adversarial_example_l2(self):
         self.help_generate_np_gives_adversarial_example(2)
 
+    def test_targeted_generate_np_gives_adversarial_example(self):
```
npapernot: Thanks for adding a test case, much appreciated!

tests_tf/test_attacks.py
```diff
@@ -158,6 +158,25 @@ def test_generate_np_gives_adversarial_example_l1(self):
...
+        x_val = np.random.rand(100, 2)
+        x_val = np.array(x_val, dtype=np.float32)
+        random_labs = np.random.random_integers(0,1,100)
```
npapernot: I think we already had a discussion about this, but it seems like this test does not have a seed for np.random. It might become flaky?

npapernot:
Here's the PEP8 checker output so you don't have to open Travis:

```
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:235:68: E711 comparison to None should be 'if cond is not None:'
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:268:19: E711 comparison to None should be 'if cond is not None:'
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:347:39: E127 continuation line over-indented for visual indent
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:405:19: E711 comparison to None should be 'if cond is not None:'
--------------------- >> end captured stdout << ----------------------
```

cleverhans/attacks.py
carlini: This wouldn't be that easy. If people use generate, then y_target is a tensor, and so we'd have to put an assert in the middle of the TF graph.

cleverhans/attacks_tf.py
carlini: The y parameter here is for the untargeted attack.

tests_tf/test_attacks.py
carlini: Done in 61dd358858264ebd8d1e63465ccf0346ef27672b.

cleverhans/attacks.py
carlini: I've slightly improved it to the first version. If you want to try to improve it more feel free to commit to the branch.

carlini:
Pending test results, this is ready to merge.

cleverhans/attacks_tf.py
npapernot: Well I guess it's for both untargeted and targeted, so it's fine to leave it like that

cleverhans/attacks_tf.py
```diff
@@ -332,7 +341,7 @@ def jsma(sess, x, predictions, grads, sample, target, theta, gamma, clip_min,
 
 
 def jsma_batch(sess, x, pred, grads, X, theta, gamma, clip_min, clip_max,
-               nb_classes, targets=None, feed=None):
+               nb_classes, y_target=None, feed=None):
```
npapernot: We should add a deprecation warning here, I'll take care of it

tests_tf/test_attacks.py
npapernot: Thanks

npapernot:
(I fixed the Pep8 as well) will push soon

PR #215, opened 2017-08-01
allow running the JSMA with extra feed dict params

asross:
This should let users run JSMA attacks (and all others using generate_np) on models that require more `feed_dict` parameters than just inputs.

Note: I'm using the attacks_tf helpers rather than the abstract attacks interface since it doesn't seem to be compatible with the rest of my code, so I only added a test for that case, but it might be good to add more tests to verify it works with the main interface.

### DIFF
--- a/tests_tf/test_attacks_tf.py
+++ b/tests_tf/test_attacks_tf.py
@@ -6,7 +6,6 @@ from __future__ import unicode_literals
 import numpy as np
 import unittest
 import tensorflow as tf
-import tensorflow.contrib.slim as slim
 
 from cleverhans import attacks_tf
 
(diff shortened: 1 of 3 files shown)


npapernot:
Thanks for the PR Andrew. It is best to discuss new features like this on `cleverhans-dev` first. 

A few things:

* I don't believe there is a need to add a feed_dict to `generate` because it does not run a session
* it would probably be best to include the feed_dict argument in the `**kwargs`

If you won't test the feed argument for the Attacks interface, it may be safer to just include it in the `attacks_tf` functions for now. This would also limit conflicts with #205 until it is merged.

asross:
Sure, sounds good! I'll slim down the PR so it just affects `attacks_tf`, which solves my problem.

tests_tf/test_attacks_tf.py
```diff
@@ -0,0 +1,33 @@
...
+
+class TestAttacksTF(unittest.TestCase):
+    def test_jsma_batch_with_feed(self):
+        with tf.Session() as sess:
+            X = np.random.rand(1, 13)
+
+            # construct a simple graph that will require extra placeholders
+            x = tf.placeholder('float', shape=(None, 13))
+            b = tf.placeholder('bool')
+            logits = slim.dropout(slim.fully_connected(x, 10),
```
npapernot: Could you do this without adding a dependency on slim? Since it's in the `tensorflow.contrib`, its API is more likely to change quickly and it's not necessary to test the logic here.

tests_tf/test_attacks_tf.py
```diff
@@ -0,0 +1,33 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import numpy as np
+import unittest
+import tensorflow as tf
+import tensorflow.contrib.slim as slim
```
npapernot: See comment below at line 21

asross:
> OK to merge once the slim dependency has been removed in the test

Done!

PR #218, opened 2017-08-08
Add clean_train flag (set to False to just do adv train)

goodfeli:


tutorials/mnist_tutorial_tf.py
```diff
@@ -230,56 +235,58 @@ def mnist_tutorial(train_start=0, train_end=60000, test_start=0,
...
-
-    # Calculate training error
-    if testing:
+    if clean_train:
+        def evaluate():
+            # Evaluate the accuracy of the MNIST model on legitimate test
+            # examples
+            eval_params = {'batch_size': batch_size}
+            acc = model_eval(
+                sess, x, y, preds, X_test, Y_test, args=eval_params)
```
npapernot: Is it normal that `preds` is defined after `evaluate()` is?

goodfeli: The code executes correctly, but for legibility it's probably better to define `preds` first, especially since this is a tutorial. I moved it earlier.

npapernot: Sounds good, thanks!

PR #222, opened 2017-08-11
Implement logging framework

carlini:
As discussed earlier, a new logging framework for enabling debug messages during attack generation.

### DIFF
--- a/cleverhans/attacks_tf.py
+++ b/cleverhans/attacks_tf.py
@@ -334,7 +334,8 @@ def jsma(sess, x, predictions, grads, sample, target, theta, gamma, clip_min,
     if current == target:
         _logger.info("Succeeded attack using {} iterations".format(iteration))
     else:
-        _logger.info("Failed to find adversarial example after {} iterations".format(iteration))
+        _logger.info(("Failed to find adversarial example " +
+                      "after {} iterations").format(iteration))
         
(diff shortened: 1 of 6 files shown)


goodfeli:
Looks basically good to me, apart from the tests not passing. Consider starting the Wiki and making a page about logging best practices, to make it easier to get your logging features adopted.

cleverhans/attacks_tf.py
```diff
@@ -576,6 +586,8 @@ def attack(self, imgs, targets):
 
         r = []
         for i in range(0, len(imgs), self.batch_size):
+            _logger.debug(("Running CWL2 attack on image " +
```
npapernot: It's probably best here to replace "image" by "input" here for generality

cleverhans/utils_tf.py
```diff
@@ -153,7 +156,7 @@ def model_train(sess, x, y, predictions, X_train, Y_train, save=False,
 
         for epoch in six.moves.xrange(args.nb_epochs):
             if verbose:
```
npapernot: If we are going to switch all of these prints to the log, it might make sense to deprecate the `verbose` argument and let users disable logging / not access the logs instead

carlini: Done in 74a5195

npapernot:
LGTM once tests pass

PR #223, opened 2017-08-11
Fixed bug in basic iterative attack where clipping was done only once…

fartashf:
… at the end of the attack but it should be done at each step.


goodfeli:
LGTM. Would you mind adding a test? You can make a dummy Model subclass that just implements a hard-coded function of 2 input features or something like that.

fartashf:
Thanks for suggesting the test. I made sure the new test would fail if the clipping is not done right. It is explained in a comment inside the test.

PR #224, opened 2017-08-14
Wrapper around Madry lab challenges

goodfeli:


goodfeli:
@carlini would you be willing to review this one?

cleverhans/model.py
```diff
@@ -28,9 +29,15 @@ def get_layer(self, x, layer):
...
         :return: A symbolic representation of the hidden features
+        :raise: NoSuchLayerError if `layer` is not in the model.
         """
         # Return the symbolic representation for this layer.
-        return self.fprop(x)[layer]
+        output = self.fprop(x)
+        try:
+            requested = output[layer]
+        except KeyError:
+            raise NoSuchLayerError()
```
goodfeli: My goal here is to avoid querying the dictionary twice, since that's slower.

cleverhans/model.py
```diff
@@ -46,9 +53,9 @@ def get_probs(self, x):
         :return: A symbolic representation of the output probabilities (i.e.,
                 the output values produced by the softmax layer).
         """
-        if 'probs' in self.get_layer_names():
```
goodfeli: In this line, there are two motivations:
- Main motivation: don't require the `Model` subclass to implement `get_layer_names`.
- Minor motivation: use exceptions instead of tests, to remove some small runtime costs. If we use tests, then `get_layer` will query the dictionary to do the test, then query the dictionary again  to get the layer, and then we'll query the dictionary again to do the test here. So 3 dict lookups when we could have just one.

cleverhans/model.py
carlini: try/catch is usually slower than the extra access, but if you are worried about that then you can do `requested = output.get(layer)` and it will be None if it doesn't exist.

cleverhans/model.py
carlini: I guess the fast-path is simpler with the try/catch in this case. The case where probs isn't defined is probably more rare, and this is okay.

examples/madry_lab_challenges/madry_mnist_model.py
```diff
@@ -0,0 +1,79 @@
...
+        h_fc1 = tf.nn.relu(tf.matmul(h_pool2_flat, self.W_fc1) + self.b_fc1)
+
+        # output layer
+        logits = tf.matmul(h_fc1, self.W_fc2) + self.b_fc2
+
+        l = locals()
+        for key in sorted(l.keys()):
+            if key in ["self", "output"]:
+                continue
+            output[key] = l[key]
```
goodfeli: I avoid using ordinary Python dicts when possible, because they give different iteration orders on different invocations of the script. When the dictionary contains Theano/TensorFlow tensors, this can actually change the binary code that executes later. That can cause weird issues that are really hard to debug, because you can't even execute the same binary twice reliably. To avoid those bugs, I'm using an OrderedDict here.

In this case, there's some trouble because `locals()` is itself an ordinary python dict. I have to sort the keys of locals() before using them to construct an OrderedDict.

carlini: How about

`output = OrderdDict(sorted(locals()))`
`del output['self']`

then?

examples/madry_lab_challenges/attack_model.py
```diff
@@ -0,0 +1,96 @@
...
+    train_start = 0
+    train_end = 60000
+    test_start = 0
+    test_end = 10000
+    X_train, Y_train, X_test, Y_test = data_mnist(train_start=train_start,
+                                                  train_end=train_end,
+                                                  test_start=test_start,
+                                                  test_end=test_end)
+
+    assert Y_train.shape[1] == 10.
```
goodfeli: That's apparently an oddity that's been copy-pasted around the tutorials all along. I fixed it both here and in the tutorials

examples/madry_lab_challenges/attack_model.py
```diff
@@ -0,0 +1,96 @@
...
+    if "MNIST_CHALLENGE_DIR" in os.environ:
+        dirs.insert(0, os.environ['MNIST_CHALLENGE_DIR'])
+    else:
+        default_checkpoint_dir = os.path.join(*dirs)
+
+    flags.DEFINE_integer('batch_size', 128, "batch size")
+    flags.DEFINE_float(
+        'label_smooth', 0.1, "Amount to subtract from correct label and distribute among other labels")
+    flags.DEFINE_string(
+        'attack_type', 'fgsm', "Attack type: 'fgsm'->fast gradient sign method, 'bim'->'basic iterative method'")
```
goodfeli: Apparently we weren't running PEP8 on the examples directory. I turned it on and fixed the PEP8 errors here and in the CIFAR10 example. The NIPS competition directory is very far from PEP8 compliant because it is following the Google style guide instead. Since it would be a lot of work to change the style guide, I just white-listed that directory as allowed not to be PEP8 compliant for now.

examples/madry_lab_challenges/madry_mnist_model.py
goodfeli: `sorted(locals())` actually gives you a list of the sorted keys, so that wouldn't work quite as written. You still need to iterate over the output of `sorted` to package up the values along with the keys. I factored it into a separate function though.

goodfeli:
Tests pass, ready for more review

Issue #217, opened 2017-08-05
deduce number of classes from model's output tensor


npapernot:
Some attacks require nb_classes to be passed as an argument, but this could be computed with the shape of the predictions tensor

carlini:
I have a branch for this and will push it some time soon.

goodfeli:
This was completed by PR #226

PR #226, opened 2017-08-14
Remove nb_classes

carlini:
Remove nb_classes argument from CW and JSMA attacks. Infer it from the shape of the result of the model.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -254,12 +254,9 @@ class FastGradientMethod(Attack):
         else:
             from .attacks_th import fgm
 
-        if self.y is not None:
-            y = self.y
-        else:
-            y = self.y_target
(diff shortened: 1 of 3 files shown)


cleverhans/attacks.py
```diff
@@ -172,6 +172,27 @@ def generate_np(self, x_val, **kwargs):
 
         return self.sess.run(x_adv, feed_dict)
 
+    def get_labels(self, x, kwargs):
```
goodfeli: docstring please

goodfeli: maybe rename to `get_or_guess_labels` since it can use the predictions instead of the labels?

goodfeli:
Looks good apart from the two comments I made

carlini:
(For github: This fixes #217.)

PR #227, opened 2017-08-14
Remove keras from tutorials

carlini:
Remove use of keras from the JSMA/CW tutorials. Replace the model with the pure TF model in mnist_tutorial_tf. Move the model in that tutorial to models.py.


goodfeli:
Since this moves the MLP and all the Layers into the core library, it's a pretty big shift in philosophy for the library, and should probably be discussed on the cleverhans-dev list first. (So far, we haven't wanted to make CleverHans provide the same kind of functionality as Keras or Sonnet do; we just needed these classes in the tutorial to show that whatever kind of code you can wrap in the Model interface will work with the attacks)

A less controversial move would be to make the tutorials directory into a module and add some tutorial_models.py submodule that the tutorials share. I think that would be fine to merge just with one reviewer (and I personally would be happy to serve as that reviewer and would approve it).

carlini:
I don't really want this to be a replacement for Keras. I moved the MLP to the core models file because utils_keras has its own cnn_model, and this was mainly meant as a drop-in replacement for that.

That said, I can see how people might now build off of (or be be confused by) a MLP implementation in the core of cleverhans. I'd be perfectly happy moving it to the tutorials folder.

goodfeli:
Yes, I think people would build off of the MLP class because it is a generic building block. cnn_model was kind of a gray area because it is model functionality, but it's also such an end product that it's hard to imagine many people wanting to import it.

In the long run, we should evaluate Sonnet as an alternative to Keras that probably supports some of the use cases we weren't able to fulfill with Keras.

goodfeli:
So I guess move MLP to it's own module in tutorials? And then I'll re-review

carlini:
Makes sense. I wonder if there's a need for using Sonnet, of if it would be possible to just nix all model-building dependencies and write it with pure tf. But yeah, that's a discussion for later.

Yeah, doing that now.

goodfeli:
By pure tf, do you mean using our own MLP / Layer classes? Or do you mean just directly implementing the Model class methods without having a large ecosystem of different classes that interoperate?

carlini:
No, not our own MLP/Layer -- just removing all specifics of libraries that help build models. CleverHans has its own Model interface now, so I don't see any need to use Keras to build a model (or Sonnet) internally.

Basically, I would argue the extent of Keras in CleverHans is a KerasModelWrapper to allow the Keras <-> CleverHans interaction.

goodfeli:
OK, I agree that CleverHans itself doesn't need to provide much in terms of a model framework. I meant we should evaluate Sonnet for implementing research project ideas using CleverHans as a dependency.

carlini:
Makes sense.

tutorials/mnist_blackbox.py
```diff
@@ -6,45 +6,28 @@
...
 from tensorflow.python.platform import app
 from tensorflow.python.platform import flags
 
-from cleverhans.utils_keras import cnn_model
 from cleverhans.utils_mnist import data_mnist
 from cleverhans.utils_tf import model_train, model_eval, batch_eval
 from cleverhans.attacks import FastGradientMethod
 from cleverhans.attacks_tf import jacobian_graph, jacobian_augmentation
-from cleverhans.utils_keras import KerasModelWrapper
+from cleverhans.model import make_basic_cnn
```
goodfeli: still needs to be changed to point to the tutorials submodule

npapernot:
I agree that it would make sense to have a discussion on this topic in the cleverhans-dev ml. It seems OK for cleverhans to provide wrappers that make it easier for people to use high-level frameworks like keras or slim or sonnet, but I agree that providing actual models seems outside the initial scope of the project

carlini:
Whoops, I forgot I never finished rewriting the black box tutorial. Will do that tomorrow.

npapernot:
Thanks @carlini

carlini:
The blackbox test failed, rate was 0.69 instead of 0.7. I've adjusted the test.

I would vote to remove all thresholds from these tests and just use them as "does it crash" test, but that's another conversation for later.

carlini:
Looks like I may have decreased the accuracy with the switch away from keras, now the other accuracy piece is failing. I'll look in to how that happened.

goodfeli:
If the switch away from keras indeed decreased the accuracy, it sounds like there is some value to the accuracy threshold in the tests :)

carlini:
Yeah ... I would just like a reproducible way to know that though. Because more often than not, the tutorials failing have just been because it's a few percent below the threshold randomly.

carlini:
The issue is that dropout is enabled during evaluation-time without keras handling this. I don't especially want to have to re-implement training/testing dropout enable/disable for our own MLP layer, but I will if I have to. Ian, any thoughts on how to do this easily?

Otherwise I'll just add a cleverhans_learning_phase to mirror keras_learning_phase.

goodfeli:
I think most our tutorials just don't have dropout. It's probably OK to not use dropout in this one too. The purpose of the tests is to make sure we notice when something changes. In this case, if we're pretty confident that dropout was the only change, it should be alright to decrease the threshold a little bit.

It could also make sense to put a **kwargs argument in Model.fprop, Model.get_layer, etc., and then we could pass in an argument like `dropout_input_prob` for the Dropout layer to pick up.

carlini:
Okay, I've removed dropout now. I don't want to decide some fairly deep architectural changes to how models exist in this PR.

carlini:
Alright, it's passing now.

cleverhans/model.py
```diff
@@ -1,4 +1,5 @@
 from abc import ABCMeta
+import numpy as np
```
npapernot: Is this import on purpose?

tutorials/mnist_tutorial_cw.py
```diff
@@ -13,10 +12,12 @@
 import logging
 import os
 from cleverhans.attacks import CarliniWagnerL2
-from cleverhans.utils import grid_visual, AccuracyReport, set_log_level
-from cleverhans.utils_keras import cnn_model, KerasModelWrapper
+from cleverhans.utils import pair_visual, grid_visual, AccuracyReport
+from cleverhans.utils import set_log_level
 from cleverhans.utils_mnist import data_mnist
 from cleverhans.utils_tf import model_train, model_eval, tf_model_load
+from cleverhans.utils_tf import model_argmax
```
npapernot: Is this import needed?

cleverhans/model.py
carlini: Nope. Was from when the model was there and not in tutorials_model.

tutorials/mnist_tutorial_cw.py
carlini: Nope

carlini:
I didn't notice a difference, but I can get the stats from before and after to verify.

npapernot:
Thanks, it would be great to double-check yeah. Besides that, LGTM

carlini:
They're identical:


master

CW: model accuracy 0.9867, distortion 3.172
JSMA: model accuracy 0.9873, distortion 0.0290
blackbox: 0.9882, secondary accuracy 0.837

remove_keras

CW: model accuracy 0.9879, distortion 3.199
JSMA: model accuracy 0.0.9887, distortion 0.0288
blackbox: 0.9882, secondary accuracy 0.822

npapernot:
Thanks for sending this out!

PR #228, opened 2017-08-15
explicit which layer should model callables provide to attacks

### LINKED ISSUE #160: Be more clear when attacks need logits or softmax probs
Some attacks operate with the model's logits and others with the softmax probs. We should improve the documentation to make sure it is clear when the `model` argument of an attack class needs to be logits and when it needs to be softmax probs (see #157). In addition, we could use #154 to have some more automated checks in the `__init__` method of attacks that test whether the `model` argument is appropriate. There could also be some more generic checks looking for the TensorFlow softmax op.

npapernot:
Closes #160

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -21,7 +21,7 @@ class Attack(object):
 
     def __init__(self, model, back='tf', sess=None):
         """
-        :param model: An instance of the Model class.
+        :param model: An instance of the cleverhans.model.Model class.
         :param back: The backend to use. Either 'tf' (default) or 'th'
                      (support for Theano is however deprecated and will
                      be removed on 2017-11-08).
@@ -34,11 +34,12 @@ class Attack(object):
         if not isinstance(model, Model):
             if hasattr(model, '__call__'):
                 warnings.warn("CleverHans support for supplying a callable"
-                              " instead of an instance of the Model class is"
+                              " instead of an instance of the"
+                              " cleverhans.model.Model class is"
                               " deprecated and will be dropped on 2018-01-11.")
             else:
                 raise ValueError("The model argument should be an instance of"
-                                 " the Model class.")
+                                 " the cleverhans.model.Model class.")
         if back == 'th':
             warnings.warn("CleverHans support for Theano is deprecated and "
                           "will be dropped on 2017-11-08.")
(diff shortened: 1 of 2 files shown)


cleverhans/attacks.py
```diff
@@ -226,6 +236,9 @@ class FastGradientMethod(Attack):
     def __init__(self, model, back='tf', sess=None):
         """
         Create a FastGradientMethod instance.
```
goodfeli: It's more straightforward to just say `cleverhans.model.Model`.

Rather than documenting how the callable should behave, maybe we should just leave that out. If people want to duck-type the `Model` class, its call method returns probabilities, so they should do the right thing when reading its doc.

In places were we require probabilities, we can usually check that the output comes from some specific op, usually Softmax. Explicit runtime checks are a lot more reliable than documentation.

cleverhans/attacks_tf.py
```diff
@@ -459,17 +462,16 @@ def __init__(self, sess, model, batch_size, confidence,
...
-                  targeted attack.
-        :param nb_classes: The number of classes the model has.
+        :param sess: a TF session.
+        :param model: a CleverHans model object.
+        :param batch_size: Number of attacks to run simultaneously.
         :param confidence: Confidence of adversarial examples: higher produces
                            examples with larger l2 distortion, but more
                            strongly classified as adversarial.
-        :param batch_size: Number of attacks to run simultaneously.
+        :param targeted: boolean controling the behavior of the adversarial
```
goodfeli: controlling

cleverhans/attacks_tf.py
```diff
@@ -459,17 +462,16 @@ def __init__(self, sess, model, batch_size, confidence,
...
 
-        :param x: (required) A tensor with the inputs.
-        :param y: (optional) A tensor with the true labels for an untargeted
-                  attack. If None (and y_target is None) then use the
-                  original labels the classifier assigns.
-        :param y_target: (optional) A tensor with the target labels for a
-                  targeted attack.
-        :param nb_classes: The number of classes the model has.
+        :param sess: a TF session.
+        :param model: a CleverHans model object.
```
goodfeli: Maybe "instance of cleverhans.model.Model" to be maximally straightforward

npapernot:
All comments have been addressed

goodfeli:
I think the build just stalled, so pushing a dummy commit should fix it. LGTM, OK to merge once the tests pass.

PR #230, opened 2017-08-15
Fix logger messages

carlini:
A few minor fixes for the logger, changing messages in a few places to be slightly better. Also enable the logger for the tf example.

Issue #203, opened 2017-07-13
Update tutorial markdown to show use of new abstraction


abhibhav14:
The tutorials can be updated to include the new model abstraction, to make it easier for people to learn about that feature. Also, the tutorial markdown use the old attack models where the attacks are called as functions directly, instead of initializing attacks as class instances, as is done in the actual tutorial code.

PR #231, opened 2017-08-16
Remove deprecated functions

npapernot:
In preparation for v.2.0 release, cleaned up deprecated functions.

OK to wait 2 days to merge if we want to strictly enforce the deprecation days (one of them is Aug 17)


goodfeli:
Well it's August 17 in Samoa so I think it's OK to merge. (Usually I write "on or after" rather than just "after" in the deprecation messages, I see these don't do that, but I doubt anyone will object)
I agree this is a good thing to take out before the release, but we also want to make sure the code really isn't used.
Maybe do code freeze until Aug 23 following merging this?

npapernot:
I checked that the code isn't used within our repo, so we are good on that end.

Yes, we can start a code freeze (we can probably merge #230 before the code freeze as well)

During the code freeze, we can take care of #203

PR #236, opened 2017-08-22
rename "tutorials" to "cleverhans_tutorials"

goodfeli:


goodfeli:
Note: it's possible that 3rd party code was using "import tutorials" and this will break it. I think it's OK not to do a 6 month deprecation warning because:
1) the tutorials aren't part of the main library and they're just meant to be example code, not supported functionality with the same contracts
2) it's only been a few days that it was possible to do "import tutorials" and most users are probably not even aware that it's possible yet

npapernot:
I think it's fine as well, deprecation seems like unnecessary overhead here and kind of defeats the purpose of changing the name anyways

npapernot:
LGTM

PR #249, opened 2017-08-25
Added the bias to Conv2D in tutorial_mnist_tf

fartashf:
The bias was initialized but not used.

PR #252, opened 2017-08-30
tighten the tests for the tutorial

goodfeli:
Some improvements to the codebase mean we should expect better results now.
See the comments in this recently merged PR that improved performance: https://github.com/tensorflow/cleverhans/pull/249


goodfeli:
All the tensorflow tests failed. It's a little bit strange, because this passes on my desktop, even when I prevent the use of GPU. It's hard to tell how bad the failure was because the tests use `assertTrue` rather than `assertGreater` so they didn't print the failing values. I'm pushing a commit to use `assertGreater` so we can get more info on how it's failing on Travis.

npapernot:
LGTM given comments made in #249

PR #255, opened 2017-08-31
Test that our seeding works. It doesn't, so the tests are loose :(

goodfeli:


npapernot:
Thanks for looking into this. Is this work in progress? If no, is there a reason for only switching `test_mnist_tutorial_tf.py` to inherit from the new CleverHans test class but not the other tutorials?

goodfeli:
This is ready to review and merge if it passes review. (Someone will need to do another PR later that fixes the non-determinism, and then they'll be able to tighten the tolerances in this test)

I changed the inheritance for just one test because it's the only test where I needed to use the `assertClose` method provided by `CleverHansTest`. I guess we could also provide non-determinism tests for the other tutorials, but I wasn't sure if having more than one non-determinism test would be valuable enough to make up for the increase in test runtime.

npapernot:
Ok thanks for the clarification. In that case, LGTM.

PR #256, opened 2017-08-31
Fix problem with relative path in adversarial competition submission validation tool.

AlexeyKurakin:


npapernot:
I restarted the test that failed because of a network error

PR #258, opened 2017-09-05
added a robust vision benchmark example submission

jonasrauber:
This pull request includes wrappers that make it easy to submit a CleverHans attack to the [`Robust Vision Benchmark`](https://robust.vision/benchmark) and an example submission.

### DIFF
--- a/examples/robust_vision_benchmark/README.md
+++ b/examples/robust_vision_benchmark/README.md
@@ -1,0 +1,39 @@
+## CleverHans Example Submission to the Robust Vision Benchmark
+
+Using the wrappers in this repository, submitting a CleverHans attack to the [Robust Vision Benchmark](https://robust.vision/benchmark) requires just a few lines of code:
+
+```python
+#!/usr/bin/env python3
+
+import numpy as np
+from cleverhans.attacks import FastGradientMethod
+from robust_vision_benchmark import attack_server
+from utils import cleverhans_attack_wrapper
+
+
+def attack(model, session, a):
+    fgsm = FastGradientMethod(model, sess=session)
+    image = a.original_image[np.newaxis]
+    return fgsm.generate_np(image)
+
+
+attack_server(cleverhans_attack_wrapper(attack))
+```
+
+The full example can be found in the `cleverhans_attack_example` folder.
+
+### Testing an attack
+
+Just install the latest version of the [robust-vision-benchmark python package](https://github.com/bethgelab/robust-vision-benchmark) using
+
+```bash
+pip install --upgrade robust-vision-benchmark
+```
+
+and run
+
+```bash
+rvb-test-attack cleverhans_attack_example/
+```
+
+to test the attack. Once the test succeeds, you can **[submit your attack](https://github.com/bethgelab/robust-vision-benchmark)**.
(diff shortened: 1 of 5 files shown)


examples/robust_vision_benchmark/cleverhans_attack_example/requirements.txt
```diff
@@ -0,0 +1,2 @@
+numpy
```
npapernot: Why are you including a requirements.txt file here?

wielandbrendel: The requirements.txt is only used in the Dockerfile and is part of the submission. We could make it part of the Dockerfile if you prefer to avoid confusion / reduce the number of files.

npapernot: It should be fine to leave like that, thanks for clarifying.

examples/robust_vision_benchmark/README.md
```diff
@@ -0,0 +1,39 @@
+## CleverHans Example Submission to the Robust Vision Benchmark
+
+Using the wrappers in this repository, submitting a CleverHans attack to the [Robust Vision Benchmark](https://robust.vision/benchmark) requires just a few lines of code:
+
+```python
```
npapernot: Could you remove the code snippet from the readme and just point to the file in the `cleverhans_attack_example` folder? We've had issues in the past with example code copied in markdown files getting out-dated.

examples/robust_vision_benchmark/cleverhans_attack_example/utils.py
```diff
@@ -0,0 +1,89 @@
...
+
+    tf.RegisterGradient(rnd_name)(grad)
+    g = tf.get_default_graph()
+    with g.gradient_override_map({
+            "PyFunc": rnd_name,
+            "PyFuncStateless": rnd_name}):
+        return tf.py_func(func, inp, Tout, stateful=stateful, name=name)
+
+
+class RVBCleverhansModel(cleverhans.model.Model):
```
npapernot: I see one issue with this wrapper is that some attacks will try to use the probabilities (post-softmax) instead of the logits. In its current form, the method `get_logits()` will work but not `get_probs()`. Could you also provide access to the probabilities?

wielandbrendel: We didn't implement get_probs because the Cleverhans model anyway falls back to softmax(logits) if probabilities are not defined, right?

npapernot: Not necessarily. If you take a look at the FastGradientMethod class, since you are passing it a `cleverhans.model.Model` object, it will call `model.get_probs()`: https://github.com/tensorflow/cleverhans/blob/master/cleverhans/attacks.py#L283

wielandbrendel: But that should be fine since RBVCleverhansModel is derived from cleverhans.model.Model, and the latter implements get_prob such that it computes softmax(logits) if no layer with the name 'probs' exists: https://github.com/tensorflow/cleverhans/blob/master/cleverhans/model.py#L58-L60 Or am I missing something?

npapernot: That's correct, my bad I forgot about that default implementation we added

examples/robust_vision_benchmark/README.md
npapernot: Can you address this? We can merge afterwards

jonasrauber: Done

PR #260, opened 2017-09-13
implementation of deepfool

cihangxie:
implementation of deepfool algorithm https://arxiv.org/pdf/1511.04599.pdf

PR #261, opened 2017-09-13
implementation of deepfool after "rebase and amend the commits"

cihangxie:


bairdzhang:
I signed it!

bairdzhang:
After rebasing and amending the commits, the email address now has been fixed. But there is still an error message says 'Failure: CLAs are signed, but unable to verify author consent'. Is there anything I am missing? Thanks!

npapernot:
The CLA bot cleared this new PR so we are all good on this side. Let me take a look at the code again.

cleverhans/attacks.py
```diff
@@ -749,6 +749,91 @@ def parse_params(self, y=None, y_target=None, nb_classes=None,
         self.clip_max = clip_max
 
 
+class DeepFool(Attack):
+
+    """
+    DeepFool is an untargeted & iterative attack which aims at finding the
+    minimum adversarial perturbations in deep networks. The implementation
```
npapernot: This makes it hard to understand what the approach is doing. How about something like: "DeepFool is an untargeted & iterative attack which is based on an iterative linearization of the classifier. "

cleverhans/attacks.py
```diff
@@ -749,6 +749,91 @@ def parse_params(self, y=None, y_target=None, nb_classes=None,
...
+        if not isinstance(self.model, Model):
+            self.model = CallableModelWrapper(self.model, 'logits')
+
+    def generate(self, x, **kwargs):
+        """
+        Generate symbolic graph for adversarial examples and return.
+        :param x: The model's symbolic inputs.
+        :param nb_candidate: The number of classes to test against, i.e.,
+                            deepfool only consider nb_candidate classes when
+                            attacking (thus accelerate speed)
```
npapernot: How are these classes chosen? Are those the first nb_candidate classes that are different from the correct class?

cleverhans/attacks.py
```diff
@@ -749,6 +749,91 @@ def parse_params(self, y=None, y_target=None, nb_classes=None,
...
+
+        # Parse and save attack-specific parameters
+        assert self.parse_params(**kwargs)
+
+        assert self.nb_candidate <= self.nb_classes,\
+            'nb_candidate should not be greater than nb_classes'
+
+        # Define graph wrt to this input placeholder
+        logits = self.model.get_logits(x)
+        preds = tf.reshape(tf.nn.top_k(logits, k=self.nb_candidate)[0],
```
npapernot: If the top k scores are used for preds, why is `gradient_graph` selecting the first k classes using xrange?

cleverhans/attacks.py
```diff
@@ -749,6 +749,91 @@ def parse_params(self, y=None, y_target=None, nb_classes=None,
...
+        # Define graph
+        def deepfool_wrap(x_val):
+            return deepfool_batch(self.sess, x, preds, logits, grads, x_val,
+                                  self.nb_candidate, self.overshoot,
+                                  self.max_iter, self.clip_min, self.clip_max,
+                                  self.nb_classes)
+        return tf.py_func(deepfool_wrap, [x], tf.float32)
+
+    def parse_params(self, nb_candidate=10, overshoot=0.02, max_iter=50,
+                     nb_classes=1001, clip_min=0., clip_max=1., **kwargs):
```
npapernot: Can you infer the number of classes from the shape of `preds` tensor instead of feeding it as an argument to this function?

cleverhans/attacks_tf.py
```diff
@@ -726,3 +726,135 @@ def compare(x, y):
...
+                        deepfool only consider nb_candidate classes when
+                        attacking (thus accelerate speed)
+    :param overshoot: A termination criterion to prevent vanishing updates
+    :param max_iter: Maximum number of iteration for DeepFool
+    :param clip_min: Minimum value for components of the example returned
+    :param clip_max: Maximum value for components of the example returned
+    :param nb_classes: Number of model output classes
+    :return: Adversarial examples
+    """
+    X_adv = np.zeros(X.shape)
```
npapernot: Why is this line needed?

cleverhans_tutorials/mnist_tutorial_deepfool.py
```diff
@@ -0,0 +1,184 @@
+# adapted from mnist_tutorial_cw.py
```
npapernot: Can you remove this tutorial? (see the discussion we had on the previous PR #260)

tests_tf/test_attacks.py
```diff
@@ -479,5 +478,64 @@ def test_generate_np_targeted_gives_adversarial_example(self):
         self.assertTrue(worked > .9)
 
 
+class TestDeepFool(CleverHansTest):
```
npapernot: Thank you for including those tests

tests_tf/test_mnist_tutorial_deepfool.py
```diff
@@ -0,0 +1,28 @@
+import unittest
```
npapernot: Can you remove this test and the tutorial that goes with it?

cleverhans/attacks_tf.py
```diff
@@ -726,3 +726,135 @@ def compare(x, y):
...
+    _logger.info("{} out of {}".format(sum(current != original),
+                                       sample.shape[0]) +
+                 " becomes adversarial examples at iteration {}".format(
+                     iteration))
+    # need to clip this image into the given range
+    adv_x = np.clip((1+overshoot)*r_tot + sample, clip_min, clip_max)
+    return adv_x
+
+
+def gradient_graph(predictions, x, nb_candidate):
```
npapernot: Can you test this function separately from the attack class for DeepFool?

cleverhans/attacks_tf.py
```diff
@@ -726,3 +726,135 @@ def compare(x, y):
...
+    :param clip_max: Maximum value for components of the example returned
+    :return: an adversarial sample
+    """
+    import copy
+
+    adv_x = copy.copy(sample)
+    # Initialize the loop variables
+    iteration = 0
+    current = utils_tf.model_argmax(sess, x, logits, adv_x, feed=feed)
+    if current.shape == ():
```
npapernot: I read your response on the other PR, thanks. So my understanding is this requires that x is an image (i.e., has 4 dimensions). We'd like to have the attacks work on inputs of any shape as much as possible. Do you think the code could be rewritten to not depend on the shape of an image and take any input with a shape whose first dimension is the batch size?

cleverhans/attacks.py
cihangxie: These classes are selected based on the prediction confidence, e.g., the classes with confidence among the top nb_candidate are chosen. The first nb_candidate classes is the same as the correct class if the prediction is correct. The parameter nb_candidate is named as parameter NUM_LABELS in the original DeepFool implementation (DeepFool/MATLAB/adversarial_perturbation.m)

cleverhans/attacks.py
cihangxie: Thanks for your description. We have changed to that

cleverhans/attacks.py
cihangxie: The preds will return the logit value of these nb_candidate classes, and the grads will return the gradients of these nb_candidate classes.

cleverhans/attacks.py
cihangxie: Yes. We have changed that. Thanks for your suggestion

cleverhans/attacks_tf.py
cihangxie: For this function, it does not require that x must have 4 dimensions (you can see it can work on random input with shape [100, 2] at test_attacks.py). The reason to judge the shape of current is due to the output of function utils_tf.model_argmax. When batch_size = 1, utils_tf.model_argmax will return a value with shape = (), while batch_size > 1, utils_tf.model_argmax will return a value with shape = (batch_size, ). In order to let the output under these two situations are in the same shape, we let current = np.array([current]) when current.shape == () (or we can replace this condition by batch_size == 1)

cleverhans/attacks_tf.py
cihangxie: The function gradient_graph is actually the same as the function jacobian_graph, with only difference at the output shape. We have deleted this function, and use jacobian_graph + reshape instead.

cleverhans_tutorials/mnist_tutorial_deepfool.py
cihangxie: we have deleted those tutorials.

goodfeli:
Thanks!

PR #263, opened 2017-09-15
Check sess is provided when using TF in Attacks

### LINKED ISSUE #247: How to get the attack examples with SaliencyMapMethod on the ImageNet datasets?
Here, I want to get the attack examples with SaliencyMapMethod on the ImageNet datasets based on the examples/nips17_adversarial_competition/sample_attacks/fgsm/attack_fgsm.py file.

And I just change the following code :

```
   fgsm = FastGradientMethod(model)
   x_adv = fgsm.generate(x_input, eps=eps, clip_min=-1., clip_max=1.)
```
with:

```
    jsma= SaliencyMapMethod(model)
    target = np.zeros((1,1000),dtype=np.float32)
    target[0,50] = 1                    #here, we suppose that the target label is 50
    jsma_params = {'theta': 1., 'gamma': 0.1,
                   'clip_min': 0., 'clip_max': 1.,
                   'y_target': target}
    x_adv = jsma.generate(x_input,**jsma_params)
```
But, it doesn't work.

npapernot:
Closes #247

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -29,12 +29,13 @@ class Attack(object):
         """
         if not(back == 'tf' or back == 'th'):
             raise ValueError("Backend argument must either be 'tf' or 'th'.")
-        
+
         if back == 'th' and sess is not None:
             raise Exception("A session should not be provided when using th.")
         elif back == 'tf' and sess is None:
-            raise ValueError("A session should be provided when using tf.")
-        
+            warnings.warn("A session should be provided when using tf.")
+            sess = tf.get_default_session()
+
         if not isinstance(model, Model):
             if hasattr(model, '__call__'):
                 warnings.warn("CleverHans support for supplying a callable"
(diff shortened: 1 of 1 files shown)


goodfeli:
Shouldn't it just use the tf default session when no explicit session is provided?

rwightman:
Won't this break some existing usages of attack classes that don't require session? Ie FastGradientMethod, in nips17 examples it's currently called without session for TF.

goodfeli:
I think @rwightman is right and I assume that's why the tests are failing but I haven't looked at their output.

npapernot:
I changed it to use the default session instead when no session are given.

Do we want to include a warning when falling back to the default session? I have a warning now but don't have a strong opinion against removing it.

goodfeli:
I don't think there should be a warning. TensorFlow doesn't issue warnings when you use the default session. There's nothing wrong with using a default session.

Issue #17, opened 2016-09-19
TensorFlow issues weird error message


goodfeli:
This seems harmless, but it is annoying. Not sure if we can fix it, since it may be a bug in TensorFlow itself.

At the end of the MNIST tutorial, I get this error message:
    Exception AttributeError: AttributeError("'NoneType' object has no attribute 'raise_exception_on_not_ok_status'",) in <bound method Session.__del__ of <tensorflow.python.client.session.Session object at 0x7fb09c05ad50>> ignored

npapernot:
@goodfeli I haven't seen this message in a while, do you agree to close this issue?

goodfeli:
I still get it consistently:
`Completed model training.
Exception ignored in: <bound method BaseSession.__del__ of <tensorflow.python.client.session.Session object at 0x7f99aa445c88>>
Traceback (most recent call last):
  File "/opt/anaconda/4.2.0/lib/python3.5/site-packages/tensorflow/python/client/session.py", line 522, in __del__
  File "/opt/anaconda/4.2.0/lib/python3.5/site-packages/tensorflow/python/client/session.py", line 518, in close
  File "/opt/anaconda/4.2.0/lib/python3.5/contextlib.py", line 133, in helper
TypeError: 'NoneType' object is not callable`

goodfeli:
Looks fixed now, maybe by updating tensorflow to 0.12

goodfeli:
It still happens, it is just flaky.

Out of three runs, I had this happen once:
    Completed model training.
    Exception ignored in: <bound method BaseSession.__del__ of <tensorflow.python.client.session.Session object at 0x7f5564ed4f28>>
    Traceback (most recent call last):
      File "/home/ian/.local/lib/python3.5/site-packages/tensorflow/python/client/session.py", line 581, in __del__
    UnboundLocalError: local variable 'status' referenced before assignment

dguera:
I get the following message which is somehow related:

Exception ignored in: <bound method Session.__del__ of <tensorflow.python.client.session.Session object at 0x7f4e28be0668>>
Traceback (most recent call last):
  File "/data/dgueraco/tensorflow/lib/python3.4/site-packages/tensorflow/python/client/session.py", line 582, in __del__
AttributeError: 'NoneType' object has no attribute 'TF_DeleteStatus'

Adding a simple `del sess` at the end of the tutorial fixes it.

goodfeli:
[...]

PR #266, opened 2017-09-19
Make CarliniWagner tutorial more general, use source_samples flag

AngusG:
The `source_samples` flag wasn't being used in the CarliniWagner tutorial, this PR attempts to make the tutorial more reasonable for various settings of the `viz_enabled`,`targeted`, and `source_samples` `FLAGS`. Magic numbers have also been replaced with the relevant variables where appropriate.

### DIFF
--- a/cleverhans_tutorials/mnist_tutorial_cw.py
+++ b/cleverhans_tutorials/mnist_tutorial_cw.py
@@ -114,13 +114,11 @@ def mnist_tutorial_cw(train_start=0, train_end=60000, test_start=0,
     cw = CarliniWagnerL2(model, back='tf', sess=sess)
 
     if viz_enabled:
+        assert source_samples == nb_classes
         idxs = [np.where(np.argmax(Y_test, axis=1) == i)[0][0]
                 for i in range(nb_classes)]
-
     if targeted:
         if viz_enabled:
-            assert source_samples == nb_classes
-
             # Initialize our array for grid visualization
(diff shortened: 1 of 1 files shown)


AngusG:
The failed test for Python 2.7/TensorFlow 1.1.0 appears to be related to #17.

npapernot:
Thanks for the PR, it looks mostly good. I would suggest one change though: right now when the attack is untargeted, the code is using the grid visualization. It would most likely make more sense to just display them with the `pair_visual` function from https://github.com/tensorflow/cleverhans/blob/master/cleverhans/utils.py#L150. Otherwise, I fear the grid may mislead users into thinking that the attack was targeted.

AngusG:
Thanks for the feedback. I actually think that the grid visualisation is reasonable for the untargeted case as shown in the following screenshot (one adversarial per corresponding clean image). I was thinking we could do something similar to the `mnist_tutorial_jsma.py` where pair_visual is used as an animation, but the adversarials are computed in a batch for the cw tutorial as opposed to one at a time so I wasn't sure it made sense here. 

cleverhans _grid_visualization

npapernot:
I see your point. That makes sense. Could you share a screenshot of the whole matplotlib window to double check how it renders?

AngusG:
Here's a screenshot:

cleverhans _grid_visualization_screenshot_untargeted

npapernot:
Perfect, thanks for sending this out.

PR #268, opened 2017-09-20
Turn on clipping for the tutorials

goodfeli:


npapernot:
It appears the tests need to be loosened up even more.

PR #272, opened 2017-09-21
Add code for linear extrapolation plot

royaurko:
I added two functions in cleverhans/utils.py:
 1. get_logits_over_interval() to get the logits over a bounded interval
 2. linear_extrapolation_plot() to plot these logits

Also added the file linear_extrapolation_plot_example.py in the example directory.


royaurko:
Fixed the commit that was causing the CLA complaint.

PR #273, opened 2017-09-21
Add linear extrapolation plots

royaurko:
I added two functions in cleverhans/utils.py:
 1. get_logits_over_interval() to get the logits over a bounded interval
 2. linear_extrapolation_plot() to plot these logits

Also added the file linear_extrapolation_plot_example.py in the example directory.

### DIFF
--- a/cleverhans/utils.py
+++ b/cleverhans/utils.py
@@ -258,7 +258,7 @@ def get_logits_over_interval(sess, model, x, adv_x, x_data,
     channels = shape[3]
     size = height * width * channels
 
-    epsilon_array_t = tf.constant(epsilon_array)
+    epsilon_array_t = tf.constant(epsilon_array, dtype=tf.float32)
     epsilon_array_t = tf.stack([epsilon_array_t] * size, axis=1)
     epsilon_array_t = tf.reshape(epsilon_array_t,
                                  shape=[num_points, height,
--- a/cleverhans/utils.py
+++ b/cleverhans/utils.py
@@ -227,20 +227,19 @@ def get_logits_over_interval(sess, model, x_data, fgsm_params,
 
     Args:
         sess: Tf session
-        model: Model for which we wish to get logits
-        x: Data tensor of shape [1, height, width, channels]
-        adv_x: Tensor for adversarial perturbation of x
-        x_data: Numpy array corresponding to single data
+        model: Model for which we wish to get logits.
+        x_data: Numpy array corresponding to single data.
                 point of shape [height, width, channels].
-        min_epsilon: Minimum value of epsilon over the interval
-        max_epsilon: Maximum value of epsilon over the interval
-        num_points: Number of points used to interpolate
+        fgsm_params: Parameters for generating adversarial examples.
+        min_epsilon: Minimum value of epsilon over the interval.
+        max_epsilon: Maximum value of epsilon over the interval.
+        num_points: Number of points used to interpolate.
 
     Returns:
-        Numpy array containing logits
+        Numpy array containing logits.
 
     Raises:
-        ValueError if min_epsilon is larger than max_epsilon
+        ValueError if min_epsilon is larger than max_epsilon.
     """
     # Get the height, width and number of channels
     height = x_data.shape[0]
--- a/cleverhans/utils.py
+++ b/cleverhans/utils.py
@@ -268,7 +268,7 @@ def get_logits_over_interval(sess, model, x_data, fgsm_params,
     epsilon = tf.reshape(tf.lin_space(float(min_epsilon),
                                       float(max_epsilon),
                                       num_points),
-						(num_points, 1, 1, 1))
+                         (num_points, 1, 1, 1))
     lin_batch = x + epsilon * eta
     logits = model.get_logits(lin_batch)
     with sess.as_default():
--- a/cleverhans/utils.py
+++ b/cleverhans/utils.py
@@ -228,26 +228,27 @@ def get_logits_over_interval(sess, model, x, adv_x, x_data,
     Args:
         sess: Tf session
         model: Model for which we wish to get logits
-        x: Tf placeholder for data
-        adv_x_real: Tf placeholder for adv perturbation of x
+        x: Data tensor
+        adv_x: Tensor for adversarial perturbation of x
         x_data: Numpy array corresponding to the data slice used to generate plot
         min_epsilon: Minimum value of epsilon over the interval
         max_epsilon: Maximum value of epsilon over the interval
         num_points: Number of points used to interpolate
 
     Returns:
-        Numpy array containing log probabilities
+        Numpy array containing logits
 
     Raises:
         ValueError if min_epsilon is larger than max_epsilon
     """
+    import tensorflow as tf
     if min_epsilon > max_epsilon:
         raise ValueError('Minimum epsilon is less than maximum epsilon')
 
-    delta = (max_epsilon - min_epsilon) / (num_points - 1)
     epsilon = min_epsilon
     eta = adv_x - x
-    for i, epsilon in enumerate(np.arange(min_epsilon, max_epsilon + 1, step=delta)):
+    eta = tf.nn.l2_normalize(eta, dim=0)
+    for i, epsilon in enumerate(np.linspace(min_epsilon, max_epsilon, num_points)):
         with sess.as_default():
             adv_x_epsilon = x + eta * epsilon
             logits = model.get_logits(adv_x_epsilon)
@@ -258,7 +259,6 @@ def get_logits_over_interval(sess, model, x, adv_x, x_data,
                 log_prob_adv_array = np.vstack((log_prob_adv_array,
                                                 log_prob_adv))
 
-        epsilon += delta
     return log_prob_adv_array
 
 
@@ -286,8 +286,7 @@ def linear_extrapolation_plot(log_prob_adv_array, y, file_name,
     fig = plt.figure()
     plt.xlabel('Epsilon')
     plt.ylabel('Log probabilities')
-    delta = (max_epsilon - min_epsilon) / (num_points - 1)
-    x_axis = np.arange(min_epsilon, max_epsilon + 1, step=delta)
+    x_axis = np.linspace(min_epsilon, max_epsilon, num_points)
     plt.xlim(min_epsilon - 1, max_epsilon + 1)
     for i in xrange(y.shape[1]):
         if i == correct_idx:
--- a/cleverhans/utils.py
+++ b/cleverhans/utils.py
@@ -220,15 +220,15 @@ def grid_visual(data):
     return figure
 
 
-def get_logits_over_interval(sess, model, x, adv_x, x_data,
-                             min_epsilon=-10, max_epsilon=10,
+def get_logits_over_interval(sess, model, x_data, fgsm_params,
+                             min_epsilon=-10., max_epsilon=10.,
                              num_points=21):
     """Get logits when the input is perturbed in an interval in adv direction.
 
     Args:
         sess: Tf session
         model: Model for which we wish to get logits
-        x: Data tensor of shape [None, height, width, channels]
+        x: Data tensor of shape [1, height, width, channels]
         adv_x: Tensor for adversarial perturbation of x
         x_data: Numpy array corresponding to single data
                 point of shape [height, width, channels].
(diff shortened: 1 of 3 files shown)


npapernot:
Thanks Aurko, unfortunately it seems like a few PEP8 errors are left

cleverhans/utils.py
```diff
@@ -220,6 +221,92 @@ def grid_visual(data):
     return figure
 
 
+def get_logits_over_interval(sess, model, x, adv_x_real, x_data,
```
npapernot: Could you add test(s) for `get_logits_over_interval`?

cleverhans/utils.py
```diff
@@ -4,6 +4,7 @@
 from __future__ import unicode_literals
 
 import numpy as np
+import tensorflow as tf
```
npapernot: At the moment, we should refrain from importing tensorflow at the top of files (because although theano is deprecated, we are still supporting for a few months --- and that's probably why the Travis tests failed on the Theano setup)

cleverhans/utils.py
```diff
@@ -220,6 +221,92 @@ def grid_visual(data):
...
+    """
+    delta = (max_epsilon - min_epsilon) / (num_points - 1)
+    epsilon = min_epsilon
+    eta = adv_x_real - x
+    eta_normalized = tf.nn.l2_normalize(eta, dim=0)
+
+    epsilon_t = tf.placeholder(tf.float32, shape=(), name="epsilon")
+    adv_x = x + epsilon_t * eta_normalized
+    logits = model.get_logits(adv_x)
+    for i in xrange(num_points):
```
npapernot: Rather than using a `for` loop on xrange(num_points), would it make sense to use `np.arange` to improve readability?

cleverhans/utils.py
```diff
@@ -220,6 +221,92 @@ def grid_visual(data):
...
+        max_epsilon: Maximum value of epsilon over the interval
+        num_points: Number of points used to interpolate
+
+    Returns:
+        Numpy array containing log probabilities
+    """
+    delta = (max_epsilon - min_epsilon) / (num_points - 1)
+    epsilon = min_epsilon
+    eta = adv_x_real - x
+    eta_normalized = tf.nn.l2_normalize(eta, dim=0)
```
npapernot: I may be missing something but is there a particular reason to use the l2 norm here?

royaurko:
I addressed your comments and added a test for the get logits function.

cleverhans/utils.py
```diff
@@ -220,6 +220,95 @@ def grid_visual(data):
...
+def get_logits_over_interval(sess, model, x, adv_x, x_data,
+                             min_epsilon=-10, max_epsilon=10,
+                             num_points=21):
+    """Get logits when the input is perturbed in an interval in adv direction.
+
+    Args:
+        sess: Tf session
+        model: Model for which we wish to get logits
+        x: Tf placeholder for data
+        adv_x_real: Tf placeholder for adv perturbation of x
```
goodfeli: The parameter is called `adv_x` but the docstring calls it `adv_x_real`

cleverhans/utils.py
```diff
@@ -220,6 +220,95 @@ def grid_visual(data):
...
 
+def get_logits_over_interval(sess, model, x, adv_x, x_data,
+                             min_epsilon=-10, max_epsilon=10,
+                             num_points=21):
+    """Get logits when the input is perturbed in an interval in adv direction.
+
+    Args:
+        sess: Tf session
+        model: Model for which we wish to get logits
+        x: Tf placeholder for data
```
goodfeli: Do these need to be placeholders? We should write most code to work with any tf tensor

cleverhans/utils.py
```diff
@@ -220,6 +220,95 @@ def grid_visual(data):
...
+        adv_x_real: Tf placeholder for adv perturbation of x
+        x_data: Numpy array corresponding to the data slice used to generate plot
+        min_epsilon: Minimum value of epsilon over the interval
+        max_epsilon: Maximum value of epsilon over the interval
+        num_points: Number of points used to interpolate
+
+    Returns:
+        Numpy array containing log probabilities
```
goodfeli: They're "logits" rather than "log probabilities", because if you take exp(output) it won't necessarily sum to 1.

cleverhans/utils.py
```diff
@@ -220,6 +220,95 @@ def grid_visual(data):
...
+
+    Raises:
+        ValueError if min_epsilon is larger than max_epsilon
+    """
+    if min_epsilon > max_epsilon:
+        raise ValueError('Minimum epsilon is less than maximum epsilon')
+
+    delta = (max_epsilon - min_epsilon) / (num_points - 1)
+    epsilon = min_epsilon
+    eta = adv_x - x
```
goodfeli: eta needs to be normalized in some way, otherwise you multiply the epsilon that was used inside the attack by the epsilon used in your for loop below.
For example, if `adv_x` came from an FGSM attack with eps=.3 and `max_epsilon` is 3, the largest adversarial perturbation we'd get in the for loop would be size .09

cleverhans/utils.py
```diff
@@ -220,6 +220,95 @@ def grid_visual(data):
...
+    for i, epsilon in enumerate(np.arange(min_epsilon, max_epsilon + 1, step=delta)):
```
goodfeli: why add 1 to max_epsilon?

goodfeli: it may be easier to do

    for i in xrange(num_points):
        epsilon = min_epsilon + (float(i) / (num_points - 1)) * (max_epsilon - min_epsilon)

I suspect the current implementation will sometimes return an array whose size is not `num_points` because of rounding error when you compute `delta`.

cleverhans/utils.py
```diff
@@ -220,6 +220,95 @@ def grid_visual(data):
...
+            adv_x_epsilon = x + eta * epsilon
+            logits = model.get_logits(adv_x_epsilon)
+            log_prob_adv = sess.run(logits, feed_dict={x: x_data})
+            if i == 0:
+                log_prob_adv_array = log_prob_adv
+            else:
+                log_prob_adv_array = np.vstack((log_prob_adv_array,
+                                                log_prob_adv))
+
+        epsilon += delta
```
goodfeli: You're incrementing `epsilon` by `delta` but then that immediately gets overwritten by the next iteration of the for loop

tests_tf/test_utils.py
```diff
@@ -79,6 +81,23 @@ def test_other_classes_return_val(self):
...
+        x = tf.placeholder(dtype=tf.float32, shape=(1, 28, 28, 1))
+        adv_x = x
+        img = np.ones(shape=(1, 28, 28, 1))
+        with tf.Session() as sess:
+            tf.global_variables_initializer().run()
+            log_prob = utils.get_logits_over_interval(sess, wrap,
+                                                      x, adv_x, img,
+                                                      min_epsilon=-10,
+                                                      max_epsilon=10,
+                                                      num_points=21)
```
goodfeli: Should add

    self.assertEqual(log_prob.shape[0], num_points)

tests_tf/test_utils.py
```diff
@@ -79,6 +81,23 @@ def test_other_classes_return_val(self):
...
+            log_prob -= log_prob[0]
+            self.assertTrue(np.count_nonzero(log_prob) == 0)
```
goodfeli: better to use `assertEqual` so that if  this fails, the travis log will show how many were nonzero

goodfeli: I don't understand this condition. It looks like you're testing that all the log probabilities are equal to each other. Why should that be the case?

tests_tf/test_utils.py
```diff
@@ -79,6 +81,23 @@ def test_other_classes_return_val(self):
         res_expected = [0, 1, 3, 4]
         self.assertTrue(res == res_expected)
 
+    def test_get_logits_over_interval(self):
+        import tensorflow as tf
+        model = cnn_model()
+        wrap = KerasModelWrapper(model)
+        x = tf.placeholder(dtype=tf.float32, shape=(1, 28, 28, 1))
+        adv_x = x
```
goodfeli: Feeding `x` for `adv_x` should probably cause some divide by zero error or something because it means that the linear cross section plot doesn't have a direction to explore

cleverhans/utils.py
royaurko: Fixed.

cleverhans/utils.py
royaurko: No they would work for any tensor. I updated the docstring.

cleverhans/utils.py
royaurko: Done.

cleverhans/utils.py
royaurko: I had a tf.nn.l2_normalize on eta, but took it out since Nicolas said importing tensorflow at the top level caused the build to fail. I imported it within the function def.

cleverhans/utils.py
royaurko: because it wouldn't take the max_epsilon value otherwise, I switched to using np.linspace which is recommended for floats.

royaurko: I had a similar implementation before, but Nicolas thought it might be more intuitive to have a np.arange. I replaced it with np.linspace which takes care of the rounding issues.

cleverhans/utils.py
royaurko: Fixed, no need to increment epsilon since we take it from np.linspace

tests_tf/test_utils.py
royaurko: Because eta = 0, so the model is just fed x. I removed this test and just check for shape.

cleverhans/utils.py
```diff
@@ -220,6 +220,97 @@ def grid_visual(data):
...
 
+def get_logits_over_interval(sess, model, x, adv_x, x_data,
+                             min_epsilon=-10, max_epsilon=10,
+                             num_points=21):
+    """Get logits when the input is perturbed in an interval in adv direction.
+
+    Args:
+        sess: Tf session
+        model: Model for which we wish to get logits
+        x: Data tensor
```
goodfeli: I realized I'm confused about the interface here.
Is x meant to be a minibatch or a single example? I would expect it to be a single example.
Is the user required to make it be a minibatch of size 1?

cleverhans/utils.py
```diff
@@ -220,6 +220,97 @@ def grid_visual(data):
...
+        ValueError if min_epsilon is larger than max_epsilon
+    """
+    import tensorflow as tf
+    if min_epsilon > max_epsilon:
+        raise ValueError('Minimum epsilon is less than maximum epsilon')
+
+    epsilon = min_epsilon
+    eta = adv_x - x
+    eta = tf.nn.l2_normalize(eta, dim=0)
+    for i, epsilon in enumerate(np.linspace(min_epsilon,
```
goodfeli: if x is a single example, it could be possible to do this without a for loop, and with only 1 sess.run call rather than 21. You would make one batch containing all the different sizes of perturbation and run them in a single sess.run call.

cleverhans/utils.py
```diff
@@ -220,6 +220,97 @@ def grid_visual(data):
...
+        y: Tf placeholder for the labels
+        file_name: Plot filename
+        min_epsilon: Minimum value of epsilon over the interval
+        max_epsilon: Maximum value of epsilon over the interval
+        num_points: Number of points used to interpolate
+    """
+    import matplotlib.pyplot as plt
+
+    # Ensure interactive mode is disabled and initialize our graph
+    plt.ioff()
```
goodfeli: why does interactive mode need to be off? can we implement this function without having side effects on other code that uses pyplot?

cleverhans/utils.py
royaurko: I have updated the code so that x is now supposed to be of shape [height, width, channels]

cleverhans/utils.py
royaurko: Done.

cleverhans/utils.py
royaurko: I updated it do it over one sess call. However the lep now looks like this: 

lep_adv3

As opposed to this:
lep_adv

I am wondering if there is an issue in the way I do it.

npapernot:
Aurko, you may have to amend this last commit with your @google.com email address because the CLA bot detected you as not having signed the CLA.

goodfeli:
I think it's OK to ignore Googlebot since we know who you are and know Googlebot is mistaken, but you should fix the Travis tests. They're currently failing due to a float32 vs float64 error. I can review further once the tests pass.

npapernot:
```
ERROR: test_get_logits_over_interval (test_utils.TestUtils)
  File "/home/travis/build/tensorflow/cleverhans/tests_tf/test_utils.py", line 99, in test_get_logits_over_interval
    num_points=num_points)
  File "/home/travis/build/tensorflow/cleverhans/cleverhans/utils.py", line 267, in get_logits_over_interval
    adv_x_epsilon = x + eta * epsilon_array_t
  File "/home/travis/miniconda/envs/test-environment/lib/python3.5/site-packages/tensorflow/python/ops/math_ops.py", line 820, in binary_op_wrapper
    y = ops.convert_to_tensor(y, dtype=x.dtype.base_dtype, name="y")
  File "/home/travis/miniconda/envs/test-environment/lib/python3.5/site-packages/tensorflow/python/framework/ops.py", line 639, in convert_to_tensor
    as_ref=False)
  File "/home/travis/miniconda/envs/test-environment/lib/python3.5/site-packages/tensorflow/python/framework/ops.py", line 704, in internal_convert_to_tensor
    ret = conversion_func(value, dtype=dtype, name=name, as_ref=as_ref)
  File "/home/travis/miniconda/envs/test-environment/lib/python3.5/site-packages/tensorflow/python/framework/ops.py", line 577, in _TensorTensorConversionFunction
    % (dtype.name, t.dtype.name, str(t)))
ValueError: Tensor conversion requested dtype float32 for Tensor with dtype float64: 'Tensor("Reshape_400:0", shape=(21, 28, 28, 1), dtype=float64)'
```

cleverhans/utils.py
```diff
@@ -220,6 +220,101 @@ def grid_visual(data):
...
 
+def get_logits_over_interval(sess, model, x, adv_x, x_data,
+                             min_epsilon=-10, max_epsilon=10,
+                             num_points=21):
+    """Get logits when the input is perturbed in an interval in adv direction.
+
+    Args:
+        sess: Tf session
+        model: Model for which we wish to get logits
+        x: Data tensor of shape [None, height, width, channels]
```
goodfeli: If the user has to pass in `model`, why not just make your own `x`? It really only makes sense to pass in `x` if the user also has to pass in the logits corresponding to `x`.

goodfeli: We probably don't want it to be shape `None` because you're relying on the user passing in exactly one example. To enforce that, it should be shape 1.

cleverhans/utils.py
```diff
@@ -220,6 +220,101 @@ def grid_visual(data):
...
+        Numpy array containing logits
+
+    Raises:
+        ValueError if min_epsilon is larger than max_epsilon
+    """
+    import tensorflow as tf
+    if min_epsilon > max_epsilon:
+        raise ValueError('Minimum epsilon is less than maximum epsilon')
+
+    eta = tf.nn.l2_normalize(adv_x - x, dim=0)
```
goodfeli: You should be able to keep this in TensorFlow rather than NumPy. Also, you should be able to do everything using broadcasting, rather than building a larger array.

Something like

     epsilon = tf.reshape(tf.lin_space(min_epsilon, max_epsilon, num_points), (num_points, 1, 1, 1))
     lin_batch = x + epsilon * eta

should be able to do it without any explicit duplication of x and eta.

cleverhans/utils.py
```diff
@@ -220,6 +220,101 @@ def grid_visual(data):
...
+    """
+    import matplotlib.pyplot as plt
+
+    figure = plt.figure()
+    figure.canvas.set_window_title('Cleverhans: Linear Extrapolation Plot')
+
+    correct_idx = np.argmax(y, axis=0)
+    fig = plt.figure()
+    plt.xlabel('Epsilon')
+    plt.ylabel('Log probabilities')
```
goodfeli: They're actually logits, not log probabilities, right?

cleverhans/utils.py
royaurko: Done

cleverhans/utils.py
royaurko: Fixed

cleverhans/utils.py
royaurko: Done

Issue #116, opened 2017-03-26
Implement JSMA symbolically 


npapernot:
Improve JSMA implementation so we can feed it more than one input at a time

IvanKush:
do you mean multithreading, etc?

npapernot:
Sorry the title of this is somewhat unrelated to the description now that I changed it. The goal was to implement the JSMA completely symbolically in TF (no Python/Numpy involved). This requires among other things to change the current Python while loop into a TF while loop.

ajing:
Sorry for my limited knowledge about Tensorflow. For the part like:

```
        search_domain = set([i for i in xrange(nb_features)
                             if adv_x[0, i] < clip_max])
```

Could this also be written symbolically?

npapernot:
It should be possible to rewrite this symbolically yes. I contacted @gongzhitao who implemented a version of this here: https://github.com/gongzhitaao/tensorflow-adversarial/blob/master/attacks/jsma.py#L102

There should be a PR to work on this soon.

gongzhitaao:
In the `test_attacks.py`, where is the `sess.run(tf.global_variables_initializer())`?

npapernot:
For instance here: https://github.com/tensorflow/cleverhans/blob/master/tests_tf/test_attacks.py#L461

PR #275, opened 2017-09-24
Symbolic implementation of JSMA

gongzhitaao:
1. This is for issue #116 
2. Test cases passed, have not test the tutorial code yet.

Please let me know if you have suggestions and comments.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -468,15 +468,15 @@ class BasicIterativeMethod(Attack):
 
 
 class SaliencyMapMethod(Attack):
-
     """
-    The Jacobian-based Saliency Map Method (Papernot et al. 2016).
-    Paper link: https://arxiv.org/pdf/1511.07528.pdf
+    The Jacobian-based Saliency Map Method (Papernot et al.  2016).  Paper link:
+    https://arxiv.org/pdf/1511.07528.pdf
     """
 
     def __init__(self, model, back='tf', sess=None):
         """
         Create a SaliencyMapMethod instance.
+
         Note: the model parameter should be an instance of the
         cleverhans.model.Model abstraction provided by CleverHans.
         """
@@ -490,80 +490,42 @@ class SaliencyMapMethod(Attack):
             raise NotImplementedError(error)
 
         import tensorflow as tf
-        self.feedable_kwargs = {'y_target': tf.float32}
-        self.structural_kwargs = ['theta', 'gamma',
+        self.feedable_kwargs = {'y_target': tf.int32}
+        self.structural_kwargs = ['theta', 'epochs', 'pair', 'min_proba',
                                   'clip_max', 'clip_min']
 
     def generate(self, x, **kwargs):
         """
-        Generate symbolic graph for adversarial examples and return.
+        Generate symbolic graph for adversarial examples.
+
         :param x: The model's symbolic inputs.
+        :param y_target: (optional) Target tensor if the attack is targeted
         :param theta: (optional float) Perturbation introduced to modified
-                      components (can be positive or negative)
-        :param gamma: (optional float) Maximum percentage of perturbed features
+            components (can be positive or negative)
+        :param epochs: (optional, int or float) maximum epochs to run.  If this
+            is a floating number between [0, 1], then it is treated as gamma,
+            the distortion factor in the paper, and the max epochs is
+            automatically calculated.
+        :param pair: If true, add noise to two pixels at a time.  Otherwise one
+            pixel at a time.
+        :param min_proba: The minimum desired probability that a model will
+            wrongly classify the adversarials.
         :param clip_min: (optional float) Minimum component value for clipping
         :param clip_max: (optional float) Maximum component value for clipping
-        :param y_target: (optional) Target tensor if the attack is targeted
-        """
-        import tensorflow as tf
-        from .attacks_tf import jacobian_graph, jsma_batch
-
-        # Parse and save attack-specific parameters
-        assert self.parse_params(**kwargs)
-
-        # Define Jacobian graph wrt to this input placeholder
-        preds = self.model.get_probs(x)
-        nb_classes = preds.get_shape().as_list()[-1]
-        grads = jacobian_graph(preds, x, nb_classes)
-
-        # Define appropriate graph (targeted / random target labels)
-        if self.y_target is not None:
-            def jsma_wrap(x_val, y_target):
-                return jsma_batch(self.sess, x, preds, grads, x_val,
-                                  self.theta, self.gamma, self.clip_min,
-                                  self.clip_max, nb_classes,
-                                  y_target=y_target)
-
-            # Attack is targeted, target placeholder will need to be fed
-            wrap = tf.py_func(jsma_wrap, [x, self.y_target], tf.float32)
-        else:
-            def jsma_wrap(x_val):
-                return jsma_batch(self.sess, x, preds, grads, x_val,
-                                  self.theta, self.gamma, self.clip_min,
-                                  self.clip_max, nb_classes,
-                                  y_target=None)
-
-            # Attack is untargeted, target values will be chosen at random
-            wrap = tf.py_func(jsma_wrap, [x], tf.float32)
-
-        return wrap
-
-    def parse_params(self, theta=1., gamma=np.inf, nb_classes=None,
-                     clip_min=0., clip_max=1., y_target=None, **kwargs):
-        """
-        Take in a dictionary of parameters and applies attack-specific checks
-        before saving them as attributes.
-
-        Attack-specific parameters:
-        :param theta: (optional float) Perturbation introduced to modified
-                      components (can be positive or negative)
-        :param gamma: (optional float) Maximum percentage of perturbed features
-        :param nb_classes: (optional int) Number of model output classes
-        :param clip_min: (optional float) Minimum component value for clipping
-        :param clip_max: (optional float) Maximum component value for clipping
-        :param y_target: (optional) Target tensor if the attack is targeted
         """
 
-        if nb_classes is not None:
-            warnings.warn("The nb_classes argument is depricated and will "
-                          "be removed on 2018-02-11")
-        self.theta = theta
-        self.gamma = gamma
-        self.clip_min = clip_min
-        self.clip_max = clip_max
-        self.y_target = y_target
+        y = kwargs.get('y_target', 0)
+        epochs = kwargs.get('epochs', 0.2)
+        theta = kwargs.get('theta', 1.0)
+        pair = kwargs.get('pair', True)
+        min_proba = kwargs.get('min_proba', 0.0)
+        clip_min = kwargs.get('clip_min', 0.0)
+        clip_max = kwargs.get('clip_max', 1.0)
 
-        return True
+        from .attacks_tf import jsma
+
+        return jsma(self.model, x, y, epochs=epochs, eps=theta, pair=pair,
+                    min_proba=min_proba, clip_min=clip_min, clip_max=clip_max)
 
 
 class VirtualAdversarialMethod(Attack):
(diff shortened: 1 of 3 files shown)


gongzhitaao:
I signed it!

gongzhitaao:
I think I need to rewrite large part of `mnist_tutorial_jsma.py` using the new interface.

npapernot:
Thank you for the PR! The tests are failing because of the PEP8 checker. Here is a dump of the test log:

```
FAIL: Test if pep8 is respected.
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/nose-1.3.7-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/tensorflow/cleverhans/cleverhans/devtools/tests/test_format.py", line 45, in test_format_pep8
    raise AssertionError("PEP8 Format not respected")
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks_tf.py:361:14: E127 continuation line over-indented for visual indent
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:883:80: E501 line too long (81 > 79 characters)
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:887:80: E501 line too long (82 > 79 characters)
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:890:80: E501 line too long (81 > 79 characters)
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:891:80: E501 line too long (82 > 79 characters)
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:895:80: E501 line too long (81 > 79 characters)
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:897:80: E501 line too long (81 > 79 characters)
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:903:80: E501 line too long (80 > 79 characters)
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:907:80: E501 line too long (80 > 79 characters)
--------------------- >> end captured stdout << ----------------------
```

gongzhitaao:
Yes, I saw the PEP8 log, weird is that it is 79 instead of 80 :rofl: .  I will update it as well as the `mnist_tutorial_jsma.py`.

carlini:
Why does the tutorial need to be rewritten? Changing the underlying implementation shouldn't impact the interface.

Also it would be good to benchmark the speed of the two implementations to make sure they're comparable. One of the good qualities of JSMA is it's fast.

gongzhitaao:
@carlini Since I change the JSMA API a bit.

1. `gamma` -> `epochs`, you could either specify max epochs to run or if epochs is a floating number, it is treat as gamma and max epochs is automatically determined.
2. `y_target` takes the class label(s) **not** one-hot encoded.  Rationale is that during JSMA the one-hot encoded labels are converted back to class labels and the one-hot labels are never used.
3. Add `pair`, If `True`, perturb two pixels at a time, otherwise one pixel at a time.
4. Add `min_proba`, denoting the minimum desired confidence for the labels to be wrongly classified.

Besides, JSMA now takes a batch of images, the 2-level nested for loops in tutorials could be reduced to a single for loop.

npapernot:
Thanks again for putting together this PR.

Could you elaborate on why 1. is needed?

Regarding 2., if we are going to change this, we should at least be backward compatible (it should require fairly low overhead) and then show a deprecation warning. 

Argument 3. could perhaps be improved: why not specify a number of pixels perturbed per iteration instead?

npapernot:
If you have spare time, could you compare the success rate + average perturbation of the non-symbolic and symbolic implementations on MNIST for instance so we have an idea of any changes in performances?

cleverhans/attacks.py
```diff
@@ -818,36 +780,34 @@ def vatm(model, x, logits, eps, back='tf', num_iterations=1, xi=1e-6,
...
-    warnings.warn("attacks.jsma is deprecated and will be removed on "
-                  "2017-09-27. Instantiate an object from SaliencyMapMethod.")
-    if back == 'tf':
-        # Compute Jacobian-based saliency map attack using TensorFlow
-        from .attacks_tf import jsma
-        return jsma(sess, x, predictions, grads, sample, target, theta, gamma,
-                    clip_min, clip_max)
-    elif back == 'th':
-        raise NotImplementedError("Theano jsma not implemented.")
+# def jsma(x, predictions, targets, theta, epochs, back='tf', clip_min=None,
```
npapernot: Instead of commenting out the code, it's best to remove it completely if we're not going to use it.

gongzhitaao: Sure, I saw the deprecating date was 9/27, so I will remove it from the code in the next PR.

gongzhitaao:
@npapernot Here is some of my thoughts
1. My rationale is that if you have N pixels, with `gamma` the algorithm runs maximum N epochs.  What if you want to run more than N epochs?  And besides, `epochs` is more straightforward to me.  But yes, I agree it is debatable, if you still think `gamma` makes more sense than `epochs`,  I would change it back in the next PR.
2. Yes, a deprecating warning is definitely more friendly.
3. For this it would be much more difficult.  As pointed out in your paper, the complexity is `O(n^k)` where `k` is the number of pixels to perturb at a time.  Naively you could just use loops but that will be inefficient.  Currently the `_jsma_impl` and `_jsma2_impl` are optimized for `k=1` and `k=2` cases by vectoring the operations as much as possible in mini batches.  I could write a version for general `k` values but IMHO, it is unnecessary.  We could discuss this if you want.
4. Yes, I would submit a benchmark in the next PR.

npapernot:
1. The main reason why I would be in favor of gamma, is that people who read the paper might be confused if there is a different parameter in the implementation on CleverHans.
2. Thanks!
3. Agreed, it might still make sense to have the general version but indeed the optimized ones are welcomed.
4. Unfortunately, we cannot merge this if we don't have any guarantees of how well it performs in comparison with the non-symbolic implementation because users will start benchmarking with the implementation that was merged.

cleverhans/attacks.py
```diff
@@ -490,80 +490,42 @@ def __init__(self, model, back='tf', sess=None):
             raise NotImplementedError(error)
 
         import tensorflow as tf
-        self.feedable_kwargs = {'y_target': tf.float32}
-        self.structural_kwargs = ['theta', 'gamma',
+        self.feedable_kwargs = {'y_target': tf.int32}
+        self.structural_kwargs = ['theta', 'epochs', 'pair', 'min_proba',
```
npapernot: In the rest of the library, `prob` and `probs` have been used to refer to probabilities so it would be nice to keep things consistent here as well.

cleverhans/attacks.py
```diff
@@ -490,80 +490,42 @@ def __init__(self, model, back='tf', sess=None):
...
-        if nb_classes is not None:
-            warnings.warn("The nb_classes argument is depricated and will "
-                          "be removed on 2018-02-11")
-        self.theta = theta
-        self.gamma = gamma
-        self.clip_min = clip_min
-        self.clip_max = clip_max
-        self.y_target = y_target
+        y = kwargs.get('y_target', 0)
+        epochs = kwargs.get('epochs', 0.2)
```
npapernot: See comment in discussion about `epochs`

cleverhans/attacks.py
```diff
@@ -490,80 +490,42 @@ def __init__(self, model, back='tf', sess=None):
...
+        theta = kwargs.get('theta', 1.0)
+        pair = kwargs.get('pair', True)
```
npapernot: See comment about generalizing to `k` in discussion. For this PR, it's OK to have the code throw an error when `k>2`

cleverhans/attacks.py
```diff
@@ -490,80 +490,42 @@ def __init__(self, model, back='tf', sess=None):
...
 
-        if nb_classes is not None:
-            warnings.warn("The nb_classes argument is depricated and will "
-                          "be removed on 2018-02-11")
-        self.theta = theta
-        self.gamma = gamma
-        self.clip_min = clip_min
-        self.clip_max = clip_max
-        self.y_target = y_target
+        y = kwargs.get('y_target', 0)
```
npapernot: Could you follow the style adopted by all other attack and have the parameters parsed in the method `parse_params`?

cleverhans/attacks.py
npapernot: It's OK to remove it in this PR directly.

cleverhans/attacks_tf.py
```diff
@@ -133,275 +133,248 @@ def vatm(model, x, logits, eps, num_iterations=1, xi=1e-6,
...
-
-    # Create a 2D numpy array of the sum of grads_target and grads_other
-    target_sum = grads_target.reshape((1, nf)) + grads_target.reshape((nf, 1))
-    other_sum = grads_other.reshape((1, nf)) + grads_other.reshape((nf, 1))
-
-    # Create a mask to only keep features that match saliency map conditions
-    if increase:
-        scores_mask = ((target_sum > 0) & (other_sum < 0))
+    xshape = tf.shape(x)
+    n = xshape[0]
```
npapernot: `xshape` is only used in this line right? Can you just combine the two lines?

cleverhans/attacks_tf.py
```diff
@@ -133,275 +133,248 @@ def vatm(model, x, logits, eps, num_iterations=1, xi=1e-6,
...
+    target = tf.cond(tf.equal(0, tf.rank(y)),
+                     lambda: tf.zeros([n], dtype=tf.int32)+y,
+                     lambda: y)
+
+    if isinstance(epochs, float):
+        tmp = tf.to_float(tf.size(x[0])) * epochs
+        epochs = tf.to_int32(tf.floor(tmp))
+
+    if pair:
+        _jsma_fn = _jsma2_impl
```
npapernot: Could you make the name of `_jsma2_impl` more descriptive? Like `_jsma_pair_impl`

cleverhans/attacks_tf.py
```diff
@@ -133,275 +133,248 @@ def vatm(model, x, logits, eps, num_iterations=1, xi=1e-6,
...
     else:
-        scores_mask = ((target_sum < 0) & (other_sum > 0))
+        _jsma_fn = _jsma_impl
 
-    # Create a 2D numpy array of the scores for each pair of candidate features
-    scores = scores_mask * (-target_sum * other_sum)
+    def _fn(i):
+        # `xi` is of the shape (1, ....), the first dimension is the number of
```
npapernot: This comment is not entirely clear. Could you rephrase a bit? In its current form, it implies that the first dimension is always 1 because the example shape is (1, ...)

cleverhans/attacks_tf.py
```diff
@@ -133,275 +133,248 @@ def vatm(model, x, logits, eps, num_iterations=1, xi=1e-6,
...
 
-    # Remove used pixels from our search domain
-    search_domain.discard(p1)
-    search_domain.discard(p2)
 
-    return p1, p2, search_domain
+def _jsma_impl(model, xi, yi, epochs, eps=1.0, clip_min=0.0, clip_max=1.0,
+               min_proba=1.1):
 
+    def _cond(x_adv, epoch, pixel_mask):
```
npapernot: Can you document this function, at least briefly? My understanding is that it checks for two things: (1) the target label is achieved, (2) with enough confidence (i.e., min_proba < proba)

cleverhans/attacks_tf.py
```diff
@@ -133,275 +133,248 @@ def vatm(model, x, logits, eps, num_iterations=1, xi=1e-6,
...
+        ybar = tf.reshape(model.get_probs(x_adv), [-1])
+        proba = ybar[yi]
+        label = tf.to_int32(tf.argmax(ybar, axis=0))
+        return tf.reduce_all([tf.less(epoch, epochs),
+                              tf.reduce_any(pixel_mask),
+                              tf.logical_or(tf.not_equal(yi, label),
+                                            tf.less(proba, min_proba))],
+                             name='_jsma_step_cond')
+
+    def _body(x_adv, epoch, pixel_mask):
```
npapernot: Same here, please document at least briefly what this functions are doing.

npapernot:
@gongzhitaao Did you get a chance to address the comments?

gongzhitaao:
@npapernot Yes.  I'm actually testing a new implementation of JSMA.  The current implementation uses a `map_fn` function to simulate the batch-JSMA which is actually slow.  I'm reimplementing the `_jsma_impl` in a batched way.  Besides I'm working on some benchmarks showing that it really works as expected.  I think I will update the PR in early November.

catherio:
@gongzhitaao Any updates? No rush, just want to make sure this is still on your radar, and check in about whether there's anything we can help with.

gongzhitaao:
@catherio Yes. LOL A bit slow, but yes, still working on the features

cleverhans/attacks.py
gongzhitaao: Quick question: why a separate `parse_params`?  So far I could see it does nothing but just assigning parameters, besides its documents is exactly the same as `__init__`

cleverhans/attacks.py
gongzhitaao: I will do that.  And I'm not considering implementing a general case.  Since it is unlikely that JSMA will be anything practical for `k>2` case.  For an image with `28x28`, each iteration goes through ~1M pixels for `k=2`, ~1B pixels for `k=3`.

catherio:
Ok, no worries!

Issue #277, opened 2017-09-26
Move clip eta to abstract Attack class


npapernot:
It might make sense to move the method `clip_eta(self, eta)` defined in the Madry et al. class to the abstract Attack class since most attacks need this clipping step at some point.

goodfeli:
I think we should keep the `Attack` class as close to abstract as possible, but we could maybe put a `clip_eta` function in `utils_tf.py`.

PR #279, opened 2017-09-26
Move clip_eta to utils_tf.py

yenchenlin:
According to #277, moving `clip_eta` function to `utils_tf.py` can help other algorithms reuse the function.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -1043,31 +1043,6 @@ class MadryEtAl(Attack):
 
         return True
 
-    def clip_eta(self, eta):
-        """
-        Clip the perturbation to epsilon l-infinity ball.
-
-        :param eta: A tensor with the current perturbation.
-        """
-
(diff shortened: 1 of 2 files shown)


goodfeli:
Thanks!

yenchenlin:
BTW, do you think adding a else for [this if else statement](https://github.com/tensorflow/cleverhans/pull/279/files#diff-ab6a61cfd3fc0e1d38198eb77671b68bR434) in `clip_eta` to launch an error when user accidentally pass value which does not belong to [np.inf, 1, 2] is uesful? @goodfeli

goodfeli:
Yes. Right now, an invalid value of `ord` will cause an error when `norm`
is referenced without being defined. We could add an else to make than
explicit `NotImplementedError` or possibly `ValueError`. We could also
think about implementing a generic L_p norm for any numeric type of `ord`
argument, but I don't know of a use case for that yet.

PR #285, opened 2017-09-28
Raise value error if `ord` value is invalid in `clip_eta` function

yenchenlin:
According to @279, it's better to notify users when `ord` value is not valid.

### DIFF
--- a/cleverhans/utils_tf.py
+++ b/cleverhans/utils_tf.py
@@ -403,6 +403,8 @@ def clip_eta(eta, ord, eps):
     """
 
     # Clipping perturbation eta to self.ord norm ball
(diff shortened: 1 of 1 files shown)


goodfeli:
This looks good, and any of the CleverHans team should feel free to merge it as soon as the tests pass. Two of the test jobs failed because the Theano installer failed to download MKL. I doubt that they will have any problem when rerun.

npapernot:
For some reason, Travis fails to setup the Theano vm. It errored while trying to install numpy.

yenchenlin:
Thanks @npapernot for solving the CI issue!
I think this can be merged?

PR #286, opened 2017-09-28
EAD Implementation

ysharma1126:
This attack builds on the C&W L2 attack by generalizing it to the elastic-net case, where L1-oriented adversarial examples can be generated. These examples were shown to be both state-of-the-art and distinct in various attack scenarios, and more importantly, were shown to have improved transferability.

Paper link: https://arxiv.org/abs/1709.04114


npapernot:
Unfortunately, I think you need to amend your commits in order to fix the email address. This should explain how: https://www.google.com/search?q=git+change+commit+email

PR #288, opened 2017-09-28
EAD Implementation

ysharma1126:
This attack builds on the C&W L2 attack by generalizing it to the elastic-net case, where L1-oriented adversarial examples can be generated. These examples were shown to be both state-of-the-art and distinct in various attack scenarios, and more importantly, were shown to have improved transferability. 

Paper link: https://arxiv.org/abs/1709.04114


ysharma1126:
The build failed on the Theano versions, no clue why, but isn't that fine as Theano support will be deprecated?

npapernot:
Yes, we are looking into it. The CLA googlebot looks good now, thank you! I'll ping you once we figure out the bug with Theano tests

npapernot:
Thank you for the PR, the tests are now fixed. We've had a lot of PRs to review lately so it might take a bit of time before someone gets to this but we will as soon as possible.

goodfeli:
This looks good. Since you're a co-author of the paper, we'll rely on you to make sure it's a correct implementation of the method you developed, and we'll mostly just review for style and maintainability.

Can you please add some tests? They don't need to be extremely exhaustive tests, but there should be at least one test per new class.

ysharma1126:
Hi Ian,

Thanks. I tested my attack by replacing a few lines of code in
mnist_tutorial_cw.py. How should I add those results as tests? Make a new
tutorial for this attack and show the output?

Let me know how best this should be done.

Thanks,

Yash Sharma

npapernot:
Hi Yash,

You can take a look at some of the tests implemented for other attacks such as this one: https://github.com/tensorflow/cleverhans/blob/master/tests_tf/test_attacks.py#L112

ysharma1126:
Thanks Nicolas. Added a test, locally tested and the attack passes the test. Waiting on the build.

ysharma1126:
Another error that doesn't appear to have anything to do with my method, can you please check?

goodfeli:
Yes, it seems like that is just the test being flaky. Can you send a commit dropping the threshold for that test to .93?

ysharma1126:
Sure

npapernot:
I loosened up the other test as well because it was still failing on some of Travis builds. OK to ignore the CLA error, it happened after I used the web interface on GitHub to edit the code.

PR #289, opened 2017-09-29
MadryEtAl tutorial

AngusG:
Similar to `mnist_tutorial_tf.py` only using MadryEtAl attack

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -1053,7 +1053,7 @@ class MadryEtAl(Attack):
         :param y: A tensor with the target labels or ground-truth labels.
         """
         import tensorflow as tf
-        from utils_tf import model_loss, clip_eta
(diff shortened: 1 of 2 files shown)


goodfeli:
Hi Angus,
Please write to cleverhans-dev before starting work on new features, to make sure you spend time only on features we plan to merge. We do not want to create more tutorials for each attack, because it is a large amount of duplicated code to maintain and test.
-Ian

goodfeli:
If you'd like to work on something like a generic trainer script or a generic attack script that can run attacks on a saved model, that could be configured to give results for the Madry attack. Please free free to discuss further on cleverhans-dev.

AngusG:
Hi Ian, that sounds like a good idea. I already have something like a generic attack script I could share. I will note to communicate on cleverhans-dev for next time.

npapernot:
Thank you Angus!

PR #302, opened 2017-10-21
PGD Targeted Attack fix

ysharma1126:
Quick fix for executing Madry PGD in the targeted case. 

Quick question: As proposed in the 'Delving into Transferable Adversarial Examples' paper, executing attacks on an ensemble models normally generates more readily transferable adversarial examples. For some tests I'm doing, I've modified the Cleverhans attacks I've been using to rather than attack a single model, attack a list of models. Therefore, if you wanted to execute a simple attack on a single model, you would pass that model in as a one-element list. 

Would this code be of interest to be integrated into Cleverhans? More generally, I do think code which makes it easier to perform attacks on an ensemble of models should be integrated.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -1203,7 +1203,9 @@ class MadryEtAl(Attack):
         eta = tf.random_uniform(tf.shape(x), -self.eps, self.eps)
         eta = clip_eta(eta, self.ord, self.eps)
 
-        if self.y is not None:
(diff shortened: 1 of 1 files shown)


npapernot:
Thanks for the PR! 

It's best to discuss the question (regarding the new feature) on the CleverHans mailing list `cleverhans-dev@googlegroups.com`, which you can ask to join.

Thanks!

Issue #281, opened 2017-09-27
check model load / save functions


npapernot:
remove FLAGS from `utils_tf.py`

yenchenlin:
Hello @npapernot ,

It seems that now the only FLAGS in `utils_tf.py` exists in function `tf_model_load`, and this function is only called in `mnist_tutorial_cw.py`.

I deleted the FLAGS in function `tf_model_load` and confirmed that `mnist_tutorial_cw.py` still works fine. Does it mean we can simply add a deprecated message in function `tf_model_load` for FLAGS and completely delete FLAGS from `utils_tf.py` in the near future?

npapernot:
Yes, we could deprecate the logic of this if statement https://github.com/tensorflow/cleverhans/blob/master/cleverhans/utils_tf.py#L276 and require that the `file_path` argument be supplied.

catherio:
@npapernot Is this issue resolved now that #303 has been merged?

npapernot:
I don't think this is resolved completely, I believe Aurko is working on it though.

catherio:
@royaurko Any updates?

PR #303, opened 2017-10-23
Deprecate FLAGS from utils_tf.py

yenchenlin:
According to #281 

Maybe we can directly remove the `if` logic here?

cc @npapernot 

Thanks!

### DIFF
--- a/cleverhans/utils_tf.py
+++ b/cleverhans/utils_tf.py
@@ -274,6 +274,9 @@ def tf_model_load(sess, file_path=None):
     with sess.as_default():
         saver = tf.train.Saver()
         if file_path is None:
+            warnings.warn("Please provide file_path argument, "
+                          "support for FLAGS.train_dir and FLAGS.filename "
+                          "will be removed on 2018-04-23.")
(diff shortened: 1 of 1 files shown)


cleverhans/utils_tf.py
```diff
@@ -274,6 +274,9 @@ def tf_model_load(sess, file_path=None):
     with sess.as_default():
         saver = tf.train.Saver()
         if file_path is None:
+            warnings.warn("FLAGS argument is deprecated. "
```
npapernot: Can you remove this part of the warning?

cleverhans/utils_tf.py
```diff
@@ -274,6 +274,9 @@ def tf_model_load(sess, file_path=None):
...
+                          "Please provide file_path argument. "
+                          "FLAGS argument will be removed after 2018-04-23.")
```
npapernot: Can you change the `FLAGS argument` part to `FLAGS.train_dir` and `FLAGS.filename`

yenchenlin:
@npapernot thanks, done!

PR #304, opened 2017-10-27
Fix softmax identification in model_loss

carlini:
model_loss determines softmax by looking for the string 'softmax' in the inputs to the tensor. Inception (and others) uses 'pre_softmax' for the logits, breaking it. This is a specific workaround for those default models.


carlini:
Actually, I decided this probably is a bad idea. Let's not do this. It's better to have people just use the model interface.

Issue #276, opened 2017-09-26
Create better tests for the MadryEtAl attack


npapernot:
@fartashf is working on this.

goodfeli:
Completed in #305

PR #305, opened 2017-10-30
Tests for Madryetal attack, Addressing #276

fartashf:


cleverhans/attacks.py
```diff
@@ -1182,12 +1182,13 @@ def attack_single_step(self, x, eta, y):
...
         adv_x = adv_x + scaled_signed_grad
-        adv_x = tf.clip_by_value(adv_x, self.clip_min, self.clip_max)
+        if self.clip_min is not None and self.clip_max is not None:
+            adv_x = tf.clip_by_value(adv_x, self.clip_min, self.clip_max)
         eta = adv_x - x
         eta = clip_eta(eta, self.ord, self.eps)
         return x, eta
 
-    def attack(self, x, **kwargs):
+    def attack(self, x, y, **kwargs):
```
npapernot: It would be best to pass y as one of the kwargs, to keep the unified interface across different attacks.

tests_tf/test_attacks.py
```diff
@@ -678,17 +679,96 @@ def test_generate_gives_adversarial_example(self):
...
+        """
+        This test will fail if an initial random step is not taken.
+        """
+        x_val = np.random.rand(100, 2)
+        x_val = np.array(x_val, dtype=np.float32)
+
+        orig_labs = np.argmax(self.sess.run(self.model(x_val)), axis=1)
+        new_labs_multi = orig_labs.copy()
+
+        for i in range(10):
```
npapernot: Can you add a one liner comment describing what the test is doing to test that an initial random step was taken? From what I'm understanding it's replacing all labels that were identical in the previous iteration by the labels predicted on adversarial examples computed on the current iteration.

cleverhans/attacks.py
fartashf: I think you are comparing to the interface for `generate` function but this line defines the `attack` function which is special to the MadryEtAl and similar to function `fgm` here: https://github.com/tensorflow/cleverhans/blob/cca980212c07b952d0849e3df4d59b6185e3ce91/cleverhans/attacks_tf.py#L23

The kwargs was unnecessary and I removed it in 0bc4a67.

tests_tf/test_attacks.py
fartashf: Added comments in 85dea1d.

fartashf:
Thanks Nicolas for the comments. I addressed them above.

fartashf:
Tests are failing on the mnist_tutorial_tf which is untouched by this PR. Looks like the atol_fac for Tensorflow=1.1.0 is too tight. Or maybe the tutorial needs better seeding.

npapernot:
Let's loosen the test then. The tutorials do need better seeding but unfortunately all of our efforts at making the results fully deterministic have been unsuccessful so far.

fartashf:
Thanks, @npapernot. Tests have passed.

npapernot:
Thanks Fartash!

Issue #253, opened 2017-08-30
Better seeding / test that seeding works


goodfeli:
Our tutorial tests still show significant differences when run multiple times. See the comments on this PR for some examples: https://github.com/tensorflow/cleverhans/pull/249

We should figure out which part of the tutorials is not seeded correctly, seed it, and add a test that running the tutorial twice gives nearly the same result.

goodfeli:
I did some work toward this but didn't complete it, in this PR: https://github.com/tensorflow/cleverhans/pull/255

The PR adds the necessary tests, but with loose tolerance. This will hopefully make sure we at least don't add more non-deterministic ops, but I haven't found the cause of the current non-determinism yet.

goodfeli:
@npapernot  and I discussed this today during the hackathon. Our current state of knowledge is: TensorFlow isn't deterministic so we won't be able to make CleverHans completely deterministic. There are still some things we could do to make CleverHans more deterministic, like set eigen to be single threaded.

catherio:
@goodfeli Can you clarify what work still has to be done now that #308 has been merged?

npapernot:
David would be the person to talk to about this :) My understanding is that there is little chance we will get this solved without making changes to tensorflow

goodfeli:
Which David?

I think the main thing to do is to run the determinism tests with Eigen configured to be single-threaded. I'm not sure whether there are more sources of non-determinism in TensorFlow besides that.

CleverHans will probably always be non-deterministic during real use because the underlying linear algebra primitives are non-deterministic, but we can write unit tests that check that CleverHans isn't adding new non-determinism by failing to seed RNGs.

npapernot:
This would be @david-berthelot :)

David can confirm, but that's exactly what he found out too I believe

david-berthelot:
I have narrowed the problem down to a trivially small test case. The problem was consecutive calls to tf.Session() seemed to have their internal seed (it's been 2 weeks I looked at it, didn't have time since then so I'm not entirely clear).

PR #308, opened 2017-11-03
Tighten mnist tutorial tests

fartashf:
Given the atol_fac, the tests had become too loose. the new values allow 1.5% change in the train/test clean/adv accuracies except for adversarial training accuracy that is allowed to change 6%.

### DIFF
--- a/tests_tf/test_mnist_tutorial_tf.py
+++ b/tests_tf/test_mnist_tutorial_tf.py
@@ -43,10 +43,10 @@ class TestMNISTTutorialTF(CleverHansTest):
                          atol=atol_fac * 5e-3)
         self.assertClose(report.train_adv_train_clean_eval,
                          report_2.train_adv_train_clean_eval,
-                         atol=atol_fac * 2e-2)
+                         atol=atol_fac * 5e-3)
         self.assertClose(report.train_adv_train_adv_eval,
                          report_2.train_adv_train_adv_eval,
-                         atol=atol_fac * 2e-1)
+                         atol=atol_fac * 2e-2)
 
 
 if __name__ == '__main__':


fartashf:
@npapernot, thanks for the merge. Is there a reason for having different atol_fac for different TF versions? Is there a difference in the underlying seeding mechanism? The change was introduced in 7ab2a7fa4a07f3a52a49e5197b347ba78c7cb8ed. I have a commit that fails the test for TF<1.1.0.

npapernot:
The reason is that some tests were previously failing for one of the two TF
versions we're testing. These tests are not ideal to begin with, but before
we solve the reproducibility issue #253, it's going to remain a problem.

npapernot:
Ah that is possible, I will try to spend some time on this Friday afternoon. This has been a problem for a long time and at this point we cannot rely on these tests for what it's worth.

npapernot:
In the meanwhile, feel free to adjust the atol_fac if needed, I trust your judgment.

goodfeli:
This causes the tests to fail on CPU on tf 1.0.1 on my MacBook Pro now.

  File "/Users/goodfellow/cleverhans/tests_tf/test_mnist_tutorial_tf.py", line 46, in test_mnist_tutorial_tf
    atol=atol_fac * 5e-3)
  File "/Users/goodfellow/cleverhans/cleverhans/devtools/checks.py", line 29, in assertClose
    assert np.allclose(x, y, *args, **kwargs), (x, y)
AssertionError: (0.94540000000000002, 0.96060000000000001)

fartashf:
As a temporary fix, I made a new PR #310 to loosen tests for TF<1.1.0.

PR #318, opened 2017-11-25
Alternative implementation for JSMA

zHiOoO:
I know there is currently another open pull request for the JSMA, but I just want to share the work I have done and hopefully it can help too. 

My motivation for the change is to improve the efficiency to generate adversarial examples for JSMA. With the change, the time needed for generation should be greatly reduced. I haven't done a precise measurement, but it roughly takes about 15 sec to generate 500 adversarial samples with GPU. 

The change tries to keeps the same interface of the original implementation, so that no modification is needed for tutorial.

I have run the tests in the test_tf and the tutorial. Some of the tests in test_tf is failing, but I think it fails even without my change. I also tried to fix the "test_generate_np_targeted_gives_adversarial_example" test, but I can't. Maybe someone can comment on what the test is trying to do. 

Any comments and suggestions are appreciated.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -486,37 +486,33 @@ class SaliencyMapMethod(Attack):
         :param y_target: (optional) Target tensor if the attack is targeted
         """
         import tensorflow as tf
-        from .attacks_tf import jacobian_graph, jsma_batch
+        from .attacks_tf import jsma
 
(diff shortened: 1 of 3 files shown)


zHiOoO:
I signed it!

zHiOoO:
I accidentally submitted the change with a wrong email. Closing this one and will create another one.

PR #319, opened 2017-11-25
Alternative implementation for JSMA

zHiOoO:
I know there is currently another open pull request for the JSMA, but I just want to share the work I have done and hopefully it can help too.

My motivation for the change is to improve the efficiency to generate adversarial examples for JSMA. With the change, the time needed for generation should be greatly reduced. I haven't done a precise measurement, but it roughly takes about 15 sec to generate 500 adversarial samples with GPU.

The change tries to keeps the same interface of the original implementation, so that no modification is needed for tutorial.

I have run the tests in the test_tf and the tutorial. I think there is something wrong with the "test_generate_np_targeted_gives_adversarial_example" test. I tried to fix it, but I can't find a good way to do so. Maybe someone can comment on what the test is trying to do.

Any comments and suggestions are appreciated.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -472,7 +472,7 @@ class SaliencyMapMethod(Attack):
         import tensorflow as tf
         self.feedable_kwargs = {'y_target': tf.float32}
         self.structural_kwargs = ['theta', 'gamma',
-                                  'clip_max', 'clip_min']
+                                  'clip_max', 'clip_min', 'symbolic_impl']
 
     def generate(self, x, **kwargs):
         """
@@ -486,36 +486,68 @@ class SaliencyMapMethod(Attack):
         :param y_target: (optional) Target tensor if the attack is targeted
         """
         import tensorflow as tf
-        from .attacks_tf import jsma_symbolic
 
         # Parse and save attack-specific parameters
         assert self.parse_params(**kwargs)
 
-        # Create random targets if y_target not provided
-        if self.y_target is None:
-            from random import randint
+        if self.symbolic_impl:
+            from .attacks_tf import jsma_symbolic
 
-            def random_targets(gt):
-                result = gt.copy()
-                nb_s = gt.shape[0]
-                nb_classes = gt.shape[1]
+            # Create random targets if y_target not provided
+            if self.y_target is None:
+                from random import randint
 
-                for i in xrange(nb_s):
-                    result[i, :] = np.roll(result[i, :],
-                                           randint(1, nb_classes-1))
+                def random_targets(gt):
+                    result = gt.copy()
+                    nb_s = gt.shape[0]
+                    nb_classes = gt.shape[1]
 
-                return result
+                    for i in xrange(nb_s):
+                        result[i, :] = np.roll(result[i, :],
+                                               randint(1, nb_classes-1))
 
-            labels, nb_classes = self.get_or_guess_labels(x, kwargs)
-            self.y_target = tf.py_func(random_targets, [labels], tf.float32)
-            self.y_target.set_shape([None, nb_classes])
+                    return result
 
-        return jsma_symbolic(x, model=self.model, y_target=self.y_target,
-                             theta=self.theta, gamma=self.gamma,
-                             clip_min=self.clip_min, clip_max=self.clip_max)
+                labels, nb_classes = self.get_or_guess_labels(x, kwargs)
+                self.y_target = tf.py_func(random_targets, [labels], tf.float32)
+                self.y_target.set_shape([None, nb_classes])
+
+            x_adv = jsma_symbolic(x, model=self.model, y_target=self.y_target,
+                                  theta=self.theta, gamma=self.gamma,
+                                  clip_min=self.clip_min, clip_max=self.clip_max)
+        else:
+            from .attacks_tf import jacobian_graph, jsma_batch
+
+            # Define Jacobian graph wrt to this input placeholder
+            preds = self.model.get_probs(x)
+            nb_classes = preds.get_shape().as_list()[-1]
+            grads = jacobian_graph(preds, x, nb_classes)
+
+            # Define appropriate graph (targeted / random target labels)
+            if self.y_target is not None:
+                def jsma_wrap(x_val, y_target):
+                    return jsma_batch(self.sess, x, preds, grads, x_val,
+                                      self.theta, self.gamma, self.clip_min,
+                                      self.clip_max, nb_classes,
+                                      y_target=y_target)
+
+                # Attack is targeted, target placeholder will need to be fed
+                x_adv = tf.py_func(jsma_wrap, [x, self.y_target], tf.float32)
+            else:
+                def jsma_wrap(x_val):
+                    return jsma_batch(self.sess, x, preds, grads, x_val,
+                                      self.theta, self.gamma, self.clip_min,
+                                      self.clip_max, nb_classes,
+                                      y_target=None)
+
+                # Attack is untargeted, target values will be chosen at random
+                x_adv = tf.py_func(jsma_wrap, [x], tf.float32)
+
+        return x_adv
 
     def parse_params(self, theta=1., gamma=1., nb_classes=None,
-                     clip_min=0., clip_max=1., y_target=None, **kwargs):
+                     clip_min=0., clip_max=1., y_target=None, 
+                     symbolic_impl=True, **kwargs):
         """
         Take in a dictionary of parameters and applies attack-specific checks
         before saving them as attributes.
@@ -538,6 +570,7 @@ class SaliencyMapMethod(Attack):
         self.clip_min = clip_min
         self.clip_max = clip_max
         self.y_target = y_target
+        self.symbolic_impl = symbolic_impl
 
         return True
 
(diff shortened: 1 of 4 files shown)


npapernot:
Thanks for the pull request, in the future it would be best to discuss this on the cleverhans-dev@ mailing list to avoid duplicate PRs. I do appreciate you sharing your work.

Let us see if it makes sense to keep #275 given that this PR has been open. @gongzhitaao are you planning to complete your PR soon, or should we close it and all work towards this current PR being merged (#319)?

cleverhans/attacks_tf.py
```diff
@@ -257,151 +184,138 @@ def jacobian_graph(predictions, x, nb_classes):
...
+
+        # Create a mask to only keep features that match conditions
+        if increase:
+            scores_mask = ((target_sum > 0) & (other_sum < 0))
+        else:
+            scores_mask = ((target_sum < 0) & (other_sum > 0))
 
-        # Update loop variables
-        iteration = iteration + 1
+        # Create a 2D numpy array of scores for each pair of candidate features
```
gongzhitaao: Quick question: suppose we have an N by N image and we need to find the best pair to perturb, according to this code, we need a matrix of N^2 by N^2, am I understanding it right? In this case, suppose we want to perturb an image of size100x100, then we need approx. 2G memory.

gongzhitaao:
I'm OK either way.  I quickly went through the implementation. As far as I could see, for smaller image size, this solution should be faster than mine since the graph is simpler and more vectorized.
@zHiOoO Have you tested 256x256 images from ImageNet?

cleverhans/attacks_tf.py
zHiOoO: Yes, you are right. And it will take more memory if the image has multiple channels.

zHiOoO:
@gongzhitaao No, I haven't tried using images with 256x256. You have a very good point. What I have done is a direct port from the original implementation. It works well for simple dataset like MNIST, but it does has memory issue for datasets with larger image size.

gongzhitaao:
@npapernot I think we could close my PR #275 and work toward this submission.
@zHiOoO Your code has some good idea that I want to integrate into my own implementation.  Thank you for the PR :smiley:

npapernot:
Thanks for taking the time to look at this PR @gongzhitaao , I will close #275 for now. Hope that you are able to build on / contribute to this PR with your implementation ideas.

cleverhans/attacks_tf.py
```diff
@@ -257,151 +184,138 @@ def jacobian_graph(predictions, x, nb_classes):
...
-        current = utils_tf.model_argmax(sess, x, predictions,
-                                        adv_x_original_shape, feed=feed)
+        target_class = tf.reshape(tf.transpose(y_in, perm=[1, 0]),
+                                  shape=[nb_classes, -1, 1])
+        other_classes = tf.cast(tf.not_equal(target_class, 1), tf.float32)
+
+        grads_target = tf.reduce_sum(grads * target_class, axis=0)
+        grads_other = tf.reduce_sum(grads * other_classes, axis=0)
+
+        # Remove the already-used input features from the search space
```
npapernot: Why not just set the values of `grads_target` and `grads_other` to 0 for the features in domain_in? These way they will not match the `scores_mask` conditions below.

cleverhans/attacks_tf.py
```diff
@@ -257,151 +184,138 @@ def jacobian_graph(predictions, x, nb_classes):
...
+        p1_one_hot = tf.one_hot(p1, depth=nb_features)
+        p2_one_hot = tf.one_hot(p2, depth=nb_features)
 
-    # Report success when the adversarial example is misclassified in the
-    # target class
-    if current == target:
-        return np.reshape(adv_x, original_shape), 1, percent_perturbed
-    else:
-        return np.reshape(adv_x, original_shape), 0, percent_perturbed
+        # Check if more modification is needed for each samples
```
npapernot: "for each sample"

cleverhans/attacks_tf.py
```diff
@@ -257,151 +184,138 @@ def jacobian_graph(predictions, x, nb_classes):
...
+        mod_done = tf.equal(tf.reduce_sum(y_in * preds_onehot, axis=1), 0)
```
npapernot: Should this be called `mod_not_done` instead?

cleverhans/attacks.py
```diff
@@ -486,39 +486,35 @@ def generate(self, x, **kwargs):
...
-            wrap = tf.py_func(jsma_wrap, [x], tf.float32)
+                for i in xrange(nb_s):
+                    result[i, :] = np.roll(result[i, :],
+                                           randint(1, nb_classes-1))
 
-        return wrap
+                return result
+
+            labels, nb_classes = self.get_or_guess_labels(x, kwargs)
+            self.y_target = tf.py_func(random_targets, [labels], tf.float32)
```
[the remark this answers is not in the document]
zHiOoO: I would love to avoid using py_func here, but I couldn't think of a good way to do this with pure tf functions.

cleverhans/attacks.py
```diff
@@ -486,39 +486,35 @@ def generate(self, x, **kwargs):
...
+            self.y_target.set_shape([None, nb_classes])
```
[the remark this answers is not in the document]
zHiOoO: This is necessary because the output of py_func has unknown shape.

cleverhans/attacks_tf.py
```diff
@@ -133,79 +133,6 @@ def vatm(model, x, logits, eps, num_iterations=1, xi=1e-6,
         return adv_x
 
 
-def apply_perturbations(i, j, X, increase, theta, clip_min, clip_max):
```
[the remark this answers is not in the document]
zHiOoO: Sure. I can add them back.

cleverhans/attacks_tf.py
```diff
@@ -257,151 +184,138 @@ def jacobian_graph(predictions, x, nb_classes):
     return list_derivatives
 
 
-def jsma(sess, x, predictions, grads, sample, target, theta, gamma, clip_min,
-         clip_max, feed=None):
+def jsma(x, y_target, model, theta, gamma, clip_min, clip_max):
     """
     TensorFlow implementation of the JSMA (see https://arxiv.org/abs/1511.07528
     for details about the algorithm design choices).
-    :param sess: TF session
```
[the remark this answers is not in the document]
zHiOoO: Because the function no longer requires this input.

cleverhans/attacks_tf.py
```diff
@@ -257,151 +184,138 @@ def jacobian_graph(predictions, x, nb_classes):
...
+        # reaches the maximum iterations
+        return tf.logical_and(tf.less(i_in, max_iters), cond_in)
 
-    _logger.debug("Starting JSMA attack up to {} iterations".format(max_iters))
-    # Repeat this main loop until we have achieved misclassification
-    while (current != target and iteration < max_iters and
-           len(search_domain) > 1):
-        # Reshape the adversarial example
-        adv_x_original_shape = np.reshape(adv_x, original_shape)
+    def body(x_in, y_in, domain_in, i_in, cond_in):
```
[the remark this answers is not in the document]
zHiOoO: Will do.

cleverhans/attacks_tf.py
```diff
@@ -257,151 +184,138 @@ def jacobian_graph(predictions, x, nb_classes):
...
-            grads_target, grads_others, search_domain, increase)
-
-        # Apply the perturbation to the two input features selected previously
-        adv_x = apply_perturbations(
-            i, j, adv_x, increase, theta, clip_min, clip_max)
-
-        # Update our current prediction by querying the model
-        current = utils_tf.model_argmax(sess, x, predictions,
-                                        adv_x_original_shape, feed=feed)
+        target_class = tf.reshape(tf.transpose(y_in, perm=[1, 0]),
```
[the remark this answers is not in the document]
zHiOoO: Will do

cleverhans/attacks_tf.py
zHiOoO: The algorithm is picking pairs of pixels/features, so setting one pixel/feature to 0 doesn't prevent it from being picked again later. That's why their values need to be set based on the maximum value from the gradients.

cleverhans/attacks_tf.py
zHiOoO: Will fix

cleverhans/attacks_tf.py
zHiOoO: Good catch. Will fix.

goodfeli:
@npapernot I think it's ready for you to re-review

npapernot:
Thanks, this is almost ready to merge! just two things would be nice to take care of:

* solve the conflict that the branch has w/ the master branch
* change the JSMA tutorial in `cleverhans_tutorials` to use the symbolic implementation by default (this will also make sure that the behavior is consistent with the previous implementation)

Thanks again

zHiOoO:
@npapernot The conflict is resolved. I have also added a new parameter for JSMA to allow switching between two implementations.

npapernot:
Thanks, I'll try a few things on my machine and then merge

PR #320, opened 2017-12-01
support different clipping values for different pixels

sli057:
The original cleverhans assumes the clip_min and clip_max to be the same for all pixels.
However, in some case, for example, preprocessing a image by subtracting the corresponding mean value on each pixel, the clip_min and clip_max are actually different for each pixel.
The modified code supports the mentioned new clip for four attack classes: FastGradientMethod, BasicIterativeMethod, VirtualAdversarialMethod, MadryEtAl; while SaliencyMapMethod, CarliniWagnerL2, DeepFool, ElasticNetMethod are not modified yet.


npapernot:
Thanks for the PR, but I don't see any files checked-in? The only two files that appear on the PR are `.DS_store` files.

sli057:
Hi,
I always fail at test_mnist_tutorial_tf, which also happened in Build # 1029.
Theoretically, I do not change anything related to this test, but it still fails.
I was testing with the original code to see if there is any chance that it is the problem of the original code.
I noticed that the solution to Buid # 1029 fail was to tolerate a larger gap. Do you think it is okay to do that in my case? I test with 8e-2 instead of 2e-2 and it succeeded.

goodfeli:
Thanks. Were you seeing the failures on Travis, your own machine, or both?

sli057:
It is good on my own machine. It failed only on Travis.

goodfeli:
OK, and now I see the tolerance you had to raise was for the determinism tests, not the correctness tests, so that's less likely to be important

sli057:
Great!
Thanks!

goodfeli:
I'm sorry, but I think we actually should not merge this feature.

In order to keep the code cleanly decomposed, implementations of universally applicable Attacks shouldn't need to contain logic related to specific preprocessing steps. This approach wouldn't scale to support all kinds of preprocessing like ZCA, GCN, etc. We're also likely to introduce bugs while trying to maintain attacks that have to take more complicated logic related to the mean preprocessing into account.

I think the right way to use this is just to pass a tensor-valued argument to clip_min and clip_max.

In general, before starting a PR with work on a new feature, it's best to write to cleverhans-dev@googlegroups.com to make sure we agree on the new feature.

PR #322, opened 2017-12-10
Fixes to black-box attack tutorial

### LINKED ISSUE #321: Error in jacobian_augmentation in attacks_tf
I think line 456 shouldn't have the "2*".  It looks like it is writing over every other x value in a stacked file of 2 copies of x values.  That means it is keeping two copies of every other x.  Also, the success rate on adversarial examples goes up when you increase the epochs, which is counter-intuitive. 

I made this change and the code is running the way it should.  The oracle has a harder time correctly identifying adversarial examples when you increase the number of augmentations to the substitute training set.

npapernot:
Closes #321

### DIFF
--- a/cleverhans_tutorials/mnist_blackbox.py
+++ b/cleverhans_tutorials/mnist_blackbox.py
@@ -92,9 +92,7 @@ def substitute_model(img_rows=28, img_cols=28, nb_classes=10):
     input_shape = (None, img_rows, img_cols, 1)
 
     # Define a fully connected model (it's different than the black-box)
-    layers = [Conv2D(64, (8, 8), (2, 2), "SAME"),
-              ReLU(),
-              Flatten(),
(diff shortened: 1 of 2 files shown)


cleverhans_tutorials/mnist_blackbox.py
```diff
@@ -91,7 +92,9 @@ def substitute_model(img_rows=28, img_cols=28, nb_classes=10):
     input_shape = (None, img_rows, img_cols, 1)
 
     # Define a fully connected model (it's different than the black-box)
-    layers = [Flatten(),
+    layers = [Conv2D(64, (8, 8), (2, 2), "SAME"),
```
goodfeli: The comment above says this model is meant to be fully connected to demonstrate transfer between dissimilar substitute and target, so we shouldn't add a Conv2D layer to this one

PR #325, opened 2017-12-12
Repeatability

david-berthelot:
This addresses the issue for better seeding  #253 .
The trick was two create two graphs since in a single graph two random variable nodes must generate different values.
There are still some repeatability issues but only GPU and of a lesser magnitude. CPU appears totally deterministic now.


npapernot:
Thanks David, LGTM

npapernot:
Did you get a chance to test the `atol` values on both GPU and CPU thoroughly after we spoke?

goodfeli:
Thank you!

PR #332, opened 2017-12-22
Adding Momentum Iterative Attacks

dongyp13:
This branch implement the momentum iterative attack method, which was proposed to improve the transferability of adversarial examples and won the first places in NIPS 2017 adversarial competitions. The current implementation includes L1, L2 and L infinity attacks.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -524,24 +524,30 @@ class MomentumIterativeMethod(Attack):
 
             # Normalize current gradient and add it to the accumulated gradient
             red_ind = list(xrange(1, len(grad.get_shape())))
-            grad = grad / tf.reduce_mean(tf.abs(grad),
-                                         reduction_indices=red_ind,
-                                         keep_dims=True)
+            avoid_zero_div = 1e-12
+            grad = grad / tf.maximum(avoid_zero_div, 
+                                     tf.reduce_mean(tf.abs(grad),
+                                                    red_ind,
+                                                    keep_dims=True))
             momentum = self.decay_factor * momentum + grad
 
             if self.ord == np.inf:
                 normalized_grad = tf.sign(momentum)
-            elif self.ord in [1, 2]:
-                reduc_ind = list(xrange(1, len(momentum.get_shape())))
-                if self.ord == 1:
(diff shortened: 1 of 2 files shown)


npapernot:
Thanks for the PR! Can you add tests for the attack? You can find an example here: https://github.com/tensorflow/cleverhans/blob/master/tests_tf/test_attacks.py#L115

dongyp13:
I have added the tests.

cleverhans/attacks.py
```diff
@@ -451,6 +451,154 @@ def parse_params(self, eps=0.3, eps_iter=0.05, nb_iter=10, y=None,
...
+                                         reduction_indices=red_ind,
+                                         keep_dims=True)
+            momentum = self.decay_factor * momentum + grad
+
+            if self.ord == np.inf:
+                normalized_grad = tf.sign(momentum)
+            elif self.ord in [1, 2]:
+                reduc_ind = list(xrange(1, len(momentum.get_shape())))
+                if self.ord == 1:
+                    norm = tf.reduce_sum(tf.abs(momentum),
```
npapernot: It would be nice to apply the same logic that @goodfeli applied to the clipping function in order to avoid numerical instabilities in the norm: see https://github.com/tensorflow/cleverhans/blob/master/cleverhans/utils_tf.py#L417-L431

npapernot: Alternatively, you may be able to reuse the same function `clip_eta` to improve maintainability

npapernot:
Thanks for adding the test, I just realized that it might make sense to take into account a recent PR by @goodfeli (#328) to prevent numerical instabilities. If you can take a look and make the edits, the rest is LGTM and ready to merge.

cleverhans/attacks.py
dongyp13: Ok, I'll address it. It's actually the same code as fgm https://github.com/tensorflow/cleverhans/blob/master/cleverhans/attacks_tf.py#L65-L87.
I reuse the 'clip_eta' after adding the perturbation.

dongyp13:
I've fixed the numerical instability issue in the recent commit.

npapernot:
Thank you!

Issue #339, opened 2018-01-19
Build is currently failing with TENSORFLOW_V=1.4.1


catherio:
The build for https://github.com/tensorflow/cleverhans/pull/336 is currently failing with TENSORFLOW_V=1.4.1 but not TENSORFLOW_V=1.1.0. It looks like we need to update a dependency.

catherio:
closed as of #343

Issue #340, opened 2018-01-19
Error in test_get_logits_over_interval (test_utils.TestUtils)


catherio:
The latest builds (with tf 1.1.0) are failing with the following error:
```
======================================================================
ERROR: test_get_logits_over_interval (test_utils.TestUtils)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/tensorflow/cleverhans/tests_tf/test_utils.py", line 97, in test_get_logits_over_interval
    num_points=num_points)
  File "/home/travis/build/tensorflow/cleverhans/cleverhans/utils.py", line 273, in get_logits_over_interval
    logits = model.get_logits(lin_batch)
  File "/home/travis/build/tensorflow/cleverhans/cleverhans/utils_keras.py", line 151, in get_logits
    logits_name = self._get_logits_name()
  File "/home/travis/build/tensorflow/cleverhans/cleverhans/utils_keras.py", line 141, in _get_logits_name
    node = softmax_layer.inbound_nodes[0]
AttributeError: 'Activation' object has no attribute 'inbound_nodes'
----------------------------------------------------------------------
```

See https://travis-ci.org/tensorflow/cleverhans/jobs/330906885 (build for https://github.com/tensorflow/cleverhans/pull/336)

goodfeli:
Did the version of Keras change recently?
@royaurko any comment?

catherio:
`inbound_nodes` got renamed to `_inbound_nodes` as of https://github.com/keras-team/keras/commit/958239c621a6be037c5f8b30be9270310735f725#diff-fe3ca734e8421a17729d0b0ea4c800ba. What's the right way to proceed?

catherio:
Proposal: replace that line (which is the only use of _inbound_nodes) with a hasattr check and a deprecation warning, to avoid breaking folks who are still using older Keras

catherio:
closed as of #343

PR #342, opened 2018-01-22
Add libgcc to conda install to address issue 339

catherio:
This might fix the ligstdc++.so.6 issue we see #339. I can't repro locally, so just gonna see if the build turns up green.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,11 +43,12 @@ install:
   # debug libgcc
   - python -c "import scipy; print(scipy.__file__)"
   - python -c "import scipy; print(scipy.__version__)"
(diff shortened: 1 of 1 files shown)


npapernot:
Thanks Catherine! Feel free to ping me directly if you need someone to look at it / LGTM once the tests turn green.

catherio:
@npapernot This would be faster if we had https://docs.travis-ci.com/user/running-build-in-debug-mode/ enabled. Is there a reason not to?

npapernot:
I think there is no reason for us to not use this debug mode because the potential security threats are not really a risk for our project. Let me email the Travis support to ask them to enable this feature for future needs. Note that according to the documentation, debug mode is limited to those with push access, which includes you.

catherio:
The build won't be happy unless I fix both issue #339 and issue #340 at once, so I'm closing this and submitting one PR for both.

PR #344, opened 2018-01-23
submodules considered harmful

goodfeli:


goodfeli:
@catherio can you review?

npapernot:
Thanks @catherio for linking to the medium article, very useful read. Submodules don't seem to be like a good idea for us, and an example importing the target repository would be a better option.

catherio:
LGTM!

Issue #361, opened 2018-02-11
cannot initialize an attack class with callable function


michaelzhou0723:
Take `FastGradientMethod` as an example, its `__init__` method checks and wraps any callable function with `CallableModelWrapper`. However it's done after `super(FastGradientMethod, self).__init__(model, back, sess)`, which raises an exception if the model argument is not an instance of `Model` class.

A simple modification would be exchanging their order of execution in every subclass of Attack.

catherio:
Thanks for the PR!

PR #364, opened 2018-02-13
fix the wrapping of callable function in attack initialization

michaelzhou0723:
**Issue:** [https://github.com/tensorflow/cleverhans/issues/361](https://github.com/tensorflow/cleverhans/issues/361)

**Proposed Fix:** If the model argument is not an instance of `model.Model`, first wrap it with `model.CallableModelWrapper` and then pass initialization to base class


goodfeli:
LGTM assuming tests pass

PR #369, opened 2018-03-02
Add SPSA gradient-free attack method

juesato:
The diff is pretty large, partly due to verbose tests, and partly because I originally wrote this using methods from another library. I've moved these helper methods into `attacks_tf`, the two main ones are `pgd_attack`, which uses a `while_loop` to optimize a loss function within the TF graph, and `UnrolledOptimizer`, which is analogous to standard Optimizers, but optimizes Tensors rather than Variables.


juesato:
This is ready to review.

I believe the TF 1.1 tests are failing at head (since even this dummy PR fails https://github.com/tensorflow/cleverhans/pull/370).

juesato:
Tests are now passing

PR #375, opened 2018-03-21
Fixing bug in get_logits method of KerasModelWrapper

### LINKED ISSUE #307: Issues with Keras trained model and the Carlin-Wagner L2 attack
I have not been able to successfully generate adversarial examples on the MNIST dataset when the model is trained in Keras.
Quite possibly I am doing something rather silly here.

The jupyter notebook highlighting this issue is here:

https://github.com/vinayprabhu/Lyapunov_defense/blob/master/MNIST_CW_Fail_Keras.ipynb

Also, cw_params = {'binary_search_steps': 15,
             'y_target': adv_ys,
             'max_iterations': 50000,
             'learning_rate': 1,
             'batch_size': source_samples * nb_classes if
             targeted else source_samples,
             'initial_const': 0.1,
            'verbose':False}

nishdesai:
Fix #307 

Keras supports two methods for adding activations to layers. The first is to add an `Activation` layer to the definition of a model:

    model_1 = Sequential()
    model_1.add(Dense(10, input_shape=(10,), name='not_logits'))
    model_1.add(Dense(10, name='logits'))
    model_1.add(Activation('softmax', name='softmax'))

The second method is to include an activation in the layer directly using the `activation` keyword arg:

    model_2 = Sequential()
    model_2.add(Dense(10, input_shape=(10,), name='not_logits'))
    model_2.add(Dense(10, name='logits_and_softmax', activation='softmax'))

The `get_logits` method of the `KerasModelWrapper` finds the logits layer by first finding the sotfmax layer and then returning the node that is the input to the softmax layer. This works with method 1 but breaks with method 2.

    >>> wrapped_1 = KerasModelWrapper(model_1)
    >>> wrapped_2 = KerasModelWrapper(model_2)
    >>> print("Model 1 Logit Layer: {}\n".format(wrapped_1.get_logits(model_1.input).name))
    Model 1 Logit Layer: model_1/logits/BiasAdd:0

    >>> print("Model 2 Logit Layer: {}\n".format(wrapped_2.get_logits(model_2.input).name))
    Model 2 Logit Layer: model_2/not_logits/BiasAdd:0

**Fix:** If the softmax layer is not an instance of `Activation`, return the input to the softmax `_op` of the softmax layer instead of the output of the previous layer.

### DIFF
--- a/cleverhans/utils_keras.py
+++ b/cleverhans/utils_keras.py
@@ -167,11 +167,10 @@ class KerasModelWrapper(Model):
         # Need to deal with the case where softmax is part of the
         # logits layer
         if logits_name == self._get_softmax_name():
-            softmax_logit_layer = self.model.get_layer(logits_name)
-            softmax_out = softmax_logit_layer.output
+            softmax_logit_layer = self.get_layer(x, logits_name)
 
             # The final op is the softmax. Return its input
-            logits_layer = softmax_out._op.inputs[0]
+            logits_layer = softmax_logit_layer._op.inputs[0]
 
         return logits_layer
 
--- a/cleverhans/utils_keras.py
+++ b/cleverhans/utils_keras.py
@@ -139,6 +139,10 @@ class KerasModelWrapper(Model):
         softmax_name = self._get_softmax_name()
         softmax_layer = self.model.get_layer(softmax_name)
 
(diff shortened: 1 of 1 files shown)


nishdesai:
CLA should be signed now.

PR #379, opened 2018-04-09
Add optimizer and decision rule options to EAD

ysharma1126:


npapernot:
Thanks for the PR, would you be able to add deprecation warning given that you are changing the API? Unless the new parameters have default values that correspond exactly to the previous behavior.

ysharma1126:
The new parameters have default values that correspond exactly to the
previous behavior.

npapernot:
Thanks for confirming. Given that you're an author, I trust that it implements correctly the corresponding paper.

Issue #376, opened 2018-03-25
Tiny bug in adversarial patch notebook


shreyashankar:
In `AdversarialPatch.ipynb`, the function `show_patch` takes one argument, but the API shows it as `show_patch()` and the function `load_patch` calls `show_patch()` incorrectly. Just FYI :-)

goodfeli:
@nottombrown can you take a look?

nottombrown:
Solved by #383 - Thanks @shreyashankar !

PR #383, opened 2018-04-12
Update AdversarialPatch.ipynb with bugfix

nottombrown:
Solves issue #376. Thanks for the report @shreyashankar!


goodfeli:
I'll send a PR to fix the build

PR #384, opened 2018-04-13
fix strings that weren't escaped correctly

goodfeli:


goodfeli:
I'm not sure how this file got into the repo in the first place. I haven't investigated it, but my guess is that older versions of `pycodestyle` weren't configured to report this kind of error by default, and an update to `pycodestyle` made our travis build start detecting this error.

PR #393, opened 2018-04-20
Update ex_cifar10_tf from fgsm to FastGradientMethod

lwneal:
Bug: Currently the `ex_cifar10_tf.py` example is broken due to a change in the fgsm API.

To reproduce: install the current version of cleverhans with `python setup.py install` and run the example.

Expected:
```
$ python ex_cifar10_tf.py
 Test accuracy on legitimate test examples: 0.4957
 Test accuracy on legitimate test examples: 0.5728
...
```

Actual:
```
$ python ex_cifar10_tf.py
Using TensorFlow backend.
  File "ex_cifar10_tf.py", line 15, in <module>
    from cleverhans.attacks import fgsm
ImportError: cannot import name 'fgsm'
```

This PR updates `ex_cifar10_tf.py` to use the new FastGradientMethod class.

### DIFF
--- a/examples/ex_cifar10_tf.py
+++ b/examples/ex_cifar10_tf.py
@@ -12,7 +12,7 @@ import tensorflow as tf
 from tensorflow.python.platform import app
 from tensorflow.python.platform import flags
 
-from cleverhans.attacks import fgsm
+from cleverhans.attacks import FastGradientMethod
 from cleverhans.utils_keras import cnn_model
 from cleverhans.utils_tf import model_train, model_eval, batch_eval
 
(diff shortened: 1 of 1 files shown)


examples/ex_cifar10_tf.py
```diff
@@ -131,7 +132,7 @@ def evaluate():
     # Redefine TF model graph
     model_2 = cnn_model(img_rows=32, img_cols=32, channels=3)
     predictions_2 = model_2(x)
-    adv_x_2 = fgsm(x, predictions_2, eps=0.3)
+    adv_x_2 = fgsm.generate(x, eps=0.3)
```
npapernot: You would have to redefine fgsm on the `model_2`

npapernot:
Thanks for the PR, there is a minor tweak missing but should be good once that is fixed

PR #394, opened 2018-04-23
Fixing Visualization to display color images

aashish-kumar:
Visualization for color/multi-channel images was not handled.


aashish-kumar:
I signed it!

PR #405, opened 2018-05-11
Add Facenet Adversarial Faces example

BrunoLopezGarcia:


examples/facenet_adversarial_faces/facenet_fgsm.py
```diff
@@ -0,0 +1,131 @@
...
+        fgsm = FastGradientMethod(model)
+        fgsm_params = {'eps': 0.01,
+                       'clip_min': 0.,
+                       'clip_max': 1.}
+        adv_x = fgsm.generate(model.face_input, **fgsm_params)
+
+        # Run FGSM
+        adv = faces1
+        steps = 1
+        for i in range(steps):
```
npapernot: It looks like you are trying to do something like the BasicIterativeMethod class (modulo the clipping), you might find that it simplifies your code.

npapernot: (This is a suggestion)

BrunoLopezGarcia: I used FGSM + loops instead of the BasicIterativeMethod class because I thought it was easier to understand for the article that I wrote.

Regarding the clipping, it wasn't necessary as I was using epsilon as epsilon divided by the number of steps. Anyways I changed the code a little bit in order to use alpha instead of epsilon.

PR #407, opened 2018-05-18
Update authors section to use `and` instead of commas

chrisfosterelli:
When referencing cleverhans using your recommended bibtex, latex warns that there are "too many commas". This is because commas are supposed to be used to [differentiate ordering of names, rather than separate authors](https://tex.stackexchange.com/a/343164). This pull request updates it to separate authors using `and`.


goodfeli:
LGTM, let's merge after the tests pass

Issue #397, opened 2018-05-02
tf 1.8 support


goodfeli:
We currently get several deprecation warnings when running tensorflow 1.8. We should update to the point that we get no tf deprecation warnings when running the tutorials and tests. We should also add tf 1.8 to the travis bot.

npapernot:
Closing this given that `mnist_tutorial_tf.py` does not raise any deprecation warnings anymore. If anyone notices another warning, feel free to open an issue.

PR #410, opened 2018-05-29
switch mnist dataset to datapythonista/mnist

npapernot:
A recent change to the TF API creates a deprecation warning because CleverHans uses the MNIST dataset loader that comes with `tf.learn`.

This implements the suggested switch to `sklearn` as the data loader.

### DIFF
--- a/cleverhans/utils_mnist.py
+++ b/cleverhans/utils_mnist.py
@@ -4,8 +4,6 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import numpy as np
-import sys
-import warnings
 
 from . import utils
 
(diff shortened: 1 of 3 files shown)


npapernot:
Partially addresses #397

nottombrown:
Added the missing `scikit-learn` dependency.

nottombrown:
I give permission to include my commits in this PR

nottombrown:
Found an issue here. The sklearn data is sorted by categories, so our train and test set are not IID. 

Two options:
1. Shuffle the mnist data ourselves
2. Use MNIST dataset from `torchvision.datasets` (https://pytorch.org/docs/stable/torchvision/datasets.html#mnist)

I would gravitate towards [2] because torchvision seems lighter-weight than sklearn, and we don't have shuffle things ourselves.

nottombrown:
Heavyweight pytorch dependency is causing issues, so I switched to a new mnist source.

cleverhans/utils_mnist.py
```diff
@@ -27,18 +25,27 @@ def data_mnist(datadir='/tmp/', train_start=0, train_end=60000, test_start=0,
...
+
+    X_train = mnist.train_images() / 255.
+    Y_train = mnist.train_labels()
+    X_test = mnist.test_images() / 255.
+    Y_test = mnist.test_labels()
+
+    X_train = np.expand_dims(X_train, -1)
+    X_test = np.expand_dims(X_test, -1)
+
+    print(X_train.max())
```
npapernot: We can probably drop these two statements to save terminal real estate

npapernot:
Does the new loader return images in a different random order every run?

nottombrown:
Removed print statements. The order of the loader is deterministic. It uses the ordering from http://yann.lecun.com/exdb/mnist/

You can see the code here:
https://github.com/datapythonista/mnist/blob/master/mnist/__init__.py

npapernot:
Thanks! LGTM, feel free to merge when tests pass

PR #412, opened 2018-06-01
Add CI jobs for TensorFlow 1.8

redshiftzero:
Progress towards #397

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,15 +11,7 @@ matrix:
         - python: 2.7
           env:
             - KERAS_BACKEND=tensorflow
-            - TENSORFLOW_V=1.1.0
-        - python: 2.7
-          env:
-            - KERAS_BACKEND=tensorflow
             - TENSORFLOW_V=1.8.0
-        - python: 3.5
-          env:
-            - KERAS_BACKEND=tensorflow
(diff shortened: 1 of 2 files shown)


npapernot:
Thanks for the PR! It might make sense to deprecate TF 1.1 at the same time since it was released over a year ago. That could also help to prevent increasing the burden on Travis when processing pull requests. 

Any thoughts @goodfeli ?

goodfeli:
I propose:
- Change the tests to do tensorflow 1.4 and 1.8
- Change README.md. Currently it says:

    Although CleverHans is likely to work on many other machine configurations, we currently test it with 
    Python {2.7, 3.5} and TensorFlow {1.0, 1.1} on Ubuntu 14.04.5 LTS (Trusty Tahr).

I propose changing it to:

    Although CleverHans is likely to work on many other machine configurations, we currently test it with 
    Python {2.7, 3.5} and TensorFlow {1.4, 1.8} on Ubuntu 14.04.5 LTS (Trusty Tahr). Support for TensorFlow 1.3 and earlier is deprecated. After 2018-11-1 we will not fix bugs reported for these versions and we will eliminate wrapper code needed for backwards compatibility with these versions.

PR #420, opened 2018-06-13
Improve utils_tf.model_train 's functionality and solve several tensorflow version issues

m-Just:
1. `model_train` now accept training data iterator as an argument, supporting some real-time data augmentation schemes such as ImageDataGenerator in Keras.
2. `model_train` can now train with weight decay by simply passing the scale of decay as an argument.
3. `model_train` now additionally log the current learning rate (in case learning rate decay is applied).
4. Fix several tensorflow warnings about deprecated usage.


m-Just:
CLA signed!

PR #423, opened 2018-06-14
Add defense sub-framework.

david-berthelot:
The Loss is the base for defining defense, it allows to manipulate inputs and labels.
Changes:

- New Loss class to create defenses.
- Implemented cross-entropy with smoothing, mixup and feature pairing.
- The Model class has a few changes: removed the get_layer_names API since it wasn't used, and taking explicitly the number of classes (None if non-applicable).
- Fixed some broken examples to be compatible with latest TF (1.8)


npapernot:
Thanks for the PR David, I will take a look asap

david-berthelot:
Note that the Model class is not backward compatible, so may be we should do a Version change. Let me know what you think.

cleverhans/loss.py
```diff
@@ -0,0 +1,50 @@
...
+from .model import Model
+
+
+class Loss(object):
+    """
+    An abstract interface for loss wrappers that allows flexible control of
+    real examples, adversarial examples and labels..
+    """
+
+    def __init__(self, model, hparams, attack=None):
```
npapernot: It might be worth thinking about having a similar style to the interface we have for the Model and Attack class when it comes to the optional parameters (hparams here)? What do you think?

npapernot:
@AlexeyKurakin could you take a look as well? I remember we had long discussions about this last summer with you

cleverhans/loss.py
david-berthelot: Yes possibly, in my case I refrained from modifying existing APIs. But for future PRs that could help to have coherent style.

cleverhans/defenses.py
```diff
@@ -0,0 +1,74 @@
+from .loss import Loss
+from .model import Model
+import tensorflow as tf
+
+
+class LossXEntropy(Loss):
```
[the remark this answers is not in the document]
david-berthelot: Done, indeed it's clearer, sometimes I use abbreviations without even noticing. Thanks.

cleverhans/attacks.py
```diff
@@ -1527,7 +1530,7 @@ def generate(self, x, g, **kwargs):
         # Parse and save attack-specific parameters
         assert self.parse_params(**kwargs)
 
-        g_feat = self.model.get_layer(g, self.layer)
```
npapernot: Is there a particular reason you removed the `get_layer` method from the `Model` class? I would rather keep it to maintain backward compatibility unless it is incompatible with the new code you added.

cleverhans/defenses.py
```diff
@@ -0,0 +1,76 @@
+from .loss import Loss
+from .model import Model
+import tensorflow as tf
+
+
+class LossCrossEntropy(Loss):
```
npapernot: Is there a particular reason for defining the three losses `LossCrossEntropy`, `LossMixUp` and `LossFeaturePairing` in `defenses.py` rather than `loss.py`?

cleverhans/loss.py
npapernot: That's fine, we can clean up and make the entire API more homogeneous in a later PR if needed

cleverhans/loss.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def fprop(self, x, y):
+        """Forward propagate the loss.
+        :param x: tensor, a batch of inputs.
+        :param y: tensor, a batch of outputs (1-hot labels typically).
+        """
+        raise NotImplementedError
+
+
+def attack_softmax_cross_entropy(y, probs, mean=True):
```
npapernot: Why the name `attack_softmax_cross_entropy` and not simply `softmax_cross_entropy`?

cleverhans/model.py
```diff
@@ -11,9 +11,18 @@ class Model(object):
     model when a specific package does not directly expose them.
     """
     __metaclass__ = ABCMeta
+    O_LOGITS, O_PROBS, O_FEATURES = 'logits probs features'.split()
 
-    def __init__(self):
-        pass
+    def __init__(self, scope, nb_classes, hparams):
```
npapernot: This breaks the prior API but does not seem to be necessary for the existing code (unless I missed something). If that's true, could you set default values to be None and add a deprecation warning?

cleverhans/model.py
```diff
@@ -11,9 +11,18 @@ class Model(object):
     model when a specific package does not directly expose them.
     """
     __metaclass__ = ABCMeta
+    O_LOGITS, O_PROBS, O_FEATURES = 'logits probs features'.split()
```
npapernot: Could make these variables modifiable through some arguments in the `__init__` method? Otherwise, what is the purpose of adding these variables?

cleverhans/model.py
```diff
@@ -23,60 +32,32 @@ def __call__(self, *args, **kwargs):
         """
         return self.get_probs(*args, **kwargs)
 
-    def get_layer(self, x, layer):
```
npapernot: See comment above on why is this deleted?

cleverhans/model.py
```diff
@@ -23,60 +32,32 @@ def __call__(self, *args, **kwargs):
...
-    def get_layer_names(self):
-        """
-        :return: a list of names for the layers that can be exposed by this
-        model abstraction.
-        """
-
-        if hasattr(self, 'layer_names'):
-            return self.layer_names
+        d = self.fprop(x, **kwargs)
+        if self.O_PROBS in d:
```
npapernot: I think this behavior is potentially misleading. Is there a particular reason needed to have this return the logits if probs are not found? I would rather have it fail if probs were not found because there is a get_logits method to return the logits

cleverhans/model.py
```diff
@@ -87,7 +68,7 @@ def get_params(self):
         Provides access to the model's parameters.
         :return: A list of all Variables defining the model parameters.
         """
-        raise NotImplementedError()
+        return tf.global_variables(self.scope)
```
npapernot: Could you comment on why you changed this line?

cleverhans/model.py
```diff
@@ -105,11 +86,8 @@ def __init__(self, callable_fn, output_layer):
         self.output_layer = output_layer
         self.callable_fn = callable_fn
 
-    def get_layer_names(self):
```
npapernot: Why remove get_layer_names?

cleverhans/model.py
```diff
@@ -23,60 +32,32 @@ def __call__(self, *args, **kwargs):
...
-        try:
-            return self.get_layer(x, 'probs')
-        except NoSuchLayerError:
-            pass
-        except NotImplementedError:
-            pass
-        import tensorflow as tf
-        return tf.nn.softmax(self.get_logits(x))
-
-    def get_layer_names(self):
```
npapernot: Why did you remove get_layer_names?

cleverhans/utils_tf.py
```diff
@@ -17,30 +17,6 @@
 _logger = create_logger("cleverhans.utils.tf")
 
 
-def model_loss(y, model, mean=True):
```
npapernot: Can you add a deprecation warning and call the new function you defined in loss.py to make this change backward compatible before we remove this function in 6 months?

cleverhans/utils_tf.py
```diff
@@ -64,20 +40,18 @@ def initialize_uninitialized_global_variables(sess):
         sess.run(tf.variables_initializer(not_initialized_vars))
 
 
-def model_train(sess, x, y, predictions, X_train, Y_train, save=False,
-                predictions_adv=None, init_all=True, evaluate=None,
-                feed=None, args=None, rng=None, var_list=None):
+def model_train(sess, loss, x, y, X_train, Y_train, save=False, init_all=True,
```
npapernot: This change needs to be backward compatible. Could you add 6 months deprecation warnings: e.g., saying that loss will have to be provided in the future etc...

cleverhans_tutorials/mnist_tutorial_cw.py
```diff
@@ -66,7 +66,7 @@ def mnist_tutorial_cw(train_start=0, train_end=60000, test_start=0,
     set_log_level(logging.DEBUG)
 
     # Get MNIST test data
-    X_train, Y_train, X_test, Y_test = data_mnist(train_start=train_start,
+    x_train, y_train, x_test, y_test = data_mnist(train_start=train_start,
```
npapernot: Why did you change the name of these variables?

cleverhans_tutorials/mnist_tutorial_jsma.py
```diff
@@ -62,7 +62,7 @@ def mnist_tutorial_jsma(train_start=0, train_end=60000, test_start=0,
     set_log_level(logging.DEBUG)
 
     # Get MNIST test data
-    X_train, Y_train, X_test, Y_test = data_mnist(train_start=train_start,
+    x_train, y_train, x_test, y_test = data_mnist(train_start=train_start,
```
npapernot: Why did you change the variable names?

cleverhans_tutorials/mnist_tutorial_tf.py
```diff
@@ -71,118 +70,107 @@ def mnist_tutorial(train_start=0, train_end=60000, test_start=0,
     sess = tf.Session(config=tf.ConfigProto(**config_args))
 
     # Get MNIST test data
-    X_train, Y_train, X_test, Y_test = data_mnist(train_start=train_start,
+    x_train, y_train, x_test, y_test = data_mnist(train_start=train_start,
```
npapernot: Same

cleverhans_tutorials/tutorial_models.py
```diff
@@ -7,183 +7,42 @@
...
-              ReLU(),
-              Conv2D(nb_filters * 2, (5, 5), (1, 1), "VALID"),
-              ReLU(),
-              Flatten(),
-              Linear(nb_classes),
-              Softmax()]
-
-    model = MLP(layers, input_shape)
-    return model
+class ModelBasicCNN(Model):
```
npapernot: @goodfeli can you confirm you are OK with changing the tutorial model in pure TF to this implementation?

examples/multigpu_advtrain/model.py
```diff
@@ -26,18 +27,179 @@ def unify_device_name(dname):
     return dname.lower().replace('device:', '')
 
 
+class MLP(Model):
```
npapernot: if we're going to keep MLP and the corresponding layer definitions here, would it make sense to keep it in the tutorials folders and add the new implementation based on `tf.layers` as a new implementation that will eventually replace it? (maybe through some deprecation warning?) 

@goodfeli ?

cleverhans/attacks.py
david-berthelot: It's not used anywhere as far as I can tell, it seems it was something inherited from Keras.
I'll put them back in for backward compatibility.

cleverhans/defenses.py
david-berthelot: These are instances used as defense, so I think it makes sense to have the API not depend on them.

cleverhans/loss.py
david-berthelot: I renamed because it was confusing as to where this function was used and for what purpose. Basically it's only used by attacks, so I wanted to clearly separate it from defense losses (which may also have softmax crossentropy but follow the Loss API).

cleverhans/model.py
david-berthelot: Yes it's not necessary, I'll put some defaults.

cleverhans_tutorials/mnist_tutorial_cw.py
david-berthelot: To be PEP8 compliant, I just did it mechanically since I was touching the file,

cleverhans_tutorials/mnist_tutorial_jsma.py
david-berthelot: To be PEP8 compliant, I just did it mechanically since I was touching the file,

examples/multigpu_advtrain/model.py
david-berthelot: It's only used in Examples now though, that's why I moved it here.
I reworked models into more tensorflow compliant APIs since (1) they do the same thing and (2) it's an additional hurdle to have to learn a mini-layer framework in addition to cleverhans.

cleverhans/model.py
david-berthelot: The purpose of these variables is not to type strings when writing models and as such to avoid typo-related bugs. In addition, you get auto-complete benefits and it's self-documented (you don't have to enumerate valid strings in the doc).

cleverhans/model.py
david-berthelot: Same response, let's keep the discussion about this on the first comment to avoid repetition.

cleverhans/model.py
david-berthelot: It's not used anywhere as far as I can tell, it seems it was something inherited from Keras.
I can put it back in for backward compatibility, but I don't know who would use this if it's not even used in the CleverHans code. What's your take on this?

cleverhans/model.py
david-berthelot: Same response, let's keep the discussion about this on the first comment to avoid repetition.

cleverhans_tutorials/mnist_tutorial_tf.py
david-berthelot: Same :p

cleverhans/utils_tf.py
david-berthelot: Yes.

cleverhans/utils_tf.py
david-berthelot: Yes. Basically I ended renaming the new function to model_train_loss.

cleverhans/model.py
david-berthelot: Same.

cleverhans/model.py
david-berthelot: It does not return logits but softmax(logits), which generate probabilities. I don't see why that would be confusing, isn't it typical to apply a softmax to logits to obtain probabilities?

cleverhans/attacks.py
goodfeli: I'd prefer to keep it too.

In my experimental directory at Google, it looks like I'm using this function about 50 times. So while it's not used within the library, it is used by users of the library, or at least it is if I'm a representative user.

cleverhans/defenses.py
goodfeli: I'd vote to put them in loss.py too

cleverhans/loss.py
```diff
@@ -0,0 +1,50 @@
...
+import os
+import tensorflow as tf
+
+from .model import Model
+
+
+class Loss(object):
+    """
+    An abstract interface for loss wrappers that allows flexible control of
+    real examples, adversarial examples and labels..
```
goodfeli: double period here

cleverhans/loss.py
goodfeli: I think while we're doing the redesign, maybe this would be a good chance to get rid of the "if probs.op.type == 'Softmax'" hack. We've used it for a long time knowing that it wasn't very safe (because sometimes the last op is a reshape, even if there is a softmax right before it).
I think it would be better to standardize on having the attacks call `get_logits` so that we don't need to ask for ancestors in the graph this way.

cleverhans/model.py
goodfeli: What does the O_ mean?

cleverhans/model.py
goodfeli: I'd also vote to keep this

cleverhans/model.py
goodfeli: Yeah, @david-berthelot is right, it returns the correct value

cleverhans/model.py
goodfeli: I think that using the scope to keep track of the parameters is too specific to put in the `Model` base class. We want the `Model` class to be able to wrap any kind of 3rd party code, including code written in other frameworks. Since not every framework will necessarily use scopes to keep track of which model owns each Variable, we shouldn't use that mechanism in the `Model` base class. Wrapping models in some frameworks will now require us to change which scope the variables from those frameworks live in.

cleverhans/utils_tf.py
goodfeli: I'd vote to just call the new function `train`. And we can leave `model_train` in place for 6 months for backwards compatibility.

cleverhans/utils_tf.py
```diff
@@ -109,13 +85,11 @@ def model_train(sess, x, y, predictions, X_train, Y_train, save=False,
...
-    # Define loss
-    loss = model_loss(y, predictions)
-    if predictions_adv is not None:
-        loss = (loss + model_loss(y, predictions_adv)) / 2
-
+    # Define optimizer
+    loss_value = loss.fprop(x, y, **fprop_args)
     train_step = tf.train.AdamOptimizer(learning_rate=args.learning_rate)
-    train_step = train_step.minimize(loss, var_list=var_list)
+    with tf.control_dependencies(tf.get_collection(tf.GraphKeys.UPDATE_OPS)):
```
goodfeli: Can you comment on what this line does? It might be OK, I just don't know about what this does.
Is there a way that we could avoid using global variables?
A lot of the time for research purposes we want to do things like train two different models in the same process, and global variables / global update lists make it easy to have bugs where we updated both models by accident when we meant to update only one.

cleverhans_tutorials/mnist_tutorial_cw.py
goodfeli: Yeah, I think this is more correct. I still write a lot of non-PEP8 names as an old habit, but probably better to fix it.

goodfeli: (Also, these variables aren't part of an API, so it won't break anyone's code)

cleverhans_tutorials/tutorial_models.py
goodfeli: since it's the tutorials no one should depend on this code and @david-berthelot says the runtime and accuracy are the same. so yeah, let's go for it

examples/multigpu_advtrain/model.py
goodfeli: I think it's fine to move it into the multi-GPU example if that's the only place it's still used. Since it was in the tutorial before, people shouldn't be depending on it. It's not part of the main library API.

goodfeli:
LGTM modulo the things we asked to change

cleverhans/attacks.py
npapernot: +1  (I also use that function a lot)

cleverhans_tutorials/mnist_tutorial_cw.py
npapernot: That makes a lot of sense, we should change the names in that case

cleverhans_tutorials/mnist_tutorial_jsma.py
npapernot: Thanks!

cleverhans_tutorials/mnist_tutorial_tf.py
npapernot: Thanks :)

cleverhans_tutorials/tutorial_models.py
npapernot: SGTM

examples/multigpu_advtrain/model.py
npapernot: SGTM

npapernot:
Same for me, LGTM once comments are addressed and the conflict w/ the master branch is resolved

cleverhans/loss.py
david-berthelot: I agree, but it's used quite extensively and there are already a lot of changes.

cleverhans/attacks.py
david-berthelot: Okay done.

cleverhans/defenses.py
david-berthelot: Okay done.

cleverhans/attacks.py
david-berthelot: Put it back, but extended it a bit to take **kwargs, it's still backward compatible but more flexible to pass arguments to fprop.

cleverhans/model.py
david-berthelot: It stands for output, I tend to prefix enums by a letter or a word so as to not confuse which function they are compatible with. I'm open to suggestions.

cleverhans/model.py
david-berthelot: I thought about it it's tricky decision:
1. if we want to avoid tensorflow in Model, we need to clean other functions (get_prob uses tf.nn.softmax for example, get_layer_names too..
2. say if you want to support Keras or PyTorch, you can overload the methods and nothing is lost (and they already do it anyway).

cleverhans/model.py
david-berthelot: Ok done putting it back.

cleverhans/model.py
david-berthelot: Ok done putting it back.

cleverhans/utils_tf.py
david-berthelot: Done

cleverhans/loss.py
david-berthelot: Done

cleverhans/utils_tf.py
david-berthelot: The globals in tf collections are local to the graph (I verified it by hand just one with a small example)
```
g1 = tf.Graph()
with g1.as_default():
  a1 = tf.get_variable('a', [], tf.float32)
  tf.add_to_collection(tf.GraphKeys.UPDATE_OPS, tf.assign_add(a1, 1))
  with tf.control_dependencies(tf.get_collection(tf.GraphKeys.UPDATE_OPS)):
    b1 = 2 * a1
  sess1 = tf.train.MonitoredSession()

g2 = tf.Graph()
with g2.as_default():
  a2 = tf.get_variable('a', [], tf.float32)
  tf.add_to_collection(tf.GraphKeys.UPDATE_OPS, tf.assign_add(a2, -1))
  tf.add_to_collection(tf.GraphKeys.UPDATE_OPS, tf.assign_add(a2, +3))
  with tf.control_dependencies(tf.get_collection(tf.GraphKeys.UPDATE_OPS)):
    b2 = 2 * a2
  sess2 = tf.train.MonitoredSession()

# On graph1
print(sess1.run(a1))
print(sess1.run(b1))  # Increments a1 by 1
print(sess1.run(a1))  # Indeed it was incremented

# Now on graph2
print(sess2.run(a2))
print(sess2.run(b2)) # Increments a2 by 2 (-1 + 3)
print(sess2.run(a2)) # Indeed it was incremented by 2
print(sess1.run(a1)) # However a1 was unchanged since it is in another graph.

# You can also inspect
with g1.as_default(): 
  print(tf.get_collection(tf.GraphKeys.UPDATE_OPS))
  # a single update in the list

with g2.as_default(): 
 print(tf.get_collection(tf.GraphKeys.UPDATE_OPS))
 # two updates in the list (-1 + 3)
```

david-berthelot: And yes I added a small comment.

goodfeli:
Looks like it needs a rebase

PR #428, opened 2018-06-18
Change in Attack.Py addressing Issue

### LINKED ISSUE #406: When Attack.get_or_guess_labels(), allow y_target to be None
To my understanding, when you call `FastGradientMethod.generate` you can't pass `y_target=None` as an argument and expect cleverhans to behave like you didn't pass it.

https://github.com/tensorflow/cleverhans/blob/fcf1e2a18e72dd42280f04a50fce3eb4e0b20fcf/cleverhans/attacks.py#L200-L205

Changing check into `'y_target' in kwargs and kwargs['y_target'] is not None` would make the function behave as one would expect.

katchu11:
Close #406


katchu11:
Ok thanks for the advice Giuseppe!

PR #430, opened 2018-06-20
added eager tutorial and attacks.

feedforward:
For eager execution. 
1. A subclass of cleverhans.model.Model for tf.eager models is added. (cleverhans.model.Model_Eager), 
this uses only tf_eager without any keras dependencies.

2. New file attacks_eager is added, (existing attacks run through session using feed_dict making it too hard to adapt.), 
but structure of attacks.py is adapted to attacks_eager.py,  so objects of both attack classes can be replaced with out any difference.

3. tutorial to show the use tutorial_mnist_eager.py

4. unit tests for all the above.

### DIFF
--- a/cleverhans/loss.py
+++ b/cleverhans/loss.py
@@ -56,7 +56,7 @@ class LossCrossEntropy(Loss):
         else:
             x = x,
 
-        # Catching RuntimeError: Variable -= value not supported.
+        # Catching RuntimeError: Variable -= value not supported by tf.eager.
         try:
             y -= self.smoothing * (y - 1. / tf.cast(y.shape[-1], tf.float32))
         except RuntimeError:
--- a/cleverhans/loss.py
+++ b/cleverhans/loss.py
@@ -55,7 +55,7 @@ class LossCrossEntropy(Loss):
             x = x, self.attack(x)
         else:
             x = x,
-        y -= self.smoothing * (y - 1. / tf.cast(y.shape[-1], tf.float32))
+        #y -= self.smoothing * (y - 1. / tf.cast(y.shape[-1], tf.float32))
         logits = [self.model.get_logits(x, **kwargs) for x in x]
         loss = sum(
             tf.nn.softmax_cross_entropy_with_logits(labels=y, logits=logit)
--- a/cleverhans/loss.py
+++ b/cleverhans/loss.py
@@ -55,7 +55,14 @@ class LossCrossEntropy(Loss):
             x = x, self.attack(x)
         else:
             x = x,
-        #y -= self.smoothing * (y - 1. / tf.cast(y.shape[-1], tf.float32))
+
+        # Catching RuntimeError: Variable -= value not supported.
+        try:
+            y -= self.smoothing * (y - 1. / tf.cast(y.shape[-1], tf.float32))
+        except RuntimeError:
+            y.assign_sub(self.smoothing * (y - 1. / tf.cast(y.shape[-1],
+                                                            tf.float32)))
+
         logits = [self.model.get_logits(x, **kwargs) for x in x]
         loss = sum(
             tf.nn.softmax_cross_entropy_with_logits(labels=y, logits=logit)
--- a/cleverhans/model.py
+++ b/cleverhans/model.py
@@ -68,8 +68,12 @@ class Model(object):
         Provides access to the model's parameters.
         :return: A list of all Variables defining the model parameters.
         """
-        scope_vars = tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES,
-                                       self.scope)
+        if tf.executing_eagerly() == False:
+            scope_vars = tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES,
+                                           self.scope)
+        else:
+            error = 'For Eager execution - get_params must be overridden.'
+            raise NotImplementedError()
         return scope_vars
 
     def get_layer_names(self):
--- a/cleverhans/model.py
+++ b/cleverhans/model.py
@@ -69,14 +69,17 @@ class Model(object):
         Provides access to the model's parameters.
         :return: A list of all Variables defining the model parameters.
         """
-        # For tf 1.4 - in which executing_eagerly throws an error.
+        # Catch eager execution and assert function overload.
         try:
             if tf.executing_eagerly():
                 raise NotImplementedError("For Eager execution - get_params "
                                           "must be overridden.")
         except AttributeError:
-            scope_vars = tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES,
-                                           self.scope)
+            pass
+
+        # For Graoh based execution
+        scope_vars = tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES,
+                                       self.scope)
         return scope_vars
 
     def get_layer_names(self):
--- a/cleverhans/model.py
+++ b/cleverhans/model.py
@@ -68,12 +68,12 @@ class Model(object):
         Provides access to the model's parameters.
         :return: A list of all Variables defining the model parameters.
         """
-        if tf.executing_eagerly() == False:
+        if tf.executing_eagerly():
+            raise NotImplementedError("For Eager execution - get_params must "
+                                      "be overridden.")
+        else:
             scope_vars = tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES,
                                            self.scope)
-        else:
-            error = 'For Eager execution - get_params must be overridden.'
-            raise NotImplementedError()
         return scope_vars
 
     def get_layer_names(self):
--- a/cleverhans/model.py
+++ b/cleverhans/model.py
@@ -1,5 +1,6 @@
 from abc import ABCMeta
 import tensorflow as tf
+from distutils.version import LooseVersion
 
 
 class Model(object):
@@ -68,10 +69,12 @@ class Model(object):
         Provides access to the model's parameters.
         :return: A list of all Variables defining the model parameters.
         """
-        if tf.executing_eagerly():
-            raise NotImplementedError("For Eager execution - get_params must "
-                                      "be overridden.")
-        else:
+        # For tf 1.4 - in which executing_eagerly throws an error.
+        try:
+            if tf.executing_eagerly():
+                raise NotImplementedError("For Eager execution - get_params "
+                                          "must be overridden.")
+        except AttributeError:
             scope_vars = tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES,
                                            self.scope)
         return scope_vars
--- a/cleverhans/utils_tfe.py
+++ b/cleverhans/utils_tfe.py
@@ -1,0 +1,232 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+
+import math
+import numpy as np
+import os
+import tensorflow as tf
+import time
+import warnings
+from six.moves import xrange
+
+from distutils.version import LooseVersion
+from cleverhans.loss import LossCrossEntropy
+from .utils import batch_indices, _ArgsWrapper, create_logger
+
+_logger = create_logger("cleverhans.utils.tfe")
+
+
+def train(model, X_train=None, Y_train=None, save=False,
+                predictions_adv=None, evaluate=None,
+                args=None, rng=None, var_list=None,
+                attack=None, attack_args=None):
+    """
+    Train a TF Eager model
+    :param model: instance of cleverhans model, takes in input batch,
+                    gives out probs(softmax layer).
+    :param X_train: numpy array with training inputs
+    :param Y_train: numpy array with training outputs
+    :param save: boolean controlling the save operation
+    :param predictions_adv: if set with the adversarial example tensor,
+                            will run adversarial training
+    :param evaluate: function that is run after each training iteration
+                     (typically to display the test/validation accuracy).
+    :param args: dict or argparse `Namespace` object.
+                 Should contain `nb_epochs`, `learning_rate`,
+                 `batch_size`
+                 If save is True, should also contain 'train_dir'
+                 and 'filename'
+    :param rng: Instance of numpy.random.RandomState
+    :param var_list: List of variables to train.
+    :param attack: Instance of the class cleverhans.attacks.attacks_eager
+    :param attack_args: Parameters required for the attack.
+    :return: True if model trained
+    """
+    args = _ArgsWrapper(args or {})
+    if ((attack is None) != (attack_args is None)):
+        error = "attack and attack_args must be passed together."
+        raise ValueError(error)
+    if X_train is None or Y_train is None:
+        raise ValueError("X_train argument and Y_train argument "
+                         "must be supplied.")
+    # Check that necessary arguments were given (see doc above)
+    assert args.nb_epochs, "Number of epochs was not given in args dict"
+    assert args.learning_rate, "Learning rate was not given in args dict"
+    assert args.batch_size, "Batch size was not given in args dict"
+
+    if save:
+        assert args.train_dir, "Directory for save was not given in args dict"
+        assert args.filename, "Filename for save was not given in args dict"
+
+    if rng is None:
+        rng = np.random.RandomState()
+
+    # Optimizer
+    tfe = tf.contrib.eager
+    optimizer = tf.train.AdamOptimizer(learning_rate=args.learning_rate)
+    batch_x = tfe.Variable(X_train[0:args.batch_size], dtype=tf.float32)
+    batch_y = tfe.Variable(Y_train[0:args.batch_size], dtype=tf.float32)
+
+    # One epoch of training.
+    for epoch in xrange(args.nb_epochs):
+        # Compute number of batches
+        nb_batches = int(math.ceil(float(len(X_train)) / args.batch_size))
+        assert nb_batches * args.batch_size >= len(X_train)
+
+        # Indices to shuffle training set
+        index_shuf = list(range(len(X_train)))
+        rng.shuffle(index_shuf)
+
+        prev = time.time()
+        for batch in range(nb_batches):
+
+            # Compute batch start and end indices
+            start, end = batch_indices(
+                batch, len(X_train), args.batch_size)
+
+            # Perform one training step
+            tf.assign(batch_x, X_train[index_shuf[start:end]])
+            tf.assign(batch_y, Y_train[index_shuf[start:end]])
+            # Compute grads
+            with tf.GradientTape() as tape:
+                # Define loss
+                loss_clean_obj = LossCrossEntropy(model, smoothing=0.)
+                loss_clean = loss_clean_obj.fprop(x=batch_x, y=batch_y)
+                loss = loss_clean
+                # Adversarial training
+                if attack is not None:
+                    batch_adv_x = attack.generate(batch_x, **attack_args)
+                    loss_adv_obj = LossCrossEntropy(model, smoothing=0.)
+                    loss_adv = loss_adv_obj.fprop(x=batch_adv_x, y=batch_y)
+                    loss = (loss_clean + loss_adv) / 2.0
+            # Apply grads
+            model_variables = model.get_params()
+            grads = tape.gradient(loss, model_variables)
+            optimizer.apply_gradients(zip(grads, model_variables))
+
+        assert end >= len(X_train)  # Check that all examples were used
+        cur = time.time()
+        _logger.info("Epoch " + str(epoch) + " took " +
+                     str(cur - prev) + " seconds")
+        if evaluate is not None:
+            evaluate()
+
+    if save:
+        save_path = os.path.join(args.train_dir, args.filename)
+        saver = tf.train.Saver()
+        saver.save(save_path, model_variables)
+        _logger.info("Completed model training and saved at: " +
+                     str(save_path))
+    else:
+        _logger.info("Completed model training.")
+
+    return True
+
+
+def model_eval(model, X_test=None, Y_test=None, args=None,
+               attack=None, attack_args=None):
+    """
+    Compute the accuracy of a TF Eager model on some data
+    :param model: instance of cleverhans.model.Model_Eager
+                    with pretrained weights for evaluation.
+    :param X_test: numpy array with training inputs
+    :param Y_test: numpy array with training outputs
+    :param args: dict or argparse `Namespace` object.
+                 Should contain `batch_size`
+    :param attack: instance of the class cleverhans.attacks.attacks_eager
+    :param attack_args: parameters required for the attack.
+    :return: a float with the accuracy value
+    """
+    args = _ArgsWrapper(args or {})
+
+    if ((attack is None) != (attack_args is None)):
+        error = "attack and attack_args must be passed together."
+        raise ValueError(error)
+    assert args.batch_size, "Batch size was not given in args dict"
+    if X_test is None or Y_test is None:
+        raise ValueError("X_test argument and Y_test argument "
+                         "must be supplied.")
+
+    # Define accuracy symbolically
+    if LooseVersion(tf.__version__) <= LooseVersion('1.0.0'):
+        raise Exception('Use Tensorflow Version greather than 1.0.0')
+
+    # Init result var
+    accuracy = 0.0
+
+    # Compute number of batches
+    nb_batches = int(math.ceil(float(len(X_test)) / args.batch_size))
+    assert nb_batches * args.batch_size >= len(X_test)
+
+    X_cur = np.zeros((args.batch_size,) + X_test.shape[1:],
+                     dtype=X_test.dtype)
+    Y_cur = np.zeros((args.batch_size,) + Y_test.shape[1:],
+                     dtype=Y_test.dtype)
+
+    tfe = tf.contrib.eager
+    batch_x = tfe.Variable(X_test[0:args.batch_size], dtype=tf.float32)
+    batch_y = tfe.Variable(Y_test[0:args.batch_size], dtype=tf.float32)
+    for batch in range(nb_batches):
+        if batch % 100 == 0 and batch > 0:
+            _logger.debug("Batch " + str(batch))
+
+        # Must not use the `batch_indices` function here, because it
+        # repeats some examples.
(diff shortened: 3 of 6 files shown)


npapernot:
@nottombrown @david-berthelot , would you be able to review this PR since you use this logic more often than I do?

nottombrown:
Hmmm, I haven't actually used `tf.eager` for a project yet. @david-berthelot have you?

cleverhans/model_eager.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+if LooseVersion(tf.__version__) < LooseVersion('1.8.0'):
+    error = ('For eager execution',
+             'use Tensorflow version greather than 1.8.0.')
+    print(error)
+    exit(1)
+
+
+class Model_Eager(Model, tf.contrib.eager.Network):
```
rajivhockey97: tfe.Network is deprecated in tensorflow rc 1.9.0. We could inherit from tf.keras.Model instead.

cleverhans/model.py
```diff
@@ -1,4 +1,5 @@
 from abc import ABCMeta
+import tensorflow as tf
```
david-berthelot: New import but not used.

cleverhans/model.py
```diff
@@ -101,7 +102,6 @@ def __init__(self, callable_fn, output_layer):
         :param output_layer: A string of the output layer returned by the
                              function. (Usually either "probs" or "logits".)
         """
-
```
david-berthelot: This change is unrelated to eager.

cleverhans/attacks_eager.py
```diff
@@ -0,0 +1,331 @@
...
+    exit(1)
+
+try:
+    from cleverhans.model_eager import Model_Eager
+except ImportError:
+    error = 'Import Error: cleverhans.model_eager'
+    raise ImportError(error)
+
+
+class Attack(object):
```
david-berthelot: I don't think you should redefine the base class Attack. May be it's an opportunity to refactor the base class from attacks.py so that you can inherit from it to form a new class (say AttackTFE).

cleverhans/attacks_eager.py
```diff
@@ -0,0 +1,331 @@
...
+    error = ('For eager execution',
+             'use Tensorflow version greather than 1.8.0.')
+    print(error)
+    exit(1)
+
+try:
+    from cleverhans.model_eager import Model_Eager
+except ImportError:
+    error = 'Import Error: cleverhans.model_eager'
+    raise ImportError(error)
```
david-berthelot: Just for my understanding, why do you intercept the import error if you're reraising it? Is it only to modify the error message? For what reason?

cleverhans/attacks_eager.py
```diff
@@ -0,0 +1,331 @@
...
+
+    def __init__(self, model, back='tfe', dtypestr='float32'):
+        """
+        :param model: An instance of the cleverhans.model.Model_Eager class.
+        :param back: The backend to use. Currently 'tfe' is the only option.
+        :param dtypestr: datatype of the input data samples and crafted
+                        adversarial attacks.
+        """
+        # Validate the input arguments.
+        if not(back == 'tfe'):
```
david-berthelot: A simpler way to do it is
if back != 'tfe':

cleverhans/attacks_eager.py
```diff
@@ -0,0 +1,331 @@
...
+            raise ValueError("Backend argument must be 'tfe'.")
+        if back == 'tfe':
```
david-berthelot: Isn't this if statement always true?

cleverhans/attacks_eager.py
```diff
@@ -0,0 +1,331 @@
...
+                                    logits=logits, labels=labels)
+        except AttributeError:
+            out = tf.nn.softmax_cross_entropy_with_logits(
+                                    logits=logits, labels=labels)
+        if mean:
+            out = tf.reduce_mean(out)
+        return out
+
+
+class ListOfAttacks(object):
```
david-berthelot: 1. What is the use case for this?
2. Why use a class instead of say?
ATTACKS = {'fsgm': FastGradientMethod}

cleverhans/attacks_eager.py
feedforward: tensorflow version (require >= 1.8.0) is to checked before importing Model_Eager, 
but PEP8 requires all import statements to be at beginning of the file.

i used try ... except as a hack around this deadlock, if there is an elegant solution please let me know.

cleverhans/attacks_eager.py
feedforward: True, will update PR to reflect this change.

cleverhans/attacks_eager.py
feedforward: True, redundant snippet is used to stay similar to previous code in attacks.py 
`        if not(back == 'tf'):
            raise ValueError("Backend argument must be 'tf'.")

        if back == 'tf':
            import tensorflow as tf
            self.tf_dtype = tf.as_dtype(dtypestr)`
strictly speaking this is not necessary, let me know what you think.

cleverhans/attacks_eager.py
feedforward: In cleverhans_tutorials for tf backend each attack has one tutorial, 
but for eager backend we can have one tutorial that accepts --attack as a argument.
mnist_tutorial_eager.py expects  a string for --attack argument, (eg: 'fgsm') and maps it to Attack class 
 (eg: FastGradientMethod).

This class is used for above mapping attack name (string) to Attack class, to reduce code repetition, what are your thoughts on this ?

cleverhans/model.py
feedforward: True, updated PR will reflect the change.

cleverhans/model.py
feedforward: True, updated PR will reflect the change.

cleverhans/model_eager.py
feedforward: #430 , suggestions were in favour of not using the keras dependencies.

cleverhans/attacks_eager.py
feedforward: I agree, 
few options i could think of, let me know your thoughts on which one looks best .... 
Option 1:
base class Attack uses feed_dict and runs through session, inheriting AttackTFE from Attack will bring graph based code structure to AttackTFE, but we can overload the functions, 

PRO: this way no changes are required to existing code base,  
CONS: unnecessary graph def code inherited to AttackTFE, not elegant.

Option 2:
Add a new base class Attack_BASE to contain functions common to Attack and AttackTFE, then inherit both these classes from Attack_BASE.
PRO: hierarchical structure is maintained.
CON: few changes to the existing code base.

feedforward:
@david-berthelot  requested changes are applied, with replies to the your comments, let me know what your thoughts are, particularly on handling AttackTFE class

cleverhans/attacks_eager.py
david-berthelot: Shouldn't it raise an error if tf version is too old and eager is not present?
What's wrong with just:
```
from cleverhans.model_eager import Model_Eager
```
If it's present, it gets imported and if it's not present it will raise an import error.

cleverhans/attacks_eager.py
david-berthelot: I don't know I didn't review the previous code but it seems redundant just the same. No need to clean all existing code but writing new code you can fix stuff that you notice along the way.

cleverhans/attacks_eager.py
david-berthelot: Right but a dictionary is just as easy to import isn't it?
If the list is for the tutorial, may be move it to the tutorial.
Another problem with the list, is that it has to be manually maintained, someone adding a new attack could forget to put it in the list, although if it's just in the tutorial folder it's okay.

cleverhans/utils_eager.py
```diff
@@ -0,0 +1,280 @@
...
+    x_shape = tf.shape(x)
+    x = tf.contrib.layers.flatten(x)
+    x /= (epsilon + tf.reduce_max(tf.abs(x), 1, keep_dims=True))
+    square_sum = tf.reduce_sum(tf.square(x), 1, keep_dims=True)
+    x_inv_norm = tf.rsqrt(np.sqrt(epsilon) + square_sum)
+    x_norm = tf.multiply(x, x_inv_norm)
+    return tf.reshape(x_norm, x_shape, scope)
+
+
+def kl_with_logits(p_logits, q_logits):
```
david-berthelot: What is specific to eager in this function?

cleverhans/utils_eager.py
```diff
@@ -0,0 +1,280 @@
...
+    :param epsilon: stabilizes division
+    :return: the batch of l2 normalized vector
+    """
+    x_shape = tf.shape(x)
+    x = tf.contrib.layers.flatten(x)
+    x /= (epsilon + tf.reduce_max(tf.abs(x), 1, keep_dims=True))
+    square_sum = tf.reduce_sum(tf.square(x), 1, keep_dims=True)
+    x_inv_norm = tf.rsqrt(np.sqrt(epsilon) + square_sum)
+    x_norm = tf.multiply(x, x_inv_norm)
+    return tf.reshape(x_norm, x_shape, scope)
```
david-berthelot: I am confused by this function:
1. what is eager specific or should it just be a tf util?
2. I don't understand how this L2 batch normalization (of vectors I assume)
- you first divide x by max(abs(x_i)) which doesn't seem be related to L2 normalization.
- also why take the sqrt of epsilon, I am not saying this wrong, but I don't see the reason.
- you can simplify the code x_norm = x * x_inv_norm

cleverhans/attacks_eager.py
feedforward: True, above code is redundant now, will modify it in updated PR.

cleverhans/attacks_eager.py
feedforward: Sure, will make the changes in updated PR.

cleverhans/attacks_eager.py
feedforward: True, i will move this class to tutorial, 
dictionary is just as easy to import and serves the purpose, modified in the updated PR.

cleverhans/utils_eager.py
feedforward: True, there is nothing eager specific to this function, haven't noticed that before. 
removed it.

cleverhans/utils_eager.py
feedforward: SAME as previous comment.

cleverhans/attacks_eager.py
```diff
@@ -0,0 +1,260 @@
...
+        self.tf_dtype = tf.as_dtype(dtypestr)
+        self.np_dtype = np.dtype(dtypestr)
+
+        if not isinstance(model, Model_Eager):
+            raise ValueError("The model argument should be an instance of"
+                             " the cleverhans.model.Model_Eager class.")
+        # Prepare attributes
+        self.layer_names = []
+
+    def get_or_guess_labels(self, x, kwargs):
```
david-berthelot: What is specific to eager in this method? It looks very similar to the one in tf_attacks.py

cleverhans/attacks_eager.py
```diff
@@ -0,0 +1,260 @@
...
+        :param x_val: A NumPy array with the original inputs.
+        :param **kwargs: optional parameters used by child classes.
+        :return: A NumPy array holding the adversarial examples.
+        """
+        tfe = tf.contrib.eager
+        x = tfe.Variable(x_val)
+        adv_x = self.generate(x, **kwargs)
+        return adv_x.numpy()
+
+    def clip_eta(self, eta, ord, eps):
```
david-berthelot: What is specific to eager in this method? It looks very similar to the one in tf_attacks.py

cleverhans/attacks_eager.py
```diff
@@ -0,0 +1,260 @@
...
+        Create a FastGradientMethod instance.
+        Note: the model parameter should be an instance of the
+        cleverhans.model.Model abstraction provided by CleverHans.
+        """
+        if not isinstance(model, Model_Eager):
+            model = CallableModelWrapper(model, 'probs')
+
+        super(FastGradientMethod, self).__init__(model, dtypestr)
+
+    def generate(self, x, **kwargs):
```
david-berthelot: What is specific to eager in this method? It looks very similar to the one in tf_attacks.py

cleverhans/attacks_eager.py
```diff
@@ -0,0 +1,260 @@
...
+                            Labels should be one-hot-encoded.
+        :param clip_min: (optional float) Minimum input component value
+        :param clip_max: (optional float) Maximum input component value
+        """
+        # Parse and save attack-specific parameters
+        assert self.parse_params(**kwargs)
+        labels, nb_classes = self.get_or_guess_labels(x, kwargs)
+        return self.fgm(x, labels=labels, targeted=(self.y_target is not None))
+
+    def parse_params(self, eps=0.3, ord=np.inf, y=None, y_target=None,
```
david-berthelot: What is specific to eager in this method? It looks very similar to the one in tf_attacks.py

cleverhans/attacks_eager.py
```diff
@@ -0,0 +1,260 @@
...
+        self.clip_max = clip_max
+
+        if self.y is not None and self.y_target is not None:
+            raise ValueError("Must not set both y and y_target")
+        # Check if order of the norm is acceptable given current implementation
+        if self.ord not in [np.inf, int(1), int(2)]:
+            raise ValueError("Norm order must be either np.inf, 1, or 2.")
+        return True
+
+    def fgm(self, x, labels, targeted=False):
```
david-berthelot: Seems like the only eager specific code, so why not inherit from the tf Attack and just redefine this function.
You could also overload the create graph one to return an error since I believe you don't create on in eager.

cleverhans/utils_eager.py
```diff
@@ -0,0 +1,252 @@
...
+import tensorflow as tf
+import time
+import warnings
+
+from .utils import batch_indices, _ArgsWrapper, create_logger
+
+_logger = create_logger("cleverhans.utils.eager")
+
+
+def model_loss(logits, labels, mean=True):
```
david-berthelot: What is specific to eager in this method? It looks very similar to the one in utils.tf

cleverhans/utils_eager.py
```diff
@@ -0,0 +1,252 @@
...
+    except AttributeError:
+        out = tf.nn.softmax_cross_entropy_with_logits(
+                                                logits=logits, labels=labels)
+
+    if mean:
+        out = tf.reduce_mean(out)
+    return out
+
+
+def model_train(model, X_train=None, Y_train=None, save=False,
```
david-berthelot: This function follows the deprecated mode, it does not use a loss. Check the function train, which replaces it.

cleverhans_tutorials/mnist_tutorial_eager.py
```diff
@@ -0,0 +1,255 @@
...
+
+FLAGS = flags.FLAGS
+
+
+# Keeps track of implemented attacks.
+# Maps attack string taken bash to attack class
+# -- 'fgsm' : FastGradientMethod
+
+ListOfAttacks = {
+                'fgsm': FastGradientMethod
```
david-berthelot: This doesn't seem to follow PEP8 alignment, or does it?

cleverhans_tutorials/tutorial_models_eager.py
```diff
@@ -0,0 +1,214 @@
+"""
```
david-berthelot: We moved away from custom layers in tutorials, I believe the tf.layers work with tf.eager by default. You can simplify this file a lot by using them.

cleverhans/attacks_eager.py
feedforward: I see your point, do you suggest overloading Attack class definitions ?? (i'm open to this ... )
my thoughts -- 
in attacks_tf.py all the functions are defined as part of graph execution and run through session.
in eager mode they run on the fly.
code logic is similar by design, (they are meant to do same thing in different context.)

cleverhans/attacks_eager.py
feedforward: Same as before, lets discuss in the previous comment.

cleverhans/attacks_eager.py
feedforward: Same as before, lets discuss in the previous comment.

cleverhans/attacks_eager.py
feedforward: Same as before, lets discuss in the previous comment.

cleverhans/attacks_eager.py
feedforward: True, this can be one way to do it, let's finalise the approach in previous comment, then i will submit an updated PR.

cleverhans/utils_eager.py
feedforward: Ohh, this change was done, after this PR is submitted, that is not an issue, i can update this accordingly.

cleverhans/utils_eager.py
feedforward: True, will import it.

cleverhans_tutorials/mnist_tutorial_eager.py
feedforward: Looks misaligned, Not sure, it did pass the travis test though

cleverhans_tutorials/tutorial_models_eager.py
feedforward: above change was done after this PR is submitted, i will update this in next PR.

cleverhans/attacks_eager.py
david-berthelot: Yes, that's exactly my thoughts, the code is the same except for gradient the tape and execution, so inheriting from tf_attack so that the eager code is minimal is a much more impressive demonstration of its simplicity than having a lot of duplicate which makes more intimidating.

cleverhans_tutorials/tutorial_models_eager.py
david-berthelot: I'm not sure what's best here. I'll say check with Nicolas / Ian and see what they think. I personally think it's best to have a PR with all the up-to-date code rather than pushing obsolete code at commit time.

david-berthelot:
A few more changes and it'll be done (the refactor of attack to reuse tf as much as possible since it's meant to be the same). Also need 3rd opinion on committing obsolete / deprecated code.

goodfeli:
Re deprecated code: generally we should not add deprecated code to the main repo. The idea of deprecated code is that 3rd party users can continue using it for a while, the library may not immediately *remove* uses of deprecated code, but the library shouldn't ever *add* uses of deprecated code.

cleverhans/attacks_eager.py
feedforward: Sure, makes sense, will do that ...

feedforward:
Refactored to
1. Inhereted from Attack class.
2. Uses recently merged loss class
3. Uses network similar to recently merged MakeBasicCNN

feedforward:
Refactored eager code.

cleverhans/loss.py
```diff
@@ -55,7 +55,7 @@ def fprop(self, x, y, **kwargs):
             x = x, self.attack(x)
         else:
             x = x,
-        y -= self.smoothing * (y - 1. / tf.cast(y.shape[-1], tf.float32))
+        #y -= self.smoothing * (y - 1. / tf.cast(y.shape[-1], tf.float32))
```
david-berthelot: Did you comment this code by accident?

cleverhans/model.py
```diff
@@ -68,8 +68,12 @@ def get_params(self):
         Provides access to the model's parameters.
         :return: A list of all Variables defining the model parameters.
         """
-        scope_vars = tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES,
-                                       self.scope)
+        if tf.executing_eagerly() == False:
```
david-berthelot: if tf.executing_eagerly() == False:

Simpler to write:
if not tf.executing_eagerly():

Or even better:
if tf.executing_eagerly():
   # raise
else:
  scope_vars

cleverhans/model.py
```diff
@@ -68,8 +68,12 @@ def get_params(self):
...
+            scope_vars = tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES,
+                                           self.scope)
+        else:
+            error = 'For Eager execution - get_params must be overridden.'
+            raise NotImplementedError()
```
david-berthelot: Did you forget to put the error in the exception?

cleverhans/utils_tfe.py
```diff
@@ -0,0 +1,232 @@
...
+                 and 'filename'
+    :param rng: Instance of numpy.random.RandomState
+    :param var_list: List of variables to train.
+    :param attack: Instance of the class cleverhans.attacks.attacks_eager
+    :param attack_args: Parameters required for the attack.
+    :return: True if model trained
+    """
+    args = _ArgsWrapper(args or {})
+    if ((attack is None) != (attack_args is None)):
+        error = "attack and attack_args must be passed together."
```
david-berthelot: Why use an intermediate "error" variable?

cleverhans/model.py
david-berthelot: In addition, I don't feel the intermediate variable "error" is needed here.

cleverhans/utils_tfe.py
```diff
@@ -0,0 +1,232 @@
...
+    :param args: dict or argparse `Namespace` object.
+                 Should contain `batch_size`
+    :param attack: instance of the class cleverhans.attacks.attacks_eager
+    :param attack_args: parameters required for the attack.
+    :return: a float with the accuracy value
+    """
+    args = _ArgsWrapper(args or {})
+
+    if ((attack is None) != (attack_args is None)):
+        error = "attack and attack_args must be passed together."
```
david-berthelot: Same comment about intermediate variable, it looks simpler to just pass the string to ValueError.

cleverhans/utils_tfe.py
```diff
@@ -0,0 +1,232 @@
...
+                      taken from FLAGS.train_dir and FLAGS.filename
+    :return:
+    """
+    assert file_path != None, 'pretrained model ckpt is not given.' 
+    saver = tf.train.Saver()
+    saver.restore(file_path)
+    return True
+
+
+def model_argmax(model, samples,):
```
david-berthelot: Trailing comma.

cleverhans_tutorials/mnist_tutorial_tfe.py
```diff
@@ -0,0 +1,253 @@
...
+tfe = tf.contrib.eager
+
+FLAGS = flags.FLAGS
+
+
+# Keeps track of implemented attacks.
+# Maps attack string taken from bash to attack class
+# -- 'fgsm' : FastGradientMethod
+
+ListOfAttacks = {
```
david-berthelot: 1. It is confusing to call a dictionary ListOfAttacks.
2. It's a global constant, so it should be all uppercase.

How about AVAILABLE_ATTACKS?

cleverhans_tutorials/mnist_tutorial_tfe.py
```diff
@@ -0,0 +1,253 @@
...
+    :param attack_string: adversarial attack name in string format
+    :return: attack class defined in cleverhans.attacks_eager
+    """
+
+    # List of Implemented attacks
+    attacks_list = ListOfAttacks.keys()
+
+    # Checking for  requested attack in list of available attacks.
+    if attack_string is None:
+        error = 'Attack type is not specified, list of available attacks\t'
```
david-berthelot: Same, it's simpler to just put the string directly in the exception line.

cleverhans_tutorials/mnist_tutorial_tfe.py
```diff
@@ -0,0 +1,253 @@
...
+        error += attacks_list
+        raise AttributeError(error)
+    if attack_string not in attacks_list:
+        error = 'Attack not available, list of available attacks\t'
```
david-berthelot: Same, it's simpler to just put the string directly in the exception line.

cleverhans_tutorials/mnist_tutorial_tfe.py
```diff
@@ -0,0 +1,253 @@
...
+        'nb_epochs': nb_epochs,
+        'batch_size': batch_size,
+        'learning_rate': learning_rate
+    }
+
+    # Initialize the attack object
+    attack_class = attack_selection(attack_string)
+    attack_params = {
+        'eps': 0.3, 'clip_min': 0.,
+        'clip_max': 1.}
```
david-berthelot: Why put this on an extra line when you still have space on the previous one?

cleverhans/utils_tfe.py
feedforward: True, removed.

cleverhans_tutorials/mnist_tutorial_tfe.py
feedforward: same as above.

cleverhans_tutorials/mnist_tutorial_tfe.py
feedforward: True, removed the extra line.

goodfeli:
is this ready for re-review?

cleverhans/loss.py
```diff
@@ -55,7 +55,14 @@ def fprop(self, x, y, **kwargs):
             x = x, self.attack(x)
         else:
             x = x,
-        #y -= self.smoothing * (y - 1. / tf.cast(y.shape[-1], tf.float32))
+
+        # Catching RuntimeError: Variable -= value not supported.
```
david-berthelot: Not supported "by tf.eager".

cleverhans_tutorials/mnist_tutorial_tfe.py
```diff
@@ -50,7 +48,7 @@
 # Maps attack string taken from bash to attack class
 # -- 'fgsm' : FastGradientMethod
 
-ListOfAttacks = {
+AVAILABLE_ATTACKS = {
                  'fgsm': FastGradientMethodTFE
```
david-berthelot: The alignment looks strange here, should be 4 characters?

feedforward:
The final changes are done. @goodfeli  Please have a look at this.

feedforward:
@npapernot Hi, review process is done, @david-berthelot approved the final version, please merge this PR.

PR #437, opened 2018-07-01
Flags removal

feedforward:
#281 - Removed the flags from utils_tf.py, uses file_path.

#397 code update to remove tensorflow depreciation warnings for tutorials.

### DIFF
--- a/cleverhans/utils_tf.py
+++ b/cleverhans/utils_tf.py
@@ -34,7 +34,18 @@ def model_loss(y, model, mean=True):
     else:
         logits = model
 
-    out = tf.nn.softmax_cross_entropy_with_logits(logits=logits, labels=y)
+    try:
+        y = tf.stop_gradient(y)
+        out = tf.nn.softmax_cross_entropy_with_logits_v2(
+                                                logits=logits, labels=y)
+    except AttributeError:
+        warning = "Running on tensorflow version " + \
+                   LooseVersion(tf.__version__).vstring + \
+                   ". This version will not be supported by cleverhans" + \
+                   "in the future."
+        _logger.warn(warning)
+        out = tf.nn.softmax_cross_entropy_with_logits(
+                                                logits=logits, labels=y)
 
     if mean:
         out = tf.reduce_mean(out)
@@ -256,10 +267,8 @@ def tf_model_load(sess, file_path=None):
     with sess.as_default():
         saver = tf.train.Saver()
         if file_path is None:
-            warnings.warn("Please provide file_path argument, "
-                          "support for FLAGS.train_dir and FLAGS.filename "
-                          "will be removed on 2018-04-23.")
-            file_path = os.path.join(FLAGS.train_dir, FLAGS.filename)
+            error = 'file_path argument is missing.'
+            raise ValueError(error)
         saver.restore(sess, file_path)
 
     return True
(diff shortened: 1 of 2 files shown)


cleverhans/utils_tf.py
```diff
@@ -411,3 +415,29 @@ def clip_eta(eta, ord, eps):
         factor = tf.minimum(1., eps / norm)
         eta = eta * factor
     return eta
+
+
+def tf_reduce_sum(input_tensor, axis=None, keepdims=None,
```
npapernot: Given that the code that this function is looking to support is only included in one tutorial folder script, I would rather modify the tutorial's logic directly at line 79 of the `tutorial_models.py` file you modified below and remove this function from the `utils_tf.py` file (we don't want to artificially extend the lifetime of features deprecated in TF by adding wrappers unless the code affected is directly part of the CleverHans core module).

feedforward: yeah makes sense to me.
few issues here, need your suggestion.

i was working on #397, attacks.py has the same depreciation warnings.
tf 1.4 - expects keep_dims and throws error for keepdims.
tf 1.8 - expects keepdims and throws warning for keep_dims.

are we planning to have long time support for tf 1.4 ?

npapernot: Thanks for pointing this out. In that case, I would recommend following the style adopted in `model_train` https://github.com/tensorflow/cleverhans/blob/master/cleverhans/utils_tf.py#L121

As far as I can tell, there are 6 lines in `attacks.py` that are potentially affected by this change in TF naming convention for keep_dims/keepdims. I would recommend modifying these 6 lines directly into something like:

```
if tf above 1.8 
    use keepdims 
else 
    use keep_dims 
    warn user that CleverHans may not support this version of TF in the future
```

feedforward: sure, i will write a if else condition inplace for ( utils_tf - 1, attacks.py - 6, attacks_tf.py - 6) these locations, we can remove them if and when tf 1.4 is not supported.

cleverhans/utils_tf.py
```diff
@@ -34,7 +34,13 @@ def model_loss(y, model, mean=True):
     else:
         logits = model
 
-    out = tf.nn.softmax_cross_entropy_with_logits(logits=logits, labels=y)
+    try:
+        y = tf.stop_gradient(y)
+        out = tf.nn.softmax_cross_entropy_with_logits_v2(
+                                                logits=logits, labels=y)
+    except AttributeError:
+        out = tf.nn.softmax_cross_entropy_with_logits(
```
npapernot: It would be great to add a warning here saying that this version of TF will not be supported by CleverHans in the future

feedforward: Sounds good, will do that.

feedforward: Please merge the PR, after i add the warning will submit a new PR for TF depreciation warnings.(there are warnings other than keep_dims, will resolve all in new PR )

npapernot: Thanks, I'll do that! Looks like your last commit broke the diff though (probably due to a conflict?)

feedforward: Yes there was some conflict, everything is sorted now.

PR #439, opened 2018-07-01
2.1 update

ysharma1126:
It's been found in recent results that increasing beta from 1e-3 to 1e-2 better leverages EAD's advantageous properties, therefore updated the default params to reflect that.


ysharma1126:
Oh it seems my pull requests were merged. The other commit is updating setup.py to reflect Cleverhans 2.1

PR #440, opened 2018-07-02
Consistent naming scheme across attacks

aidangomez:
WIP

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -15,1346 +15,1349 @@ _logger = utils.create_logger("cleverhans.attacks")
 
 
 class Attack(object):
-    """
+  """
     Abstract base class for all attack classes.
     """
-    __metaclass__ = ABCMeta
+  __metaclass__ = ABCMeta
 
-    def __init__(self, model, back='tf', sess=None, dtypestr='float32'):
-        """
-        :param model: An instance of the cleverhans.model.Model class.
-        :param back: The backend to use. Currently 'tf' is the only option.
-        :param sess: The tf session to run graphs in
-        """
-        if not (back == 'tf'):
-            raise ValueError("Backend argument must be 'tf'.")
+  def __init__(self, model, back='tf', sess=None, dtypestr='float32'):
+    """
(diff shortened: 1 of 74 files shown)


npapernot:
I think it would make sense to discuss this change on the cleverhans-dev mailing list first because it would be a major API change

npapernot:
(just to be clear, I think coming up with an unified terminology for the different attack parameters is a good idea and would be beneficial to the community if done based on a consensus)

goodfeli:
I think the result of the discussion on the mailing list was that we're OK with it, right @npapernot ?
However, the PEP8 tests are currently failing, because some lines are too long now.

npapernot:
Yes there were no objections on the mailing list so we are good to proceed with the change

npapernot:
@aidangomez any updates?

aidangomez:
@npapernot sorry, super preoccupied with work at the moment. Will return to this shortly.

npapernot:
Ping @aidangomez ?

aidangomez:
Okay so since the diffs are unreadable here is what has been changed:
  - nb_iters -> step_count
  - eps_iter -> step_size
  - epsilon -> eps (or eps_iter depending on the intended use)
  - 2 space indent following the Tensorflow style guide

npapernot:
Thanks for making the changes @aidangomez 

I talked about this offline w/ @goodfeli. We'd rather avoid merging such a large PR right before ICLR so we'll hold on this for another 2 weeks.

It would also be cleaner for the history to make two different PRs: one that changes to 2 space indents and one that changes the names. Would you be able to work on that after ICLR or should I take care of it?

aidangomez:
@npapernot Hmm so I was thinking about this and maybe we should create a decorator that warns when someone is using an deprecated name?

npapernot:
SGTM, will check in after ICLR

goodfeli:
Let's close this PR for now. After the ICLR deadline passes, let's do one mega-PR that changes all the whitespace. After that, let's redo the other changes proposed here.

Issue #400, opened 2018-05-03
Substitue training optimizes for both BlackBox variables and Substitute variables


VigneshSrinivasan10:
Hi, 

Thanks for the great repository. 

In the tutorial code `mnist_blackbox.py`, The blackbox is first prepared and then a substitute is trained using a subset of the test samples. When the subsitute is trained with the function `model_train`, 

https://github.com/tensorflow/cleverhans/blob/6b065c4507f50e2545c009f75e82678f92353163/cleverhans_tutorials/mnist_blackbox.py#L150

I notice that at this line, 

https://github.com/tensorflow/cleverhans/blob/6b065c4507f50e2545c009f75e82678f92353163/cleverhans/utils_tf.py#L117

the optimizer minimizes the loss. Since there is no `var_list` defined, it considers the variables of both the Blackbox and Substitute models. 

I would suggest adding another parameter to `train_params` - `tvars` which contains the variables list to optimize for. 

Best,

npapernot:
Thanks for flagging the issue. If you're able to put together a PR, feel free to assign me as a reviewer.

VigneshSrinivasan10:
Thanks for the quick response. 

I will be happy to place a PR - but unfortunately I am on time crunch - so perhaps, I can do it next month.

PR #398, opened 2018-05-02
add `Model.get_params` and `var_list` argument to `model_train`

goodfeli:

PR #442, opened 2018-07-03
Add handling for training keras models in mnist_blackbox tutorial

aspratyush:
The patch in #398, also discussed in #400, adds `model.get_params()` to `model_train`. However, since `get_params()` is a tensorflow specific API, it causes a regression when using Keras-based model. This PR checks the model type, uses a conditional `model_train` implementation, and also sets `model.trainable=False` when using Keras-based models.

### DIFF
--- a/cleverhans_tutorials/mnist_blackbox.py
+++ b/cleverhans_tutorials/mnist_blackbox.py
@@ -16,7 +16,6 @@ from six.moves import xrange
 
 import logging
 import tensorflow as tf
-import keras
 from tensorflow.python.platform import flags
 
 from cleverhans.loss import LossCrossEntropy
--- a/cleverhans_tutorials/mnist_blackbox.py
+++ b/cleverhans_tutorials/mnist_blackbox.py
@@ -243,7 +243,6 @@ def mnist_blackbox(train_start=0, train_end=60000, test_start=0,
 
     # Train substitute using method from https://arxiv.org/abs/1602.02697
     print("Training the substitute model.")
-
     train_sub_out = train_sub(sess, x, y, bbox_preds, X_sub, Y_sub,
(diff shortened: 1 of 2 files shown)


npapernot:
Thanks for the PR. Unfortunately, we decided a few months ago to remove all Keras dependencies from the CleverHans module and tutorials to the exception of two files: `cleverhans/utils_keras.py` and `cleverhans_tutorials/mnist_tutorial_keras_tf.py`. This is primarily because supporting Keras created a lot of overhead so I would rather move a blackbox tutorial using Keras to the `examples/` folder where we have looser constraints for merging code. 

@goodfeli  any opinion?

goodfeli:
Yes, I agree. I don't want to provide any ongoing support for more files using Keras.

@aspratyush , please let us know whether you'd prefer to move this to `examples/` (where there is no obligation for ongoing support) or whether you'd prefer to close the PR.

In general, it's a good idea to write to cleverhans-dev@googlegroups.com before starting work on a feature to make sure it will be approved.

npapernot:
@aspratyush Do you want to move this file to `examples/` or should I close this PR? Thanks!

aspratyush:
@npapernot : Have moved the keras-based black-box model file to `examples/`

While with the recent update to `cleverhans.utils_tf.train` that makes `var_list` an optional parameter resolves this specific regression, usage of `cleverhans.loss.LossCrossEntropy` requires that the defined model be an instance of `cleverhans.model.Model`. Hence, a Keras-based black-box model would need to be wrapped in a `cleverhans.utils_keras.KerasModelWrapper` to successfully run this example.

This updated PR addresses handling such an implementation.

npapernot:
Thanks!

PR #444, opened 2018-07-03
Generalizing the cleverhans_tutorial for non-MNIST dataset 

aashish-kumar:
The cleverhans_tutorial has been written only for MNIST with extensive hard coding of data-set parameters(like image width/height/channels. I have modified the code to handle other data-sets. The user need to only modify utils_mnist.py for his dataset and rest of the code will work out of box. Also I made sure the tutorials work for all image-size & channels. Image parameter are obtained from the X_train variable loaded by utils_mnist.py

### DIFF
--- a/cleverhans_tutorials/mnist_tutorial_tf.py
+++ b/cleverhans_tutorials/mnist_tutorial_tf.py
@@ -116,7 +116,7 @@ def mnist_tutorial(train_start=0, train_end=60000, test_start=0,
 #            assert X_test.shape[0] == test_end - test_start, X_test.shape
             print('Test accuracy on legitimate examples: %0.4f' % acc)
         model_train(sess, x, y, preds, X_train, Y_train, evaluate=evaluate,
-                    args=train_params, rng=rng)
+                    args=train_params, rng=rng, var_list=model.get_params())
 
         # Calculate training error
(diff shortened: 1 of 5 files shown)


aashish-kumar:
I will re-submit after fixing PEP8 Errors

cleverhans_tutorials/mnist_tutorial_tf.py
```diff
@@ -108,10 +113,10 @@ def evaluate():
             acc = model_eval(
                 sess, x, y, preds, X_test, Y_test, args=eval_params)
             report.clean_train_clean_eval = acc
-            assert X_test.shape[0] == test_end - test_start, X_test.shape
+#            assert X_test.shape[0] == test_end - test_start, X_test.shape
             print('Test accuracy on legitimate examples: %0.4f' % acc)
         model_train(sess, x, y, preds, X_train, Y_train, evaluate=evaluate,
-                    args=train_params, rng=rng, var_list=model.get_params())
```
goodfeli: Please keep the var_list argument. It makes it less likely to cause global variable bugs when copy-pasting code.

aashish-kumar:
@goodfeli  This is an older pull request. I have taken these into account in #445

PR #445, opened 2018-07-03
Fixing the tutorial for non-MNIST datasets

aashish-kumar:
The cleverhans_tutorial has been written only for MNIST with extensive hard coding of data-set parameters(like image width/height/channels. I have modified the code to handle other data-sets. The user need to only modify utils_mnist.py for his dataset and rest of the code will work out of box. Also I made sure the tutorials work for all image-size & channels. Image parameter are obtained from the X_train variable loaded by utils_mnist.py

Conformed to PEP8

### DIFF
--- a/cleverhans_tutorials/mnist_tutorial_cw.py
+++ b/cleverhans_tutorials/mnist_tutorial_cw.py
@@ -129,7 +129,7 @@ def mnist_tutorial_cw(train_start=0, train_end=60000, test_start=0,
     if targeted:
         if viz_enabled:
             # Initialize our array for grid visualization
-            grid_shape = (nb_classes, nb_classes, img_rows, img_cols, 
+            grid_shape = (nb_classes, nb_classes, img_rows, img_cols,
                           nchannels)
             grid_viz_data = np.zeros(grid_shape, dtype='f')
 
--- a/cleverhans_tutorials/mnist_tutorial_cw.py
+++ b/cleverhans_tutorials/mnist_tutorial_cw.py
@@ -67,8 +67,8 @@ def mnist_tutorial_cw(train_start=0, train_end=60000, test_start=0,
                                                   test_end=test_end)
 
     # Obtain Image Parameters
-    img_rows, img_cols, nchannels = X_train.shape[1:4]
-    nb_classes = Y_train.shape[1]
+    img_rows, img_cols, nchannels = x_train.shape[1:4]
+    nb_classes = y_train.shape[1]
 
     # Define input TF placeholder
     x = tf.placeholder(tf.float32, shape=(None, img_rows, img_cols,
--- a/cleverhans_tutorials/mnist_tutorial_cw.py
+++ b/cleverhans_tutorials/mnist_tutorial_cw.py
@@ -66,8 +66,7 @@ def mnist_tutorial_cw(train_start=0, train_end=60000, test_start=0,
                                                   test_start=test_start,
                                                   test_end=test_end)
 
-    # Use label smoothing
-    print(X_train.shape)
+    # Obtain Image Parameters
     img_rows, img_cols, nchannels = X_train.shape[1:4]
     nb_classes = Y_train.shape[1]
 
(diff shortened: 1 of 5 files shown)


aashish-kumar:
The auto-build has not started for sometime. please check.

aashish-kumar:
Changelog:-
1) Removed Hard-coded values and used the info from dataset to populate them:
      - Images:  (28,28,1)  to (img_rows,img_cols,nchannels).
      - Number of classes (10) to nb_classes  (some tutorials had it hard coded it while others were not using it for model generation)
 2)  Commented `assert X_test.shape[0] == test_end - test_start, X_test.shape`  : The assertion is too specific to MNIST 
3)  Made Label Smoothing optional(added an argument) as some tutorials are using it while others do not.
4) In Keras Tutorial: Changed the model dump from /tmp to local folder. This can be problem when multiple users on the same machine :- This one is debatable.
5) Removed the flag `-    flags.DEFINE_integer('nb_classes', 10, 'Number of output classes')` as changing this parameter doesn't actually change the model parameters. 

Comments?  @npapernot

cleverhans_tutorials/mnist_tutorial_cw.py
```diff
@@ -104,7 +107,7 @@ def mnist_tutorial_cw(train_start=0, train_end=60000, test_start=0,
     # Evaluate the accuracy of the MNIST model on legitimate test examples
     eval_params = {'batch_size': batch_size}
     accuracy = model_eval(sess, x, y, preds, X_test, Y_test, args=eval_params)
-    assert X_test.shape[0] == test_end - test_start, X_test.shape
+#    assert X_test.shape[0] == test_end - test_start, X_test.shape
```
goodfeli: What's going on here?

If there's a good reason to remove the test, it's fine to remove.

If there's no reason to remove it, I'd prefer to keep it. A bit of defensive programming helps to prevent bugs and catch them quickly when they arise.

We generally don't want commented code in the repo.

cleverhans_tutorials/mnist_tutorial_jsma.py
```diff
@@ -67,12 +62,20 @@ def mnist_tutorial_jsma(train_start=0, train_end=60000, test_start=0,
                                                   test_start=test_start,
                                                   test_end=test_end)
 
+    print(X_train.shape)
```
goodfeli: Let's not add debugging prints to the repo

cleverhans_tutorials/mnist_tutorial_jsma.py
```diff
@@ -94,7 +97,7 @@ def mnist_tutorial_jsma(train_start=0, train_end=60000, test_start=0,
     # Evaluate the accuracy of the MNIST model on legitimate test examples
     eval_params = {'batch_size': batch_size}
     accuracy = model_eval(sess, x, y, preds, X_test, Y_test, args=eval_params)
-    assert X_test.shape[0] == test_end - test_start, X_test.shape
+#    assert X_test.shape[0] == test_end - test_start, X_test.shape
```
goodfeli: Same issue with commented test here

cleverhans_tutorials/mnist_tutorial_tf.py
```diff
@@ -75,15 +76,20 @@ def mnist_tutorial(train_start=0, train_end=60000, test_start=0,
                                                   train_end=train_end,
                                                   test_start=test_start,
                                                   test_end=test_end)
-
     # Use label smoothing
-    assert Y_train.shape[1] == 10
-    label_smooth = .1
-    Y_train = Y_train.clip(label_smooth / 9., 1. - label_smooth)
+    print(X_train.shape)
```
goodfeli: Debugging print

goodfeli:
LGTM once the commented code / print statements are fixed

cleverhans_tutorials/mnist_tutorial_cw.py
aashish-kumar: I commented this line as test_end and test_start are hard coded and depend on the dataset(number of training/test samples) hence making the code specific to MNIST. I am not sure for the purpose of having test_start and test_end. I felt it might be to control number of samples getting used but then its index specific too.

goodfeli: The user can pass in any number for `test_start` and `test_end` so they're not specific to MNIST. The idea is that you can evaluate on just a subset of the test set if you want your evaluation to run fast. I think we use this in the tests.

aashish-kumar: I am keeping the assert as of now. Can you please elaborate on the kind of bugs this assert is planned to detect? Will those be data loading bugs or model evaluation bugs?

I feel we should also keep an assert for test_end < len(test_data) immediately after the data loading.

aashish-kumar:
Resolved conflict with latest commit.
There is an interesting update.
X_train,Y_train have been changed to x_train,y_train. Is this change planned?
I have seen that these variables names are not consistent for all the tests. I have fixed most of them as of now.

npapernot:
Yes that was a recent change, variable names were changed from X_train to x_train in #423

goodfeli:
Thanks!

PR #451, opened 2018-07-07
update tutorials

ysharma1126:


cleverhans_tutorials/mnist_tutorial_ead.py
```diff
@@ -0,0 +1,254 @@
+"""
```
goodfeli: We generally don't add new tutorials for each attack, because this leads to a lot of copy-pasted code and support / maintenance burden. Is there something conceptually unique about how to write code using this particular attack? You could also consider contributing this to the `examples` directory, where there isn't an obligation for us to provide ongoing support. (In general, remember to write to cleverhans-dev@googlegroups.com before starting work on a feature to make sure it will be accepted. New single-attack tutorials are one of the most common contributions we ask to have moved to `examples/`)

README.md
```diff
@@ -88,25 +88,34 @@ To help you get started with the functionalities provided by this library, the
 tutorial covers how to train a MNIST model using TensorFlow,
 craft adversarial examples using the [fast gradient sign method](https://arxiv.org/abs/1412.6572),
 and make the model more robust to adversarial examples using adversarial training.
+* **MNIST with CW** ([code](cleverhans_tutorials/mnist_tutorial_cw.py)): this
+tutorial covers how to train a MNIST model using TensorFlow, and
+craft adversarial examples using the [C&W attack](https://arxiv.org/abs/1608.04644).
+* **MNIST with EAD** ([code](cleverhans_tutorials/mnist_tutorial_ead.py)): this
```
goodfeli: As commented below, we probably won't want this to be in the `tutorials` directory, so after resolving that issue we'll need to update this text.

goodfeli:
Looks mostly good. There is one typo to fix, and we probably want the new tutorial to go in the `examples` directory. Otherwise fine.

ysharma1126:
HI Ian,

First of all, sorry for not first going to the Google Group, I'll make sure
to do so next time.

The reason why I added the EAD tutorial is that recently we've found that
at the cost of a minimal increase in distortion, EAD can be configured to
as computationally efficient as attacks such as BIM and PGD. Due to the
minimal increase, we would rather keep the default params to reflect the
optimized config that was presented in the introductory paper, but in the
tutorial we can show that EAD is not destined to be much slower than BIM or
PGD, given the right configuration it can be made as fast with only a
minimal increase in distortion.

As tutorials have already been uploaded for FGSM, JSMA, and CW, attacks
which are all called similarly (due to the consistent Cleverhans API), I
thought this was reason enough to include a tutorial for EAD. I hope that's
ok.

Let me know.

goodfeli:
I don't actually remember why we have a special tutorial for CW or JSMA. Note that we don't have BIM or PGD tutorials. FGSM was just included as a tutorial because we want the tutorial to run fast (10s/epoch for adv training) so that people can run the tutorial at about the same speed they read it. @npapernot do you remember why we have specialized tutorials for CW and JSMA? In any case, I'd prefer not to have a proliferation of more tutorials. It's fine to put them in the `examples` directory where there's no obligation for the rest of the CleverHans team to provide ongoing support for them.

npapernot:
I can remember two reasons for adding a tutorial for the JSMA but not sure if it's exhaustive: (1) demonstrate how to run the source-target misclassification attack (i.e., run for all possible targets based on each sample' source class) and (2) demonstrate how to use the adversarial example visualization utilities. Again from what I can recall, that's why the JSMA tutorial does not include adversarial training (unlike the FGSM tutorial): it was focusing on demonstrating other aspects of the library.

I don't remember why we added the tutorial for the CW but my guess would be that at that time we had a different API that was not as homogeneous across the different attacks?

npapernot:
I would also vote against a proliferation of attack tutorials if they are using a consistent API

ysharma1126:
Understood. Just to clarify the CW tutorial does both (1) and (2) cited
here, making it, at least with the current API, identical to the JSMA
tutorial. Given that the CW tutorial is in effect identical to the JSMA
tutorial, EAD is quite similar (can be reduced to CW with beta=0, and has
been empirically shown to improve attack strength with non-zero beta), and
at least I feel this EAD hyperparameter setting is important for users to
know as it allows EAD/CW to retain its attack strength while being as
computationally efficient as iterative FGSM, I would like to instead just
add EAD as a option in the CW tutorial.

This will require no additional support/maintenance burden as there will be
no additional tutorial. Hope this at least is ok!

goodfeli:
What are the big differences between the tensorflow tutorial with FGSM and
the CW tutorial? Are the APIs still different for those attacks?

ysharma1126:
In regards to the attack, only the fact that FGSM uses the 'generate()'
function while CW uses the 'generate_np()' function, which is a pretty big
difference. EAD shares this difference with FGSM and essentially all of the
other attacks, and this difference isn't reflected in the docs (nor in the
Arxiv paper). For CW, this is only reflected in the tutorial. Another
reason why I would like to include EAD in the CW tutorial.

Also, the differences outlined by Nicolas for JSMA, which are independent
of actually calling the attack.

ysharma1126:
Any thoughts? CW is called differently than FGSM/BIM/PGD, and EAD shares
that property. Users realize that that's the case when they check out the
tutorial for CW, but there's no indication in the repo that this property
is shared by EAD. Plus, again, I'd like to share that hyperparameter config
which gives good results while staying as computationally efficient as
PGD/BIM. So by adding EAD to the CW tutorial, those issues would be
resolved, with no additional support/maintenance required.

goodfeli:
Do you mind starting a discussion on cleverhans-dev so more people will weigh in?

ysharma1126:
Relatively busy currently, so for now I'll pass.

goodfeli:
OK, I'll close the PR for now then. Can re-open it if we make a plan for the new features later.

PR #457, opened 2018-07-15
JSMA_bug_fix

aashish-kumar:
I think there is a bug in jsma_symbolic. The reduction of constant from already-used features should be dependent on the action on pixels(increase/decrease). Please have a look.


npapernot:
Is the behavior you are trying to implement in the bug fix not already implemented through `increase_coef`? See https://github.com/tensorflow/cleverhans/pull/457/files#diff-7c744e6cc2d9648195be4db1c389e9b8R495

aashish-kumar:
I missed that. The increase_coef is handling it.

PR #459, opened 2018-07-16
 Allow omitted nb_classes parameter in KerasModelWrapper()

giuscri:
a36dae6c4d7abf2750871d5884f683243c1e3ce5 introduced the need to pass a `nb_classes` parameter to KerasModelWrapper constructor.

This is not used by KerasModelWrapper instances. More importantly, this breaks code that used KerasModelWrapper before that commit.

### DIFF
--- a/cleverhans/model.py
+++ b/cleverhans/model.py
@@ -13,7 +13,7 @@ class Model(object):
     __metaclass__ = ABCMeta
     O_LOGITS, O_PROBS, O_FEATURES = 'logits probs features'.split()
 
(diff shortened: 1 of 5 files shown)


giuscri:
> On 18 Jul 2018, at 13:28, Nicolas Papernot <notifications@github.com> wrote:
> Also what do you mean by this breaks code that used KerasModelWrapper before?

As nb_classes becomes a new required positional argument of KerasModelWrapper you have to update your code if you use that class.

Also, that argument does't seem to be used in KerasModelWrapper code.

PR #461, opened 2018-07-17
Important bug fix for attacks_tf.py

yang-song:


yang-song:
I signed it!

npapernot:
Thanks for the PR. Could you be a bit more specific on what the bug is and when it occurs? Also @juesato could you take a look since you wrote most of the code that this PR would impact.

yang-song:
I think the problem is that in the method `_compute_gradients` of class `UnrolledOptimizer`, argument `x` is assumed to be a *list* while the same argument `x` of function `loss_fn` is assumed to be a *Tensor*. This doesn't type check, and will definitely cause problems when using some subclasses of `UnrolledOptimizer`, such as `UnrolledAdam`. For example, consider line 1746-1747.

In the class `SPSAAdam`, there is already a similar fix for this problem

```Python
        # Assumes `x` is a list, containing a [1, H, W, C] image
        assert len(x) == 1 and x[0].get_shape().as_list()[0] == 1
        x = x[0]
``` 
Therefore, we should do something for `UnrolledOptimizer`, even though I am not sure my fix is optimal.

yang-song:
@goodfeli I have just added an assert to check input format. Without this fix there will be an exception raised around line 1742

goodfeli:
Would you mind adding a small test that would cause the exception to be triggered if the fix weren't there? This bug indicates that we don't have good enough test coverage.

goodfeli:
Awesome, thanks.

yang-song:
@goodfeli @npapernot I have fixed another bug in `UnrolledGradientDecent` and added tests for `pgd_attack()`.

List of major changes in the code:
1. In `attacks_tf.py`, I implemented method `init_state()` for class `UnrolledGradientDescent`, which is required for using the `UnrolledGradientDescent` optimizer in `pgd_attack()`
2. As discussed previously, I fixed the bug in `_compute_gradients()` of class `UnrolledOptimizer`
3. In `test_attacks_tf.py`, I added tests for `pgd_attack()`. I considered cases of targeted / untargeted attacks, as well as using `UnrolledAdam` or `UnrolledGradientDescent` as the optimizer.

yang-song:
Any updates?

goodfeli:
Thanks!

Issue #455, opened 2018-07-13
Undefined name 'self' in madry_cifar10_model.py


cclauss:
flake8 testing of https://github.com/tensorflow/cleverhans on Python 2.7.14

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./examples/madry_lab_challenges/cifar10/madry_cifar10_model.py:267:31: F821 undefined name 'self'
    return tf.nn.xw_plus_b(x, self.w, self.b)
                              ^
./examples/madry_lab_challenges/cifar10/madry_cifar10_model.py:267:39: F821 undefined name 'self'
    return tf.nn.xw_plus_b(x, self.w, self.b)
                                      ^
2       F821 undefined name 'self'
2
```

goodfeli:
@david-berthelot could this be related to the recent changes to `Model`?

david-berthelot:
Yes I think so, looks like set_input_shape is not called. I can look into it.

david-berthelot:
Looked at it, it's working from the latest master as far as  can tell.

cclauss:
Are you sure that line 267 is getting executed?

david-berthelot:
I am not familiar with that example, I checked if some recent refactoring broke it, but running it I see no problem, it runs properly (using the latest master branch). Did you try running:
`python cifar10/attack_model.py`

PR #462, opened 2018-07-17
Travis CI: Python syntax errors or undefined names

cclauss:
Use [flake8](http://flake8.pycqa.org) to look for Python syntax errors or undefined names on both Python 2 and Python 3.

Python 2 tests should pass when #455 is fixed.

__E901,E999,F821,F822,F823__ are the "showstopper" flake8 issues that can halt the runtime with a SyntaxError, NameError, etc.  Most other flake8 issues are merely "style violations" -- useful for readability but they do not effect runtime safety.  This PR therefore recommends a flake8 run of these tests on the entire codebase.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable `name` referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@
 sudo: required
 dist: trusty
 language: python
+# RUN_FLAKE8 should be set once for each version of Python under test
 matrix:
     include:
         - python: 2.7
@@ -12,6 +13,7 @@ matrix:
           env:
             - KERAS_BACKEND=tensorflow
             - TENSORFLOW_V=1.8.0
+            - RUN_FLAKE8=true
         - python: 3.5
           env:
             - KERAS_BACKEND=tensorflow
@@ -20,6 +22,7 @@ matrix:
           env:
             - KERAS_BACKEND=tensorflow
             - TENSORFLOW_V=1.8.0
+            - RUN_FLAKE8=true
 install:
   # code below is taken from http://conda.pydata.org/docs/travis.html
   # We do this conditionally because it saves us some downloading if the
@@ -63,6 +66,14 @@ install:
   # install dependencies for adversarial competition eval infra tests
   - pip install google-cloud
   - pip install Pillow
+
+
+before_script:
+    # stop the build if there are Python syntax errors or undefined names
+    if [[ "$RUN_FLAKE8" ]]; then
+      pip install flake8;
+      flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics;
+    fi
 
 # command to run tests
 script:
--- a/examples/multigpu_advtrain/utils_cifar.py
+++ b/examples/multigpu_advtrain/utils_cifar.py
@@ -31,6 +31,7 @@ import cPickle as pkl
 
 import numpy as np
 import tensorflow as tf
+from six.moves import xrange
 
 # Global constants describing the CIFAR-10 data set.
 IMAGE_HEIGHT = 32
--- a/examples/multigpu_advtrain/utils_svhn.py
+++ b/examples/multigpu_advtrain/utils_svhn.py
@@ -33,6 +33,7 @@ import os
 import numpy as np
 import tensorflow as tf
 import scipy.io as sio
+from six.moves import xrange
 
 # Global constants describing the SVHN data set.
 IMAGE_HEIGHT = 32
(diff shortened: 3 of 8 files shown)


goodfeli:
Should re-run the tests after merging https://github.com/tensorflow/cleverhans/pull/465

.travis.yml
```diff
@@ -64,6 +64,13 @@ install:
   - pip install google-cloud
   - pip install Pillow
 
+
+before_script:
+    # stop the build if there are Python syntax errors or undefined names
+    if [[ "$TENSORFLOW_V" == "1.8.0" ]]; then
```
goodfeli: is this intended to make sure this test only runs on one build? It currently runs on 2/4 builds

goodfeli: is there another way to trigger this test? I'm concerned that if we change the versions of tensorflow we use, this will silently stop running and we won't notice

goodfeli:
Do you know how this compares to the tests we're already running in devtools/tests/test_format? Do we just need to change a flag in there?

.travis.yml
cclauss: The very first line of the commit message says that we want to run flake8 on __both Python 2 and Python 3__.  This is because Abstract Syntax Trees are different for the two and so they give different results.  See the [very important note at the top of the flake8 docs](http://flake8.pycqa.org/en/latest).

* __print "hi"__ is a syntax error in Python 3 but is valid Python 2.
* __async def my_func():__ is a syntax error in Python to but is valid Python 3

Flake8 knows how to find and flag them both.

cclauss: If you have a better solution, I am all ears.  I did not want flake8 to run 4 times but I did want it to run on both Py2 and Py3.  I figured that 1.8.0 was the latest version under test so it would change last.

cclauss:
I know that the other tests did not find what flake8 found.  I see no problem with running multiple tests if they run efficiently and they discover different issues.

cclauss:
The __devtools/tests/test_format__ approach is leveraging the pycodestyle module.  Flake8 is a superset of pycodestyle + PyFlakes + McCabe complexity analysis plus a few extra tests of its own.   The __--select=E901,E999,F821,F822,F823__ turns off the pycodestyle stuff as I described in the commit message.  If you want to turn the screws down tight (Django does this...) just remove the __--select__ option and then flake8 will be a full replacement for the devtools/tests/test_format but will be even more comprehensive.

goodfeli:
"I see no problem with running multiple tests if they run efficiently and they discover different issues." : if we have a good understanding of what both tests are doing and decide that the right solution is to run both, then sure. What I want to do is make sure that we understand all the tests are doing and that we make conscious decisions based on knowledge of the design, rather than throwing more tests at the wall until one fails. If there's a bug in an existing feature, we should probably understand what the bug is rather than ignore it and paper over the bug with new features.

.travis.yml
goodfeli: What about making an explicit `RUN_FLAKE8` variable and setting it in the `matrix` section, with a comment that it should be on for exactly one job each for Python 2 and Python 3? Long term maintainers will read comments but probably won't read commit messages, especially considering that the original commit may not show up in the blame layer after a while.

cclauss:
I agree 100%...  The commit message starts with a link to the full docs of flake8 and the five specific errors that this PR is meant to address are each described thereafter.

.travis.yml
goodfeli: It's probably worth adding a comment like
` # This installs a version of flake8 specific to the default python interpreter version`
or to change the install command to be explicit about which version to install, so that long term maintains of this file can see more clearly what's going on.

goodfeli:
Yes, you have documented what the new feature does, but an important part of the design process for the PR is to think through how that new behavior relates to the existing codebase.

It looks like the new tests are catching an instance of `F821`. Are we sure that it's intended behavior for the pre-existing tests to not catch `F821`?

cclauss:
Perhaps [this](https://travis-ci.org/cclauss/flake8_for_tensorflow) and [this](https://github.com/tensorflow/models/pulls?q=is%3Apr+author%3Acclauss+is%3Aclosed) will help to explain.  If this project had tests in place to detect undefined names then they would already be fixed.  If they were already fixed then this repo would be green on the former and this PR would not exist.

cclauss:
Were you able to find existing tests that were looking for _undefined names_?  I was not.

PR #463, opened 2018-07-19
PGD-L1 and PGD-L2

ysharma1126:


npapernot:
Thanks for the PR. If we are going to add this logic to the PGD attacks, wouldn't it be easier to call `fgm` directly from `attacks.py` in the `attack_single_step` method of class `MadryEtAl`? Unless I missed something, the logic implemented is the same. This way we reduce duplicated code.

goodfeli:
Yes, especially since this kind of code has frequently had bugs in the past, we shouldn't duplicate it.

ysharma1126:
Made the update.

npapernot:
PEP8 is failing on the latest commit: 

```
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:1380:1: E302 expected 2 blank lines, found 1
/home/travis/build/tensorflow/cleverhans/cleverhans/attacks.py:1483:5: E301 expected 1 blank line, found 0
```

npapernot:
Thanks for changing the code to reuse the `FastGradientMethod`. This helps reduce code duplication. However, I just realized that now `BasicIterativeMethod` and `MadryEtAl` are identical to the exception that `MadryEtAl` has a `rand_init` argument (which makes sense). 

I'm sorry I had not thought about that but I think it would make sense in that case to also consolidate these two classes into one. To make the change backward-compatible, would you be able to create a new class called `ProjectedGradientDescent` that implements what is called `MadryEtAl` at the moment and then have `BasicIterativeMethod` and `MadryEtAl` return an object instantiated from the `ProjectedGradientDescent` class? Ideally, the `ProjectedGradientDescent` would have as the `rand_init` argument not a boolean but a floating point value which is used as the abs value of minval and maxval in the call to `tf.random_uniform`. This way we can also support the attack presented in Florian's work on transferability at the same time.

ysharma1126:
At the moment I am pretty busy, but I welcome any other contributors to
make the change!

npapernot:
Closing because of #508

PR #470, opened 2018-07-23
Updated APIs used in CIFAR-10 attack

jiayunhan:
Adapt train() api to be compatible with the latest utils_tf version, so that the script is runnable again.


goodfeli:
This fails the PEP8 tests

jiayunhan:
PEP8 checker error fixed for ex_cifar10_tf. py

PR #472, opened 2018-07-25
Handle deprecated messages caused by tf.nn.softmax_cross_entropy_with_logits

WindQAQ:
#397 
Use `tf.nn.softmax_cross_entropy_with_logits_v2` instead of `tf.nn.softmax_cross_entropy_with_logits`, which causes deprecated warnings (e.g. when running `cleverhans_tutorials/mnist_tutorial_tf.py`) for Tensorflow 1.8.

### DIFF
--- a/cleverhans/compat.py
+++ b/cleverhans/compat.py
@@ -100,9 +100,9 @@ def softmax_cross_entropy_with_logits(sentinel=None, labels=None, logits=None, d
         raise ValueError("Both labels and logits must be provided.")
 
     try:
-        y = tf.stop_gradient(y)
+        labels = tf.stop_gradient(labels)
         loss = tf.nn.softmax_cross_entropy_with_logits_v2(
-            labels=y, logits=logits, dim=dim)
+            labels=labels, logits=logits, dim=dim)
     except AttributeError:
         warning = "Running on tensorflow version " + \
             LooseVersion(tf.__version__).vstring + \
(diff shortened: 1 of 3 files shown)


cleverhans/compat.py
```diff
@@ -23,9 +23,9 @@ def reduce_function(op_func, input_tensor, axis=None, keepdims=None,
 
     if LooseVersion(tf.__version__) < LooseVersion('1.8.0'):
         warning = "Running on tensorflow version " + \
-                   LooseVersion(tf.__version__).vstring + \
-                   ". This version will not be supported by CleverHans" + \
-                   "in the future."
+            LooseVersion(tf.__version__).vstring + \
+            ". This version will not be supported by CleverHans" + \
+            "in the future."
```
goodfeli: Can we change this to "Support for this version in CleverHans is deprecated and may be removed on or after 2019-01-26"?

cleverhans/compat.py
```diff
@@ -85,3 +85,35 @@ def reduce_any(input_tensor, axis=None, keepdims=None,
...
+        raise ValueError("Both labels and logits must be provided.")
+
+    try:
+        labels = tf.stop_gradient(labels)
+        loss = tf.nn.softmax_cross_entropy_with_logits_v2(
+            labels=labels, logits=logits, dim=dim)
+    except AttributeError:
+        warning = "Running on tensorflow version " + \
+            LooseVersion(tf.__version__).vstring + \
+            ". This version will not be supported by CleverHans" + \
```
goodfeli: same message here as above

goodfeli:
Thanks! This looks good. Let's just put a specific date on the deprecation warning.

WindQAQ:
Already added the date into warnings. Please check it!

PR #478, opened 2018-07-27
Fix missing shape problems in generate for models that use py_func.

### LINKED ISSUE #471: CarliniWagner, DeepFool, ElasticNetMethod, LBFGS Do not work with ModelBasicCNN
The `ModelBasicCNN` defined in `cleverhans_tutorials.tutorial_models` does not work with the following attacks: CarliniWagner, DeepFool, ElasticNetMethod and LBFGS.
I don't know exactly what the reason is, but it seems that the output tensor of generate function cannot infer the shape of the input.

david-berthelot:
Fixes #471


goodfeli:
@nottombrown remember to review this (I haven't used pytorch so I'm not qualified)

PR #480, opened 2018-07-31
Moving a snippet of code to a better place

aashish-kumar:
As part of creating graph, None arguments are removed. This works as expected, but when generate_np is called in a loop, graph will be constructed only during first iteration. So further calls to generate_np can contain None arguments(as they were removed in function construct_graph) which result in following code to throw error:

`        for name in feedable:
             feed_dict[new_kwargs[name]] = feedable[name]`

None arguments should be removed for every call to generate_np so that a single class instantiation could be used.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -100,11 +100,6 @@ class Attack(object):
         class_name = str(self.__class__).split(".")[-1][:-2]
         _logger.info("Constructing new graph for attack " + class_name)
 
-        # remove the None arguments, they are just left blank
-        for k in list(feedable.keys()):
-            if feedable[k] is None:
-                del feedable[k]
-
         # process all of the rest and create placeholders for them
         new_kwargs = dict(x for x in fixed.items())
         for name, value in feedable.items():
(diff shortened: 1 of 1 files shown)


aashish-kumar:
@goodfeli I think there is some problem with the build setup. Can you please check?

goodfeli:
Looks like the Travis build doesn't specify a version for `google-cloud`. I think this project got updated with an API-breaking change recently. I'm trying to fix this here: https://github.com/tensorflow/cleverhans/pull/482

goodfeli:
The build is fixed now. I restarted the build for this PR.

goodfeli:
@carlini can you review this one?

aashish-kumar:
@carlini 
The problem is with calling `generate_np `multiple times in a loop. For the first iteration the `construct_graph ` will be called, it will not be called after that because of `hask_key`. All the other `generate_np `calls will have `kwargs` with some unused arguments too, which will not be deleted(deletion is part of `construct_graph`). These arguments throw errors while creating the feed_dict. I had proposed a remedy for that but I think removing it from construct_graph is not safe. 

I agree with you that removing unused arguments will cause errors when we call `construct_graph` from different paths where the deletion is not done. An alternative could be, to check for unused arguments while creating `feed_dict ` in this line. 
`         for name in feedable:
             feed_dict[new_kwargs[name]] = feedable[name]`

Also some paths could be using the unused arguments as control variable, so deletion in construct_variables is not safe.

carlini:
Yeah, I understand the error case.

It seems like putting the deletion in `construct_graph` is where we want it, though, because not everyone uses `generate_np`. For example, if you use `generate` it still needs to delete the extra arguments.

aashish-kumar:
I added a snippet in generate_np to fix the error without any modification to construct_graph. This will have minimum implications and still fixes the error.

carlini:
Sounds good.

PR #483, opened 2018-08-01
Fix label smoothing in tutorials

WindQAQ:
`y_train` in both `cleverhans_tutorials/mnist_tutorial_tf.py` and `cleverhans_tutorials/mnist_tutorial_keras_tf.py` has been label smoothed twice.

One is [here](https://github.com/tensorflow/cleverhans/blob/master/cleverhans_tutorials/mnist_tutorial_keras_tf.py#L82), and another occurs when `y_train` passes into `LossCrossEntropy` , which is invoked in `train()`, see [this part](https://github.com/tensorflow/cleverhans/blob/master/cleverhans_tutorials/mnist_tutorial_keras_tf.py#L133) for more details.


goodfeli:
Thanks!

PR #484, opened 2018-08-01
Add PyTorch as extra and add PyTorch CI jobs to build matrix

### LINKED ISSUE #452: PyTorch in setup.py
I think the torch tutorial requires pytorch but it is not part of setup.py.

redshiftzero:
Fixes #452.

This PR:
 - Adds `PyTorch` based dependencies as an [extra](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) in `setup.py`
 - Adds PyTorch-based CI jobs (note: I made some changes to the test running such that total CI time is kept under 1 hour)

In the future, if users wish to install with PyTorch extra dependencies, they can do: 

```
pip install -e . cleverhans[pytorch]
```


nottombrown:
Fantastic. This is the pythonic way of including an optional dependency. Merged

PR #485, opened 2018-08-02
Handle deprecated warning from tf.reduce_*

WindQAQ:
Handle deprecated warning from `tf.reduce_*` in `BasicIterativeMethod` and `MomentumIterativeMethod`.

By the way, this should have already been handled in c02b0d1, but appears again after 3a362785.


goodfeli:
Thanks!

PR #489, opened 2018-08-03
Migrate from old and deprecated API to the new one

WindQAQ:
As shown in #423, APIs like `model_train` and `model_loss` in `utils_tf.py` are deprecated. Change them to `train` and `attack_softmax_cross_entropy` instead.


goodfeli:
Thanks!

Issue #495, opened 2018-08-15
SPSA attack is not tested by Travis


nottombrown:
SPSA has some nice tests:
https://github.com/tensorflow/cleverhans/blob/2ebda1ddab0cc3084316b0db7516c8b8e2cc796a/examples/test_imagenet_attacks.py

However, since they live in the `examples` directory, rather than in `cleverhans` or `tests_tf`, they don't get executed by the [travis.yml manifest](https://github.com/tensorflow/cleverhans/blob/master/.travis.yml#L77)

Is this intentional @juesato ?

nottombrown:
Looking a bit more deeply, it seems like it's not there because it is too expensive to run on Inception on a travis CPU. Could we perhaps test it on MNIST or a toy dataset?

goodfeli:
I would vote to test it on a small subset of MNIST (e.g. one small minibatch).

nottombrown:
Yeah, the small test ended up being quite easy to add.

#498 adds tests using the same task as the other attacks in `attacks.py`. Closing this because this seems to now be covered sufficiently.

PR #497, opened 2018-08-15
Cleanup fixes in imagenet attack tests

catherio:
Tidying and fixes in the imagenet attack tests.

This is partial progress to addressing the issue here: 
https://github.com/tensorflow/cleverhans/issues/495

but does not fully fix the issue.


nottombrown:
It looks like the CLA check is mad because both catherio and I are authors on this PR. I consent to the CLA

npapernot:
The CLA error is OK to disregard here (as you mentioned, it was thrown because Tom pushed a commit to Catherine's PR)

OK to merge when Travis tests pass

PR #498, opened 2018-08-16
Check that image pixels are within [0.0, 1.0] bounds at the beginning of PGD attack

nottombrown:
Our PGD implementation requires that inputs are bounded to [0.0, 1.0]

We had a bug in data preprocessing that resulted in an input that was slightly out of bounds, this was hard to debug initially.

These assertions should help people track down the errors faster.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -1733,6 +1733,21 @@ class SPSA(Attack):
 
     def __init__(self, model, back='tf', sess=None, dtypestr='float32'):
         super(SPSA, self).__init__(model, back, sess, dtypestr)
+
+        self.feedable_kwargs = {
+            'epsilon': self.np_dtype,
+            'y': self.np_dtype,
+            'y_target': self.np_dtype,
+        }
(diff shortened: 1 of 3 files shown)


npapernot:
Why does the implementation need the input to be bounded between 0 and 1?

nottombrown:
It currently crops the output to be between 0 and 1, and then it checks that the perturbation is less than epsilon. 

If the initial input is less than 0 or more than 1, then it will always error out with a confusing assertion error.

npapernot:
Tom and I discussed this offline and agreed to move the check within the `project_perturbation` helper before another PR addresses consolidation of the BIM, Madry and pgd_attack code snippets

nottombrown:
Also adds a test that addresses #495

goodfeli:
It seems strange to have a hardcoded range from 0.0 to 1.0. For other attacks, we have `clip_min` and `clip_max` parameters, and it's also possible to leave the clipping turned off. Is there a reason we can't have this attack match the others?

PR #499, opened 2018-08-16
Rename spsa_samples from confusing "batch_size"

nottombrown:
This is independent from the size of the batch coming in (which must be one for SPSA)


npapernot:
Since this change breaks the API, would you be able to add a 6-month deprecation warning first?

nottombrown:
Ah, good idea. Updated

npapernot:
OK to merge when tests pass

PR #503, opened 2018-08-22
Fix sphinx docs TF issue

nottombrown:
This requirement lets sphinx build our docs correctly


nottombrown:
Merging this in because tests pass and it is a tiny change.

PR #507, opened 2018-08-24
Picklable Model framework

goodfeli:


cleverhans/picklable_model.py
```diff
@@ -0,0 +1,372 @@
...
+            assert x is not None
+            out[layer.name] = x
+        return out
+
+
+class Layer(PicklableModel):
+    def __init__(self, name=None, parent=None):
+        if name is not None:
+            self.name = name
+        self.parent = None
```
npapernot: Was this meant to be `self.parent = parent`?

cleverhans/picklable_model.py
```diff
@@ -0,0 +1,372 @@
...
+
+    def get_params(self):
+        return [self.W.var, self.b.var]
+
+
+class Conv2D(Layer):
+
+    def __init__(self, output_channels, kernel_shape, strides, padding,
+                 use_bias=False,
+                 init_scale=1.,
```
npapernot: Are those unintentional or intentional extra line breaks?

cleverhans/picklable_model.py
```diff
@@ -0,0 +1,372 @@
...
+
+class Linear(Layer):
+
+    def __init__(self, num_hid, init_scale=1., init_b=0., **kwargs):
+        super(Linear, self).__init__(**kwargs)
+        self.num_hid = num_hid
+        self.init_scale = init_scale
+        self.init_b = init_b
+
+    def set_input_shape(self, input_shape):
```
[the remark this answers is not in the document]
goodfeli: It's true that I do most of the initialization inside this method for most of the layers I've implemented, but I don't think we need to make that a requirement of the interface. I suspect that when we start wrapping models written by other people we'll find that they do almost all the initialization inside the `fprop` method and that the model just exists in an invalid state until `fprop` is called. That's allowed by the current interface (`get_params` will raise an exception if you call it before `fprop` on models that use that strategy, but that's allowed).

cleverhans/picklable_model.py
```diff
@@ -0,0 +1,370 @@
...
+        assert len(kernel_shape) == 4
+        assert all(isinstance(e, int) for e in kernel_shape), kernel_shape
+        init = tf.random_normal(kernel_shape, dtype=tf.float32)
+        init = self.init_scale * init / tf.sqrt(1e-7 +
+                                                tf.reduce_sum(tf.square(init),
+                                                              axis=(0, 1, 2)))
+        self.kernels = PV(init)
+        if self.use_bias:
+            self.b = PV(np.zeros((self.output_channels,)).astype('float32'))
+        orig_input_batch_size = input_shape[0]
```
npapernot: Where is this variable used?

cleverhans/picklable_model.py
```diff
@@ -0,0 +1,370 @@
...
+        input_shape = list(input_shape)
+        input_shape[0] = 1
+        dummy_batch = tf.zeros(input_shape)
+        dummy_output = self.fprop(dummy_batch)
+        output_shape = [int(e) for e in dummy_output.get_shape()]
+        output_shape[0] = 1
```
npapernot: Was this meant to be: `output_shape[0] = orig_input_batch_size`?

cleverhans/picklable_model.py
goodfeli: looks like it's not used anymore. deleted it

cleverhans/picklable_model.py
```diff
@@ -0,0 +1,370 @@
...
+        return self.output_shape
+
+    def fprop(self, x):
+        return tf.nn.tanh(x)
+
+    def get_params(self):
+        return []
+
+
+class LeakyReLU(Layer):
```
npapernot: Why are there two implementations of leaky ReLU? (this one and through `leak=0.` in ReLU class)?

cleverhans/picklable_model.py
goodfeli: Yes, thanks. Fixed.

cleverhans/picklable_model.py
```diff
@@ -0,0 +1,370 @@
...
+        return self.output_shape
+
+    def fprop(self, x):
+        return tf.nn.elu(x)
+
+    def get_params(self):
+        return []
+
+
+class SELU(Layer):
```
npapernot: I guess a few of these layers that just apply an activation function could inherit from the same class to save a few repeated lines but I'm not sure how much that would help/hurt readability.

cleverhans/picklable_model.py
goodfeli: I changed LeakyReLU to have nothing but a constructor supplying a default leak amount

npapernot: Thanks!

cleverhans/picklable_model.py
goodfeli: They're short enough that I mostly prefer to have duplicated code so that it's easy to change them later. For example, we might want to add a feature where you can learn the leak parameter of LeakyReLU, and then its get_params won't be an empty list anymore. We might want to add a feature where Softmax takes n - 1 inputs instead of n and then it's ouptut_shape isn't equal to its input_shape anymore, etc.

cleverhans/picklable_model.py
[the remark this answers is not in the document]
npapernot: I see, I guess this would be kind of like the few lines that you added here: https://github.com/tensorflow/cleverhans/pull/507/files#diff-597d409a81b3230e149314553adbd67bR24

npapernot:
LGTM, OK to merge when tests pass

PR #508, opened 2018-08-24
consolidate both attacks based on PGD

npapernot:
This replaces #463 by removing the MadryEtAl class that duplicated the BasicIterativeMethod. 

Instead, both classes now inherit from a ProjectedGradientDescent class with a `rand_minmax` used to control the noise added before the gradient descent and switch from one attack to the other.

This also allows using L1 and L2 distances w/ MadryEtAl, which was previously only possible w/ BasicIterativeMethod.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -10,6 +10,7 @@ from distutils.version import LooseVersion
 from cleverhans.compat import reduce_sum, reduce_mean
 from cleverhans.compat import reduce_max, reduce_min
 from cleverhans.compat import reduce_any
+from cleverhans.utils_tf import clip_eta
 
 _logger = utils.create_logger("cleverhans.attacks")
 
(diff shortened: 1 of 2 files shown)


cleverhans/attacks.py
```diff
@@ -353,23 +354,27 @@ def parse_params(self,
...
 
 
-class BasicIterativeMethod(Attack):
+class ProjectedGradientDescent(Attack):
     """
-    The Basic Iterative Method (Kurakin et al. 2016). The original paper used
+    Class implements either the Basic Iterative Method (Kurakin et al. 2016)
+    when rand_minmax is set to 0. or the Madry et al. (2017) method when
+    rand_minmax is larger than 0. The original Kurakin et al. (2016) paper used
     hard labels for this attack; no label smoothing.
```
goodfeli: Do we know what's going on with the comment about label smoothing?

cleverhans/attacks.py
```diff
@@ -505,6 +521,20 @@ def parse_params(self,
         return True
 
 
+class BasicIterativeMethod(ProjectedGradientDescent):
+    def __init__(self, model, back='tf', sess=None, dtypestr='float32'):
+        super(BasicIterativeMethod, self).__init__(model, back, sess=sess,
+                                                   dtypestr=dtypestr)
+        self.child = 'BasicIterativeMethod'
```
goodfeli: why do we need this self.child variable?

cleverhans/attacks.py
```diff
@@ -505,6 +521,20 @@ def parse_params(self,
         return True
 
 
+class BasicIterativeMethod(ProjectedGradientDescent):
```
goodfeli: do we actually need different subclasses? The only difference is the value of one parameter, right?

cleverhans/attacks.py
```diff
@@ -485,6 +494,12 @@ def parse_params(self,
         :param clip_min: (optional float) Minimum input component value
         :param clip_max: (optional float) Maximum input component value
         """
+        if self.child is not None:
+            if self.child == 'BasicIterativeMethod':
+                rand_minmax = 0.
+            elif self.child == 'MadryEtAl':
+                if 'rand_init' in kwargs and not kwargs['rand_init']:
+                    rand_minmax = 0.
```
goodfeli: I would suggest the following. Change this line to

    rand_minmax = self.default_rand_minmax()

Then in `BasicIterativeMethod` implement

    def default_rand_minmax(self):
        return 0.

and in the Madry subclass implement

     def default_rand_minmax(self):
          if self.rand_init:
              return self.epsilon
           return 0.

(if epsilon will already have been parsed? maybe epsilon needs to be an argument to this method)

cleverhans/attacks.py
goodfeli: talked with Nicolas offline. It makes sense to keep the different subclasses so that there is one class per published paper and for backwards compatibility.

cleverhans/attacks.py
goodfeli: my comment above shows how to get rid of this using inheritance / different implementations of the default_rand_minmax method

goodfeli:
@npapernot what's the status of this PR?

npapernot:
Been sidetracked by ICLR deadline, will work on it this Friday afternoon

goodfeli:
I will finish this PR. Closing this so I can open another copy of it from my own fork

PR #515, opened 2018-08-29
Add BasicIterativeMethod in eager execution

WindQAQ:
This PR adds basic iterative method in Tensorflow eager mode, and mainly modifies the following two parts:

1. Gradient tape in [this line](https://github.com/tensorflow/cleverhans/compare/master...WindQAQ:master#diff-1e01328cea7defb5113accf896de8718R146) should watch at `x` so that the `tape.gradient(loss, x)` can be calculated. Original `FastGradientMethodTFE` works without any error because the input `x` is created by `tfe.Variable` ([here](https://github.com/tensorflow/cleverhans/blob/master/cleverhans/utils_tfe.py#L70)), which is automatically watched by gradient tape ([ref](https://www.tensorflow.org/api_docs/python/tf/GradientTape)).

2. Add `BasicIterativeMethodTFE` and add it into available attacks in `cleverhans_tutorial/mnist_tutorial_tfe.py`.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -431,11 +431,14 @@ class BasicIterativeMethod(Attack):
         }
 
         if not tf.executing_eagerly():
-            FGM = FastGradientMethod(self.model, back=self.back, sess=self.sess, dtypestr=self.dtypestr)
+            FGM = FastGradientMethod(
+                self.model, back=self.back,
+                sess=self.sess, dtypestr=self.dtypestr)
         else:
             # eager execution
             import cleverhans.attacks_tfe as attacks_tfe
-            FGM = attacks_tfe.FastGradientMethod(self.model, dtypestr=self.dtypestr)
+            FGM = attacks_tfe.FastGradientMethod(
+                self.model, dtypestr=self.dtypestr)
 
         def cond(i, _):
             return tf.less(i, self.nb_iter)
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -430,7 +430,12 @@ class BasicIterativeMethod(Attack):
             'clip_max': self.clip_max
         }
 
-        if not tf.executing_eagerly():
+        try:
+            eager_executing = tf.executing_eagerly()
+        except:
+            eager_executing = False
+
+        if eager_executing:
             FGM = FastGradientMethod(
                 self.model, back=self.back,
                 sess=self.sess, dtypestr=self.dtypestr)
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -49,6 +49,7 @@ class Attack(object):
         self.model = model
         self.back = back
         self.sess = sess
+        self.dtypestr = dtypestr
 
         # We are going to keep track of old graphs and cache them.
         self.graphs = {}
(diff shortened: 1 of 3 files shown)


goodfeli:
I'm not very familiar with tf Eager, but is it really necessary to have totally parallel versions of every class to work with Eager?
@david-berthelot @nottombrown Do you know if there's a cleaner way of supporting Eager?

Even if it is, I don't think we need TFE in the name of every class. Just having the classes live in attacks_tfe should be enough.

WindQAQ:
It seems that we can use `tfe.py_func()` to selectively enable eager execution, but in this way, the computational graph will also exist.
See [Eager](https://www.tensorflow.org/guide/eager) for more details.

cleverhans/attacks_tfe.py
```diff
@@ -189,3 +194,100 @@ def fgm(self, x, labels, targeted=False):
...
+        y_kwarg = 'y_target' if targeted else 'y'
+        fgm_params = {
+            'eps': self.eps_iter,
+            y_kwarg: y,
+            'ord': self.ord,
+            'clip_min': self.clip_min,
+            'clip_max': self.clip_max
+        }
+
+        FGM = FastGradientMethodTFE(self.model, dtypestr=self.dtypestr)
```
david-berthelot: If t he only reason why you need to define this method entirely is to be able to call FastGradientMethodTFE, you could refactor the original class as follows.

Declare a class constant: FGM = FastGradientMethodTFE
And in the method, just use fgm = self.FGM(...) to create an instance of it.

cleverhans/attacks_tfe.py
```diff
@@ -189,3 +194,100 @@ def fgm(self, x, labels, targeted=False):
...
+        def body(i, e):
+            adv_x = FGM.generate(x + e, **fgm_params)
+
+            # Clipping perturbation according to clip_min and clip_max
+            if self.clip_min is not None and self.clip_max is not None:
+                adv_x = tf.clip_by_value(adv_x, self.clip_min, self.clip_max)
+
+            # Clipping perturbation eta to self.ord norm ball
+            eta = adv_x - x
+            from cleverhans.utils_tf import clip_eta
```
david-berthelot: Any particular reason for using a local import?

cleverhans/attacks_tfe.py
```diff
@@ -189,3 +194,100 @@ def fgm(self, x, labels, targeted=False):
         if (self.clip_min is not None) and (self.clip_max is not None):
             adv_x = tf.clip_by_value(adv_x, self.clip_min, self.clip_max)
         return adv_x
+
+
+class BasicIterativeMethodTFE(AttackTFE, BasicIterativeMethod):
```
david-berthelot: I agree with Ian that it would be preferable not to postfix every name with FTE and possibly use import mechanisms instead to separate different targets (like TF and TFE and ...).

cleverhans/attacks.py
```diff
@@ -429,7 +430,20 @@ def generate(self, x, **kwargs):
             'clip_max': self.clip_max
         }
 
-        FGM = FastGradientMethod(self.model, back=self.back, sess=self.sess)
```
david-berthelot: I think it's better not to modify this class to test for TFE, my original idea was just to define the fast gradient method to use as a constant attached to class, for example:
```
class BasicIterativeMethod(Attack):
    FGM_CLASS = FastGradientMethod

    ...
    def generate(self, x, **kwargs):
        ...
        FGM = self.FGM_CLASS(self.model, back=self.back, sess=self.sess)
        ...
 ```

This way in TFE when you create your new TFE specific class you overwrite the constant with the proper TFE FGM class.
```
# In attacks_tfe.py
class FastGradientMethod(Attack, attacks.FastGradientMethod):
    FGM_CLASS = FastGradientMethod  # It will pick the one local to your file.
```

This way you don't have to rewrite the method for generate and the code changes are minimal.

WindQAQ: Got it!

PR #516, opened 2018-08-29
add Add Layer

goodfeli:


cleverhans/picklable_model.py
```diff
@@ -354,3 +354,76 @@ def fprop(self, x):
...
+                layer.name = name
+            self.layer_names.append(name)
+
+    def get_params(self):
+        out = []
+        for layer in self.layers:
+            out = ordered_union(out, layer.get_params())
+        return out
+
+    def fprop(self, x, set_ref=False):
```
npapernot: Could you document was `set_ref` is for? I missed that in the `MLP` class as well. I assumed it is used to save a reference to the Python object for serialization w/ Pickle but correct me if I'm wrong.

cleverhans/picklable_model.py
```diff
@@ -354,3 +354,76 @@ def fprop(self, x):
 
     def get_params(self):
         return []
+
+
+class Add(Layer):
+    """
+    A Layer that adds a function to its input.
+    The function to add is specified in terms of multiple layers, just like
+    in the MLP class.
```
npapernot: This comment mentions multiple layers but the one in `fprop` says only one parent is allowed. Could you explain why?

cleverhans/picklable_model.py
goodfeli: This was actually for reference batch norm / virtual batch norm, which we don't really use anymore. I just took it out.

cleverhans/picklable_model.py
goodfeli: This comment means the Add layer itself contains several layers. The output of those internal layers is added to the one layer that is the parent of the Add layer. For example, to use a single resnet layer, you use an Add(layers=[Conv2D, ReLU, Conv2D]).

npapernot: Got it, thanks for clarifying. I had missed that the comment applied to layers in `self.layers` rather than the parent they are being added to.

PR #518, opened 2018-08-30
Dataset class

goodfeli:
This Dataset class is intended to pave the way for generic evaluation scripts. A single script can be pointed at a pickled model. The script can then read a pickled Factory out of the model and reconstruct the same dataset that was used to train the model. This way we can write one script that evaluates a wide range of models on a wide range of datasets, rather than needing to write many scripts that are hard-coded to each (model, dataset) combination.

### DIFF
--- a/cleverhans/dataset.py
+++ b/cleverhans/dataset.py
@@ -16,7 +16,7 @@ class Dataset(object):
     """
 
     # The number of classes in the dataset. Should be specified by subclasses.
-    nb_classes = None
+    NB_CLASSES = None
 
     def get_factory(self):
         """Returns a picklable callable that recreates the dataset.
@@ -38,7 +38,7 @@ class Dataset(object):
 class MNIST(Dataset):
     """The MNIST dataset"""
 
-    nb_classes = 10
+    NB_CLASSES = 10
 
     def __init__(self, train_start=0, train_end=60000, test_start=0,
                  test_end=10000, center=False):
@@ -62,7 +62,7 @@ class MNIST(Dataset):
 class CIFAR10(Dataset):
     """The CIFAR-10 dataset"""
 
-    nb_classes = 10
+    NB_CLASSES = 10
 
     def __init__(self, train_start=0, train_end=60000, test_start=0,
                  test_end=10000, center=False):
--- a/cleverhans/dataset.py
+++ b/cleverhans/dataset.py
@@ -56,6 +56,29 @@ class MNIST(Dataset):
         self.x_test = x_test
         self.y_test = y_test
 
+class CIFAR10(Dataset):
+  """The CIFAR-10 dataset"""
+
+  nb_classes = 10
+
+    def __init__(self, train_start=0, train_end=60000, test_start=0,
+                 test_end=10000, center=False):
+        self.kwargs = locals()
+        del self.kwargs["self"]
+        x_train, y_train, x_test, y_test = data_cifar10(train_start=train_start,
+                                                      train_end=train_end,
+                                                      test_start=test_start,
+                                                      test_end=test_end)
+
+        if center:
+            x_train = x_train * 2. - 1.
+            x_test = x_test * 2. - 1.
+
+        self.x_train = x_train
+        self.y_train = y_train
+        self.x_test = x_test
+        self.y_test = y_test
+
 
 class Factory(object):
     """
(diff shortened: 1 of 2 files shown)


cleverhans/dataset.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+
+    # The number of classes in the dataset. Should be specified by subclasses.
+    nb_classes = None
+
+    def get_factory(self):
+        """Returns a picklable callable that recreates the dataset.
+        """
+
+        if hasattr(self, 'args'):
```
david-berthelot: I would suggest not supporting args, since positional arguments can lead to errors (if the order changed in the class). On the other hand, kwargs are perfect since even if the order changes or new elements are added, it will still work fine.

cleverhans/dataset.py
```diff
@@ -0,0 +1,146 @@
...
+            args = self.args
+        else:
+            args = []
+
+        if hasattr(self, 'kwargs'):
```
david-berthelot: No need to test, simply make a class constructor such that kwargs are always required.

```
def __init__(self, kwargs):
  if 'self' in kwargs:
   del kwargs['self']
  self.kwargs = kwargs
```

cleverhans/dataset.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+class MNIST(Dataset):
+    """The MNIST dataset"""
+
+    nb_classes = 10
+
+    def __init__(self, train_start=0, train_end=60000, test_start=0,
+                 test_end=10000, center=False):
+        self.kwargs = locals()
```
david-berthelot: Here you could call  the Dataset constructor
```
Dataset.__init__(self, locals())
```

This way you don't have to copy the `del` statement for every dataset. You can also use the `super` statement if you prefer.

cleverhans/dataset.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+class Factory(object):
+    """
+    A callable that creates an object of the specified type and configuration.
+    """
+
+    def __init__(self, cls, args, kwargs):
+        self.cls = cls
+        self.args = args
```
david-berthelot: Same I suggest dropping args since I think it's error prone.

cleverhans/dataset.py
```diff
@@ -0,0 +1,146 @@
...
+        from keras.datasets import cifar10
+        from keras.utils import np_utils
+
+    # These values are specific to CIFAR10
+    img_rows = 32
+    img_cols = 32
+    nb_classes = 10
+
+    # the data, shuffled and split between train and test sets
+    (X_train, y_train), (X_test, y_test) = cifar10.load_data()
```
david-berthelot: Naming (PEP8), X_* should be replaced with x_*.

cleverhans/dataset.py
goodfeli: Mostly I was thinking that some classes might not take any arguments at all. For these classes, it's slightly more convenient if the child class doesn't need to set self.kwargs = {} or call super(Subclass, self).__init__ to make Dataset.__init__ set self.kwargs = {}.

cleverhans/dataset.py
```diff
@@ -0,0 +1,146 @@
...
+
+keras = None  # Only load keras if user tries to use a dataset that requires it
+
+
+class Dataset(object):
+    """Abstract base class representing a dataset.
+    """
+
+    # The number of classes in the dataset. Should be specified by subclasses.
+    nb_classes = None
```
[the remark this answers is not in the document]
goodfeli: Thanks, done

cleverhans/dataset.py
goodfeli: Thanks, done

cleverhans/dataset.py
goodfeli: Thanks, done

cleverhans/dataset.py
goodfeli: Thanks, done

cleverhans/dataset.py
goodfeli: Thanks, done

cleverhans/dataset.py
```diff
@@ -0,0 +1,144 @@
...
+        self.x_test = x_test
+        self.y_test = y_test
+
+
+class Factory(object):
+    """
+    A callable that creates an object of the specified type and configuration.
+    """
+
+    def __init__(self, cls, args, kwargs):
```
david-berthelot: Note, you can remove args here since it's unused.

cleverhans/dataset.py
```diff
@@ -0,0 +1,144 @@
...
+    def __init__(self, **kwargs):
+        if "self" in kwargs:
+            del kwargs[self]
+        self.kwargs = kwargs
+
+    def get_factory(self):
+        """Returns a picklable callable that recreates the dataset.
+        """
+
+        if hasattr(self, 'kwargs'):
```
david-berthelot: Isn't this statement always true? If it was false it would be because the children would not have invoked the parent constructor which would be unexpected behavior. You could do `assert(hasattr(self, 'kwargs'))` to ensure the constructor was properly called.

cleverhans/dataset.py
```diff
@@ -0,0 +1,144 @@
...
+
+class MNIST(Dataset):
+    """The MNIST dataset"""
+
+    NB_CLASSES = 10
+
+    def __init__(self, train_start=0, train_end=60000, test_start=0,
+                 test_end=10000, center=False):
+        super(MNIST, self).__init__(**locals())
+        self.kwargs = locals()
```
david-berthelot: This statement is not longer necessary since you moved the logic to the constructor.

cleverhans/dataset.py
david-berthelot: Also following up on your comment if the child class has no parameters, locals() will only have self and everything will work as expected.

cleverhans/dataset.py
goodfeli: thanks, done

cleverhans/dataset.py
goodfeli: ok, I took out the check

cleverhans/dataset.py
goodfeli: thanks, done

PR #529, opened 2018-09-04
use mean value for loss

goodfeli:
Using the mean value rather than a sum makes the loss invariant to batch size.
Invariance to batch size is nice if we want to search over hyperparameters or if we want to do synchronous multi-GPU training, since the size of the batch received by a single GPU is inverse proportional to the number of GPUs.

### DIFF
--- a/cleverhans/loss.py
+++ b/cleverhans/loss.py
@@ -4,6 +4,7 @@ import os
 from .model import Model
 from .compat import softmax_cross_entropy_with_logits
 import tensorflow as tf
+import warnings
 
 
 class Loss(object):
@@ -55,8 +56,7 @@ class WeightedSum(Loss):
 
         return tf.add_n(terms)
 
-
-class LossCrossEntropy(Loss):
+class CrossEntropy(Loss):
     def __init__(self, model, smoothing=0., attack=None, **kwargs):
         """Constructor.
         :param model: Model instance, the model on which to apply the loss.
@@ -90,7 +90,7 @@ class LossCrossEntropy(Loss):
         return loss
 
 
-class LossMixUp(Loss):
+class MixUp(Loss):
     def __init__(self, model, beta, **kwargs):
         """Constructor.
         :param model: Model instance, the model on which to apply the loss.
@@ -106,11 +106,12 @@ class LossMixUp(Loss):
         xm = x + mix * (x[::-1] - x)
         ym = y + mix * (y[::-1] - y)
         logits = self.model.get_logits(xm, **kwargs)
-        loss = softmax_cross_entropy_with_logits(labels=ym, logits=logits)
+        loss = tf.reduce_mean(softmax_cross_entropy_with_logits(labels=ym,
+            logits=logits))
         return loss
 
 
-class LossFeaturePairing(Loss):
+class FeaturePairing(Loss):
     def __init__(self, model, weight, attack, **kwargs):
         """Constructor.
         :param model: Model instance, the model on which to apply the loss.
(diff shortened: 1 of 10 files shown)


cleverhans/loss.py
```diff
@@ -66,7 +66,8 @@ def fprop(self, x, y, **kwargs):
 
         logits = [self.model.get_logits(x, **kwargs) for x in x]
         loss = sum(
-            softmax_cross_entropy_with_logits(labels=y, logits=logit)
+            tf.reduce_mean(softmax_cross_entropy_with_logits(labels=y,
```
npapernot: Should we add a warning here alerting users of the change in behavior?

goodfeli: I think the warning would be annoying, because even people who know about and want the new behavior would always see the warning.

Maybe one option would be to leave the existing class untouched, make a new class with the new behavior, and put a deprecation warning in the old class. e.g., put a deprecation warning in LossCrossEntropy, and make a new class called CrossEntropy with the new behavior. Are you OK with that?

npapernot: Agreed on the warning being potentially annoying. Your proposal sounds great, thanks!

goodfeli:
I realized we shouldn't merge this yet because I ought to update the tutorials to not get deprecation warnings everywhere. Will do so shortly

npapernot:
Thanks!

goodfeli:
This needed a rebase after we merged WeightDecay, so the review history might get confusing. I'm about to do a push -f

goodfeli:
OK, I fixed the tutorials so it is ready to review again.

Note that I had to increase the *attack* learning rate to make the CW tutorial pass. That is a little bit strange; I would have expected that Adam is scale invariant enough that I would need to make no change, or that I would need to change the *training* learning rate.

cleverhans_tutorials/mnist_tutorial_cw.py
```diff
@@ -25,11 +25,16 @@
 
 FLAGS = flags.FLAGS
 
+LEARNING_RATE = .001
+CW_LEARNING_RATE = .2
+ATTACK_ITERATIONS = 100
```
npapernot: This is a nitpick but could you create flags for these (under `if __name__ == '__main__':`) so we maintain consistency with the rest of the hyper-parameters and tutorials?

npapernot:
It is strange yes, I would also have expected that the training learning rate would be impacted. It may be the case that the previous attack learning rate had not been tuned for this particular model given that the tutorial model was changed recently?

goodfeli:
I think the most likely explanation is that the trained model changed slightly, due to the different scale of the loss. Adam is partly but not completely invariant to rescaling of the loss. Under this hypothesis, the model as configured for the test ended up being just slightly more robust to adversarial examples (it was only one example that failed).

I don't think there is a good way to keep the model completely unchanged. For example, we can't just scale up the learning rate by a factor of `batch_size` because Adam is already using the second moments of the gradient to compensate for that scaling factor.

cleverhans_tutorials/mnist_tutorial_cw.py
goodfeli: There are flags for them down below. I'm defining constants up here so that we can refer to the same constant when defining the default value for the python function and when defining the default value for the flags. Currently the defaults are hard-coded in two separate locations, and it can be confusing. While working on making the test pass, I accidentally changed the default for the python function but didn't change the default for the flag, and it took me a while to notice that my changes weren't having any effect.

npapernot: SGTM, that's a good catch. I'll make an issue so we don't forget to standardize this across the remaining tutorials

PR #545, opened 2018-09-07
Dropout

goodfeli:


cleverhans/picklable_model.py
```diff
@@ -496,5 +506,36 @@ def set_input_shape(self, shape):
...
+
+    def fprop(self, x, dropout_dict=None, **kwargs):
+        """
+        Forward propagation as either no-op or dropping random units.
+        :param x: The input to the layer
+        :param dropout_dict: dict mapping layer names to dropout inclusion
+            probabilities.
+            This dictionary should be passed as a named argument to the MLP
+            class, which will then pass it to *all* layers' fprop methods.
+            Other layers will just recieve this as an ignored kwargs entry.
```
npapernot: Typo

cleverhans/picklable_model.py
```diff
@@ -496,5 +506,36 @@ def set_input_shape(self, shape):
...
+        self.output_shape = shape
+
+    def get_params(self):
+        return []
+
+    def fprop(self, x, dropout_dict=None, **kwargs):
+        """
+        Forward propagation as either no-op or dropping random units.
+        :param x: The input to the layer
+        :param dropout_dict: dict mapping layer names to dropout inclusion
```
npapernot: I am curious what is the reason for using a dictionary common to the model rather than defining the dropout inclusion probabilities at each dropout layer individually?

goodfeli: I think i will rewrite it to define the inclusion probability for each layer in that layer's constructor, and then use fprop's kwargs only to say whether to drop or not drop. Does that sound good to you?

npapernot: Yes, that sounds easier to keep track of. Thanks!

cleverhans/picklable_model.py
goodfeli: Fixed, thanks

cleverhans/picklable_model.py
goodfeli: Done

goodfeli:
I left the `dropout_dict` feature in so users can still change the dropout include_prob on each fprop if they need to (e.g. if it helps to answer a research question). Most users can ignore it and just use the `dropout` bool though.

goodfeli:
A pytorch test failed. It seems unlikely that this PR affects the pytorch tests, so I think it was just a flaky test. I restarted that worker.

goodfeli:
All tests pass, ready to re-review

cleverhans/picklable_model.py
```diff
@@ -496,5 +509,48 @@ def set_input_shape(self, shape):
...
+            dropout include probability.
+            This dictionary should be passed as a named argument to the MLP
+            class, which will then pass it to *all* layers' fprop methods.
+            Other layers will just receive this as an ignored kwargs entry.
+            Each dropout layer looks up its own name in this dictionary
+            to read out its include probability.
+        """
+        include_prob = self.include_prob
+        if dropout_dict is not None:
+            assert dropout
```
npapernot: Just making sure I understand the intended use here. Does that mean that all layers must have `dropout` set to `True` even if their `include_prob` is 0 when `dropout_dict` is not `None`?  Does that also mean that an include probability must be specified for all dropout layers in the `dropout_dict`?

goodfeli: The default use case is that you never specify `include_prob` anywhere. During evaluation, you don't do anything special regarding dropout, and nothing gets dropped. During training, you pass "dropout=True" to make units get randomly dropped. If you've done nothing else, include_prob defaults to 0.5 in the Dropout class constructor.

A slightly more advanced use case is that you want to use different include_probs for some layer. For example, people usually use include_prob=0.8 for the input layer. To do this, you specify include_prob in the constructor arguments for those layers. Other than that, it's the same as the basic use case case. You do nothing special at test time and nothing is dropped. You pass dropout=True at train time and units are dropped based on their input prob.

The most advanced use case is if you want to change dropout include probabilities for a specific fprop call. In this case, you pass dropout=True and dropout_dict for that call. Each layer uses the include_prob specified in the dropout_dict for that call.

npapernot: My comment was mostly wondering about 2. and is more clear now, thanks for making the change! Also there was a typo in my comment, I meant when `include_prob` is 1. (basically disable dropout for one layer only), sorry about that.

npapernot:
LGTM

PR #546, opened 2018-09-07
Layers needed for making a resnet

goodfeli:


cleverhans/picklable_model.py
```diff
@@ -575,3 +575,114 @@ def fprop(self, x, dropout=False, dropout_dict=None, **kwargs):
         if dropout:
             return tf.nn.dropout(x, include_prob)
         return x
+
+
+class Residual(Layer):
```
npapernot: I guess there can be different residual layers as suggested by the comment below. Would it make sense to name this `ResidualWithGroupNormalization` in case we want to implement other residual network layers or does that sound too long?

cleverhans/picklable_model.py
```diff
@@ -575,3 +575,114 @@ def fprop(self, x, dropout=False, dropout_dict=None, **kwargs):
...
+    """A residual network layer that uses group normalization."""
+
+    def __init__(self, out_filter, stride, activate_before_residual=False):
+        self.__dict__.update(locals())
+        del self.self
+        self.lrelu = LeakyReLU(0.1)
```
npapernot: Would it make sense to let users control the leaky relu parameter through an arg to the constructor of `Residual`?

npapernot:
I restarted the Travis build in hope that it will successfully show up as passing here this time so we can merge the PR

PR #548, opened 2018-09-09
better error message

goodfeli:


cleverhans/loss.py
```diff
@@ -47,12 +47,26 @@ def fprop(self, x, y):
...
     def __init__(self, model, terms):
         self.terms = terms
+
         Loss.__init__(self, model, locals())
 
     def fprop(self, x, y, **kwargs):
-        terms = [weight * loss.fprop(x, y, **kwargs)
-                 for weight, loss in self.terms]
-        assert all(len(term.get_shape()) == 0 for term in terms)
+        weights = [weight for weight, loss_object in self.terms]
```
AlexeyKurakin: FIY, assuming that self.terms is a list of tuples you can do following:
  weights, losses = zip(*self.terms)
instead of following two lines:
  weights = [weight for weight, loss_object in self.terms]
  losses = [loss.fprop(x, y, **kwargs) for loss in loss_objects]

goodfeli: Done. I also added a `safe_zip` function that checks that the lengths match, to avoid bugs where one list is short.

goodfeli:
Something is wrong with Travis. All the individual builds are done and green, but the icon here is still yellow. I'll go ahead and merge because the tests actually did pass. I won't try to debug why the in-PR icon is yellow unless it starts affecting other PRs.

PR #551, opened 2018-09-09
multi-device evaluation

goodfeli:


cleverhans/evaluation.py
```diff
@@ -0,0 +1,338 @@
...
+        depending on how work is distributed to devices.
+        This factory must respect "with tf.device()" context managers
+        that are active when it is called, otherwise work will not be
+        distributed to devices correctly.
+    :param numpy_inputs:
+        A list of numpy arrays defining the dataset to be evaluated.
+        The list should have the same length as tf_inputs.
+        Each array should have the same number of examples (shape[0]).
+        Example: numpy_inputs is [MNIST().x_test, MNIST().y_test]
+    :batch_size: Number of examples to use in a single evaluation batch.
```
david-berthelot: Should it be :param batch_size: ?

cleverhans/evaluation.py
```diff
@@ -0,0 +1,338 @@
...
+        be divided up evenly among available devices. If you can fit 128
+        examples in memory on one GPU and you have 8 GPUs, you probably
+        want to use a batch size of 1024 (unless a different batch size
+        runs faster with the ops you are using, etc.)
+    :param devices: List of devices to run on. If unspecified, uses all
+        available GPUs if any GPUS are available, otherwise uses CPUs.
+    :param feed: An optional dictionary that is appended to the feeding
+             dictionary before the session runs. Can be used to feed
+             the learning phase of a Keras model for instance.
+    """
```
david-berthelot: :returns: List of numpy arrays corresponding the outputs prodced by the graph_factory?

cleverhans/evaluation.py
```diff
@@ -0,0 +1,338 @@
...
+        replicated_tf_inputs, replicated_tf_outputs = packed
+        p = len(replicated_tf_outputs[0])
+        assert p > 0
+    else:
+        # This graph has not been built before.
+        # Build it now.
+
+        for device in devices:
+            with tf.device(device):
+                tf_inputs, tf_outputs = graph_factory()
```
david-berthelot: There's a subtlety that's escaping me here. Why do the the placeholders have to be created per GPU?
Generally I've seen GPU parallelizing done with
```
# For simplicity, the example is single input/single output.
# data_input = placeholder created on CPU or anywhere for that matter.
# model_func = function that build output from input
data_input_split = split(data_input, num_gpus)
tf_outputs = []
for i, device in enumerate(devices):
    with tf.device(device):
        tf_outputs.append(model_func(data_input_split[i]))
tf_outputs = tf.concat(tf_outputs, axis=0)
```

cleverhans/evaluation.py
```diff
@@ -0,0 +1,338 @@
...
+    n = len(numpy_inputs)
+    assert n > 0
+    assert n == len(tf_inputs)
+    m = numpy_inputs[0].shape[0]
+    for i in range(1, n):
+        assert numpy_inputs[i].shape[0] == m
+    out = []
+    for _ in tf_outputs:
+        out.append([])
+    with sess.as_default():
```
david-berthelot: I don't think you need this statement.

cleverhans/evaluation.py
goodfeli: I just tried switching to a single tf_inputs created on CPU and benchmarked it against the version in the PR. I get that the CPU version is about 12% slower.

I think what's going on is that the GPU version involves one copy (from numpy input array in main memory to the GPU memory buffer for the placeholder) and the CPU version involves two copies (CPU memory-to-memory copy from numpy to the tf placeholder, then a second copy from tf placeholder to GPU for the first op on GPU to use as input).

Also, I think the placeholders are not the greatest and we should eventually transition to queues, like in this example: https://github.com/tensorflow/models/blob/master/tutorials/image/cifar10/cifar10_multi_gpu_train.py
I'm not actually sure if queues vs placeholders have a big effect on runtime when the data is stored in main memory like this. We might also be able to get a speedup for smaller models by storing the data on GPU (e.g. just have a Variable holding 1/n of the dataset on each of n GPUs, and then take a slice of the Variable fore each minibatch).

cleverhans/evaluation.py
goodfeli: Thanks, I removed it.

cleverhans/evaluation.py
goodfeli: Thanks, I added this to the docstring.

cleverhans/evaluation.py
goodfeli: Yes, thanks. I fixed it.

goodfeli:
Thanks, ready for more review

cleverhans/evaluation.py
david-berthelot: Thanks, I learned something new :)

PR #567, opened 2018-09-16
Add some canary code, intended to fail if there is a GPU driver problem, etc.

goodfeli:


cleverhans/train.py
```diff
@@ -104,78 +106,122 @@ def train(sess, loss, x_train, y_train,
...
                 x = x_batch_preprocessor(x)
+
+            # We need to keep track of these so that the canary can feed
+            # preprocessed values. If the canary had to feed raw values,
+            # stochastic preprocessing could make the canary fail.
+            preprocessed_xs.append(x)
+
+            loss.model.get_params()
             loss_value = loss.fprop(x, y, **fprop_args)
+            loss.model.get_params()
```
npapernot: Could you clarify why the second call to get_params() is needed here? Is it impacted by fprop?

cleverhans/train.py
```diff
@@ -104,78 +106,122 @@ def train(sess, loss, x_train, y_train,
...
+    # Check whether the hardware is working correctly
+
+    # So far the failure has only been observed with 3 or more GPUs
+    run_canary = num_devices > 2
+    if run_canary:
+        canary_feed_dict = {}
+        for x, y in safe_zip(preprocessed_xs, ys):
+            canary_feed_dict[x] = x_train[:device_batch_size].copy()
+            canary_feed_dict[y] = y_train[:device_batch_size].copy()
+        param_to_test = 0
```
npapernot: Is this set to 0 because the canary only tests the gradient corresponding to the first parameter in the list of gradients returned by the optimizer's `compute_gradient` ?

cleverhans/train.py
goodfeli: That was debugging code that I didn't mean to commit. I've removed it. Actually neither call to get_params is needed. I just had them there for debugging purposes, to make sure the model I was working with didn't add new parameters when I ran it on each new GPU.

cleverhans/train.py
goodfeli: Yes. I added some comments to explain that.

npapernot: Thanks

cleverhans/train.py
npapernot: I understand, thanks for the explanation

PR #569, opened 2018-09-19
Unicode->Ascii Python2 fix for utils_mnist.

ftarlao:
Mnist tutorial was broken with Python2, it invokes array.array method providing a unicode string param, where an ascii string is expected. This is a working fix.

### DIFF
--- a/cleverhans/utils_mnist.py
+++ b/cleverhans/utils_mnist.py
@@ -10,6 +10,7 @@ import operator
 import os
 import struct
 import tempfile
+import sys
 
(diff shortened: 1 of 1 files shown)


ftarlao:
It seems that continuous integration tests fail cause job errors, not related to code.. it seems.
e.g., two jobs have failed with 
"Unable to download 3.5 archive. The archive may not exist. Please consider a different version."
That is a initialization problem, imho this travis CI stuff sucks a lot.
Please check

PR #571, opened 2018-09-19
Add exponentially moving average to training script

goodfeli:


cleverhans/train.py
```diff
@@ -124,6 +129,23 @@ def train(sess, loss, x_train, y_train,
     with tf.control_dependencies(tf.get_collection(tf.GraphKeys.UPDATE_OPS)):
         train_step = optimizer.apply_gradients(grad)
 
+    if use_ema:
+        ema = tf.train.ExponentialMovingAverage(decay=ema_decay)
+        with tf.control_dependencies([train_step]):
+            train_step = ema.apply(var_list)
+        avg_params = [ema.average(param) for param in var_list]
+        tmp_params = [tf.Variable(param, trainable=False)
```
npapernot: Is the following used in place of a custom getter for tf.variable_scope ? Could you add a line or two explaining at a high level?

cleverhans/train.py
```diff
@@ -227,7 +247,13 @@ def train(sess, loss, x_train, y_train,
         _logger.info("Epoch " + str(epoch) + " took " +
                      str(cur - prev) + " seconds")
         if evaluate is not None:
+            if use_ema:
```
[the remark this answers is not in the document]
goodfeli: I added a comment next to each call saying what that call is for. Let me know whether the comments make sense.

cleverhans/train.py
goodfeli: I added a few lines of comments, let me know whether they make sense / whether they're sufficient

PR #574, opened 2018-09-20
Add missing value for NB_FILTERS in mnist_tutorial_tfe.py

vipinpillai:
NB_FILTERS was missing from the mnist_tutorial_tfe.py file. 
I have added the same value as was present in mnist_tutorial_tf.py.


vipinpillai:
I signed it!

PR #580, opened 2018-09-22
enable logger by default (so that `train` gives a progress indicator)

goodfeli:


npapernot:
Quick comment: it might be nice to set the log level to INFO at the beginning of train and then set it back to what it was before when train returns so that this doesn't affect people beyond their use of train?

goodfeli:
Maybe I'm clashing with a culture of how to use loggers that I'm not aware. If so, let me know.

My thinking about this is based on what I as a user want to happen, and that might be different from a convention I don't know I'm meant to follow.

As a user, I want cleverhans to print out results by default. I don't want to have to to specifically ask to see the results. When I'm tuning a model, one of the most important results is how fast it runs, so I definitely want to see the epoch messages printed out.

Given that I want to see these messages by default, there's a question of how cleverhans should log them. Nothing has gone wrong, so they shouldn't be logged as "warning". This means they can be logged at most as "info". This means to satisfy my preferences as a user, the logger should show "info" messages by default.

If cleverhans code is going to log a lot of stuff the user doesn't want to see by default, that can be done with logging.debug.

In terms of conventions and matching the behavior of other software, the conventions I'm most familiar with are tensorflow / keras, which print quite a lot by default: https://github.com/tensorflow/tensorflow/issues/8340
This means most of my scripts start with a lot of boilerplate telling tensorflow to print less than it prints by default and then also telling cleverhans to print more than it does by default.

npapernot:
My bad, your response actually made me realize that my comment doesn't make sense because the way you have it set up right now gives more control to the user and is what I was also looking for. 

I also just checked the tutorials and they set the logger to debug but it's better to set it to info by default in this file indeed because we changed all of the trainer's prints to logging.

OK to merge once pep8 passes:

```
FAIL: Test if pep8 is respected.
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/tensorflow/cleverhans/cleverhans/devtools/tests/test_format.py", line 45, in test_format_pep8
    raise AssertionError("PEP8 Format not respected")
```

PR #592, opened 2018-09-26
PGD consolidation

goodfeli:


cleverhans/attacks.py
```diff
@@ -354,25 +355,29 @@ def parse_params(self,
...
+    Madry et al. (2017) method when rand_minmax is larger than 0.
+    Paper link (Kurakin et al. 2016): https://arxiv.org/pdf/1607.02533.pdf
+    Paper link (Madry et al. 2017): https://arxiv.org/pdf/1706.06083.pdf
     """
 
     FGM_CLASS = FastGradientMethod
 
-    def __init__(self, model, back='tf', sess=None, dtypestr='float32'):
+    def __init__(self, model, back='tf', sess=None, dtypestr='float32',
+                 rand_init=True):
```
npapernot: We initially had agreed to not add any attack specific parameters to the constructor and include those in `generate` or `generate_np` only. Did you also conclude that it was a lot of overhead to follow the usual convention here?

tests_tf/test_attacks.py
```diff
@@ -865,7 +867,13 @@ def test_multiple_initial_random_step(self):
             I = (orig_labs == new_labs_multi)
             new_labs_multi[I] = new_labs[I]
 
-        self.assertTrue(np.mean(orig_labs == new_labs_multi) < 0.1)
+        self.assertTrue(np.mean(orig_labs == new_labs_multi) < 0.5)
+
+
+class TestProjectedGradientDescent(TestMadryEtAl):
```
npapernot: Could you add a separate test for the MadryEtAl class and the BasicIterativeMethod class? I found that some of the code broke for one but not the other when trying to consolidate the code earlier so I'm worried that changes down the road might introduce bugs we don't notice if we don't have a simple test for both classes in addition to the ProjectedGradientDescent class.

cleverhans/attacks.py
goodfeli: Are the conventions documented somewhere?

I'd prefer to just get rid of generate_np and then not need to do all this weird stuff to support the caching, but that's an issue for another day.

Working off of my understanding of how the cache system works, but working without an understanding of what all the conventions are meant to be, I think that it's safe to put rand_init in the constructor. Every attach instance gets its own cache, so you'll never have the cache get populated by an instance that has rand_init=False and then later get a cache hit from an instance that has rand_init=True. The only way something could go wrong is if someone edited an instance to change its rand_init variable. It's possible for someone to edit that field after construction because Python doesn't have const or privacy, but people generally have a cultural understanding that they shouldn't do it.

I don't understand very well how things would work if we made `rand_init` an argument to `generate`. Why would we have a separate `BasicIterativeMethod` and `MadryEtAl` class if the user still has to pass `rand_init` to `generate`?

npapernot: I wasn't necessarily thinking about the caching issue when I made that comment but more about a design choice we made a while ago when introducing the attacks interface. I looked back in the merged PRs and found at least part of the related conversation we had. From what you can see [here](https://github.com/tensorflow/cleverhans/pull/75#pullrequestreview-34174290), we were concerned that users would have to instantiate multiple objects to run the same attack with different parameters (e.g., eps) if the parameters were parsed in the constructor (which also ties back to what you were saying about the cache in some way). However, here the situation is a little different because a user would most likely not change the value of `rand_init` at each call to `generate`. So I understand your point and I think it's OK in this case to include the `rand_init` parameter in the constructor, especially given that it makes for a cleaner implementation overall. 

To answer your question about moving rand_init to be an argument of generate, my original idea was to let people use PGD with or without the random restart, while at the same time allowing new comers who read the papers for BIM or Madry et al.'s attack to understand the difference between the two approaches and easily find a corresponding implementation in CleverHans (with the right default value for rand_init in each class). Your implementation achieves the same goal by documenting the difference between the two attacks and how that relates to `rand_init` in the documentation of the `ProjectedGradientDescent` class.

I agree that we should also have the discussion about generate_np again. Probably on the dev mailing list so others can pitch in if they have an opinion or use case for keeping generate_np. My opinion is that if we continue to support `utils_tf.batch_eval`, `generate_np` can be deprecated as long as the attacks that cannot be defined symbolically can be wrapped properly w/ `tf.py_func`.

tests_tf/test_attacks.py
goodfeli: Done

cleverhans/attacks.py
goodfeli: OK, I changed it so that:
- `rand_init` is an argument to `generate`. It's default value is `None` so we can tell whether the use passed it in or not.
- If the user doesn't pass it in, it gets filled in with `self.default_rand_init`.
- `default_rand_init` is specified in the constructor.
- Different subclasses pass different values to `default_rand_init`

This way both classes do their own thing, but the user can also turn `rand_init` on or off for each call.

Re getting rid of `generate_np`: maybe we'll need to do that anyway when we move to tf 2.0 / Eager, so maybe the answer is I should spend some time writing Eager code before forming an opinion.

Issue #597, opened 2018-09-27
Enable pylint check for W0613: Unused argument


goodfeli:

goodfeli:
Decided not to enable it, see comment in .pylintrc

PR #621, opened 2018-09-27
Switch to TensorFlow style

goodfeli:
This PR switches to TensorFlow style, i.e. PEP8 with 2 spaces.

Most of this was done automatically with a script.

The script was added in commit "autopep8_all.py (The script to turn the whole library tf style)"

The script was run in commit "ran autoep8_all.py". This makes the PR as a whole have a huge diff, but the diffs for the other individual commits should be OK to read.

The remaining commits are based on removing the old test based on PyCodeStyle and setting up a new test based on pylint. As far as I can tell, PyCodeStyle does not support using 2 spaces for tab.

Good news: pylint is very thorough. It actually imports the python files and catches many errors that PyCodeStyle was not able to with text analysis alone.

Bad news: pylint is very slow.


.pylintrc
```diff
@@ -0,0 +1,22 @@
...
+# pylint should print out the file path, line number, and ID code for the
+# messages it prints. This is partly so we can turn off the ID codes that
+# we do not actually want to enforce.
+msg-template='{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}'
+
+# ID codes to disable
+# C0103: Complains about the variable name 'x'
+# C0411: Import order
+# C0111: Missing docstring
+# W0613: Unused argument
```
npapernot: Are there any arguments not used in the API right now? Or is the attacks interface flagged as having unused arguments because of how we structured it around `parse_params` right now?

cleverhans/utils.py
```diff
@@ -420,3 +420,37 @@ def safe_zip(*args):
         raise ValueError("Lengths of arguments do not match: "
                          + str([len(arg) for arg in args]))
     return zip(*args)
+
+def shell_call(command, **kwargs):
+  """Calls shell command with argument substitution.
+
+  Args:
+    command: command represented as a list. Each element of the list is one
+      token of the comman. For example "cp a b" becomes ['cp', 'a', 'b']
```
npapernot: Typo (missing a d in command)

npapernot: Nevermind, just saw it in the next commit...

.pylintrc
goodfeli: I'm not 100% sure, but I think this warning is only triggered by the tutorials or examples.
I do think W0613 is one of the warnings we should prioritize turning back on though. Hopefully during the hackathon tomorrow.

cleverhans_tutorials/tutorial_models_tfe.py
```diff
@@ -113,11 +113,3 @@ def get_params(self):
   def get_layer_names(self):
     """:return: the list of exposed layers for this model."""
     return self.layer_names
-
-  def call(x):
```
npapernot: Why was this deleted here?

npapernot:
Why was the `examples/linear_extrapolation_plot_example.py` file deleted?

goodfeli:
`examples/linear_extrapolation_plot_example.py` turned out to be broken enough that I didn't think it was worthy fixing, partly because it never had tests. I'll probably just write a properly tested version from scratch sometime soon. I'm planning to start a plotting module within the next week or so.

.pylintrc
```diff
@@ -15,8 +15,48 @@ msg-template='{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}'
...
 # C0111: Missing docstring
 # W0613: Unused argument
-# E1101: Seems to have a bug. Says numpy.random has no member RandomState
-# E0611: Seems to have a bug. Complains about imports of tensorflow
+# E1101: pylint seems to have a bug. Says numpy.random has no member RandomState
+# E0611: pylint seems to have a bug. Complains about imports of tensorflow
 # R0913: Too many arguments
 # R0914: Too many local variables
-disable=C0103,C0411,C0111,W0613,E1101,E0611,R0913,R0914
+# W0611: Unused import
```
npapernot: Why disable this one?

cleverhans_tutorials/tutorial_models_tfe.py
goodfeli: - It's not used
- It's always been broken (no "self") argument
- It's not even the right name (should be "__call__")
- The right thing (__call__) method is implemented in the base class, so this class shouldn't implement it anyway

.pylintrc
goodfeli: Just to keep this PR smaller. We should get it back on soon: https://github.com/tensorflow/cleverhans/issues/598

cleverhans/devtools/tests/test_format.py
```diff
@@ -51,8 +84,12 @@ def print_files_information_pep8():
   list of files which do not respect PEP8 formatting that aren't in the
   whitelist
   """
+
+  raise NotImplementedError("Broken by move to TensorFlow style")
+
```
npapernot: Why are we keeping what's below if it is broken?

.pylintrc
npapernot: Just saw #597 so it's ok to discard for now

cleverhans_tutorials/tutorial_models_tfe.py
npapernot: SGTM

.pylintrc
npapernot: SGTM

npapernot:
Regarding plotting, I agree that a new module is worth it. We can also refactor existing plotting tools in `utils.py`

cleverhans/devtools/tests/test_format.py
goodfeli: A lot of it will be useful for implementing the pylint version

goodfeli:
Thanks!

PR #635, opened 2018-09-28
deprecate back argument in attacks

### LINKED ISSUE #631: Deprecate backend argument to attacks class

npapernot:
Closes #631 

(WIP)

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -44,7 +44,8 @@ class Attack(object):
 
     if sess is None:
       sess = tf.get_default_session()
-    assert isinstance(sess, tf.Session)
+    if not isinstance(sess, tf.Session):
+      raise ValueError("sess is not an instance of tf.Session")
 
     import cleverhans.attacks_tf as attacks_tf
     attacks_tf.np_dtype = self.np_dtype
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -1,8 +1,9 @@
 from abc import ABCMeta
+import collections
 import numpy as np
 from six.moves import xrange
+import tensorflow as tf
 import warnings
-import collections
 
 import cleverhans.utils as utils
 from cleverhans.model import Model, CallableModelWrapper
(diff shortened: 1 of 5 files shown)


cleverhans/attacks.py
```diff
@@ -21,22 +21,30 @@ class Attack(object):
...
+      if kwargs['back'] == 'tf':
+        warnings.warn("Argument back to attack constructors is not needed"
+                      " anymore and will be removed on or after 2019-03-26."
+                      " All attacks are implemented using TensorFlow.")
+      else:
+        raise ValueError("Backend argument must be 'tf' and is now deprecated"
+                         "It will be removed on or after 2019-03-26.")
 
+    self.tf_dtype = tf.as_dtype(dtypestr)
     self.np_dtype = np.dtype(dtypestr)
```
goodfeli: Add a check on the type of `sess`. This will also make sure that an old positional argument passed for `back` doesn't end up going to `sess` now that `back` has been removed.

goodfeli:
OK to merge after adding the check on type of `sess`

cleverhans/attacks.py
```diff
@@ -47,7 +57,7 @@ def __init__(self, model, back='tf', sess=None, dtypestr='float32'):
 
     # Prepare attributes
     self.model = model
-    self.back = back
+    self.back = 'tf'
```
goodfeli: Do we still need self.back at all?

goodfeli: If we need self.back for backwards compatibility purposes, maybe we should make it a python property ( https://www.programiz.com/python-programming/property ) that issues a deprecation warning.

tests_tf/test_attacks.py
```diff
@@ -1055,21 +1054,21 @@ def test_no_transformation(self):
...
-
-        x_adv_p = self.attack.generate(x, batch_size=100, dx_min=-0.2,
-                                       dx_max=0.2, n_dxs=3, dy_min=-0.2,
-                                       dy_max=0.2, n_dys=3, angle_min=-45,
-                                       angle_max=45, n_angles=3)
-        x_adv = self.sess.run(x_adv_p, {x: x_val})
-        new_labs = np.argmax(self.sess.run(self.model(x_adv)), axis=1)
-        self.assertTrue(np.mean(orig_labs == new_labs) < 0.7)
-        print(np.mean(orig_labs == new_labs) )
+    # def test_attack_strength(self):
```
goodfeli: Why is this test commented out? That looks scary

npapernot:
1. Since self.back is not used, I removed it.
2. I commented this test out because it is not implemented correctly (e.g., it passes a numpy array to a CleverHans model and then runs it in a session). I created #704 to make sure we don't forget to fix the test

goodfeli:
Thanks for removing self.back.

Regarding the test, two further thoughts:
1. I'd suggest just deleting it entirely rather than commenting it out. You could put a link to the commit that removes it in the issue, so that when someone works on the issue they can find the broken implementation quickly. My concern is that if we have a test for attack strength and we comment it out, it looks like the attack got weaker and we thought that was OK. I understand that that's not what happened, but we don't want to even look like we're doing that.
2. If this test wasn't implemented correctly, why wasn't it failing before? It sounds like it should have had some syntax error or raised an exception from passing the wrong type, etc.

npapernot:
Thanks for the feedback. I removed the test and added a pointer to the commit in the issue. I'm not sure why the test was not failing before (it was sometimes giving me a different outcome after running it multiple times).

goodfeli:
LGTM. Feel free to merge once the tests pass

PR #640, opened 2018-09-28
create model zoo

goodfeli:


NullConvergence:
In the past I wanted to develop all primitives that will allow composition of all important architectures. Some initial code is here: https://github.com/NullConvergence/dnnmodels/tree/develop/dnnmodels 
I would gladly develop some models, but do you have any recommendations/guides?

PR #643, opened 2018-09-28
WIP: switch from Travis to CloudBuild (this shouldn't work yet)

goodfeli:
Opening this PR just to see if I'm successfully triggering CloudBuild. All I want right now is for Travis not to run, CloudBuild to run, and CloudBuild to give me an error message because the build file is written for Travis rather than Cloud.


goodfeli:
OK, I think Travis is going to run because of our github settings. Apparently just not having a .travis.yml isn't enough to stop it. I don't want to turn this off for the other branches yet, so I'll just leave travis running and crashing on this branch for now.

goodfeli:
According to this documentation https://cloud.google.com/cloud-build/docs/run-builds-on-github I think there should be 1 check in the checks tab now but there are 0.

PR #644, opened 2018-09-28
add a dummy cloudbuild file

goodfeli:
This is just a copy of our travis config file. I don't intended for it to work. Right now I'm just attempting to get a cloud build to start and fail.


goodfeli:
Merging, speculating that the cloudbuild file has to be in master before the PR is opened for Cloud Build to work

PR #650, opened 2018-09-28
Closes #618 (Enable E1121)

### LINKED ISSUE #618: Enable pylint check E1121 (too many arguments)

goodfeli:


goodfeli:
I fixed the test failures by just removing the broken example. I propose removing it because:
- We shouldn't have duplicated code for the whole cartesian product of different ways of running everything. There's already an MNIST keras tutorial and and MNIST black box tutorial, no need for an MNIST keras black box tutorial.
- This was broken (it's a real error, not just a style problem) and no one has noticed, so it's not getting used
- It's in the examples directory so there's no obligation to keep supporting it

PR #669, opened 2018-09-29
warn user when tutorials are run from diff dir

### LINKED ISSUE #538: Make tutorials check cleverhans.__path__
It's common for users to install two versions of CleverHans and get confused when they run the tutorial from a new version and it fails to load new classes from the old version that takes precedence in their PYTHONPATH. We could cut down on this confusion a bit if we added a check from inside the CleverHans tutorials that the cleverhans module is in the same directory as the tutorial it is imported from.

npapernot:
Closes #538


cleverhans_tutorials/mnist_blackbox.py
```diff
@@ -285,6 +286,13 @@ def mnist_blackbox(train_start=0, train_end=60000, test_start=0,
 
 
 def main(argv=None):
+  # Warn user if running cleverhans from a different directory than tutorial.
+  cur_dir = os.path.split(os.path.dirname(os.path.abspath(__file__)))[0]
+  ch_dir = os.path.split(cleverhans.__path__[0])[0]
+  if cur_dir != ch_dir:
+    warnings.warn("Rnning cleverhans from a different directory than "
```
goodfeli: typo "Rnning"

goodfeli: I'd suggest a longer message with more explanation.

`It appears that you have at least two versions of cleverhans installed, one at %(cur_dir)s and one at %(ch_dir)s. You are running the tutorial script from the former but python imported the library module from the latter. This may cause errors, for example if the tutorial version is newer than the library version and attempts to call new features.`

cleverhans_tutorials/mnist_tutorial_cw.py
```diff
@@ -237,6 +237,13 @@ def mnist_tutorial_cw(train_start=0, train_end=60000, test_start=0,
 
 
 def main(argv=None):
```
goodfeli: Could we factor this into one function that is called by several tutorials?

We could define it in `cleverhans_tutorials/__init__.py` and then just put

`
from cleverhans_tutorials import check_installation
check_installation()
` in each tutorial.

goodfeli:
Done with first review, two suggestions

npapernot:
Those are both good suggestions. I just pushed the changes but had to create a new file `check_installation.py` otherwise both paths would be equal

npapernot:
Agreed, thanks for the suggestion! I pushed the update

goodfeli:
LGTM when tests pass. Unfortunately this needs a rebase already.

PR #696, opened 2018-10-01
Fix a bug in batch_eval

### LINKED ISSUE #572: a problem about mnist_blackbox.py
problem

How should i do?

WindQAQ:
Fix #572 .
The problem is caused by the reason that `batch_eval` in `utils_tf.py` does not return any value.


cleverhans/evaluation.py
```diff
@@ -3,6 +3,8 @@
 Includes multi-GPU support for fast evaluation.
 """
 
+import warnings
```
npapernot: Why was this import added?

WindQAQ: Because module `warnings` is used in `batch_eval`, but it has never been imported. See [here](https://github.com/tensorflow/cleverhans/blob/3b4820ab723db3dae549154064d748695e81805d/cleverhans/evaluation.py#L259) for more details.

npapernot: Thanks for catching that! Could you move the import after six.moves so they are in alphabetical order?

WindQAQ: Sure!

npapernot:
I closed and reopened the PR to trigger a new Travis build.

WindQAQ:
@npapernot Should I do anything for the CLA?

npapernot:
No it just broke because I pushed a minor fix to your branch so we can still merge. Thanks for asking!

Issue #593, opened 2018-09-26
Speed up PerImageStandardize


goodfeli:
Currently, this class is presumably slow, due to both the use of a map_fn and being on CPU:
https://github.com/tensorflow/cleverhans/blob/23e530ee60c826b840526b39bfee9624f1c14af4/cleverhans/picklable_model.py#L553

It should be straightfoward to implement the same thing on GPU, just check that the change doesn't affect model accuracy at all.

Issue #703, opened 2018-10-03
consolidate tests


npapernot:
they are located in two separate directories currently

PR #712, opened 2018-10-04
Speed up PerImageStandardize

WindQAQ:
Address #593 . The implementation is based on [here](https://github.com/tensorflow/tensorflow/blob/r1.11/tensorflow/python/ops/image_ops_impl.py#L1176) (including adjusted stddev).
On single P100 GPU, this is 50x faster than `map_fn` for input shape `(128, 32, 32, 3)`


goodfeli:
Thanks! The speed sounds great. Can you add a unit test that the 2 methods are equivalent?

WindQAQ:
It's done!

cleverhans/compat.py
```diff
@@ -81,6 +81,16 @@ def reduce_mean(input_tensor, axis=None, keepdims=None,
                          reduction_indices=reduction_indices)
 
 
+def reduce_prod(input_tensor, axis=None, keepdims=None,
+                name=None, reduction_indices=None):
+  """
+  Wrapper around the tf.reduce_prod to handle argument keep_dims
+  """
+  return reduce_function(tf.reduce_mean, input_tensor, axis=axis,
```
npapernot: Shouldn't this be ` return reduce_function(tf.reduce_prod, input_tensor, axis=axis,` ?

cleverhans/picklable_model.py
```diff
@@ -562,10 +563,22 @@ def get_params(self):
   def fprop(self, x, **kwargs):
     # TODO: before adding dataset augmentation, we didn't have to do this.
```
npapernot: I believe we can remove the TODO if we removed the with tf.device() call. Is this correct @goodfeli ?

cleverhans/compat.py
npapernot: Is the test sensitive enough if it did not break despite the function not being the right one?

WindQAQ: It seems that the random generated samples will not trigger adjusted stddev, so the test pass even if this is wrong. Maybe I can come up with some dedicated cases for this?

npapernot:
Yes, that would be nice to make sure that future changes to compat.py don't
break the changes made here. Thanks!

npapernot:
Sorry I just realized why the tests were not failing, they were added to a different non-monitored repo. Can you move them to the `tests_tf` folder? You can find an example test in `tests_tf/test_attacks.py` for instance. Sorry again for not noticing this earlier.

WindQAQ:
It's okay. 
I think it's better to create another PR for moving all of them because some of the files in `cleverhans/tests` have the same name with those in `tests_tf` (like test_utils_tf.py), and some of them have different testing styles (class wrapper or method). Therefore, I just put my test in the right place in this PR, and I will create another one to address #703

npapernot:
Tests are failing due to a formatting issue
```
cleverhans/dataset.py:117: [W0612(unused-variable), maybe_download_file] Unused variable 'urljoin'
```

The last item remaining is to remove the TODO (see comment above). Once this is fixed, we can merge the PR

WindQAQ:
Fixed :smiley:

PR #714, opened 2018-10-04
C0414

goodfeli:
Merge this first: https://github.com/tensorflow/cleverhans/pull/713


goodfeli:
Let's actually just combine PRs, rather than doing multiple rebases. The work in here is in https://github.com/tensorflow/cleverhans/pull/714

PR #720, opened 2018-10-04
W0612

goodfeli:
Merge https://github.com/tensorflow/cleverhans/pull/719 first

PR #721, opened 2018-10-04
E1129

goodfeli:
Merge https://github.com/tensorflow/cleverhans/pull/720 first


npapernot:
Let's wait for #654 to be merged so we don't have to refactor it

.pylintrc
```diff
@@ -24,9 +24,9 @@ msg-template='{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}'
 #        method call.
 # W0201: doesn't let any attributes be added outside __init__
 # C1801: do not use len(s) to tell if s is empty
-# C0123: uniodimatic typecheck
```
npapernot: Should this still be here because it is not removed from disable below?

PR #723, opened 2018-10-04
Initial code for certification of neural network against adversarial examples

AlexeyKurakin:
Originally code written by Aditi Raghunathan

PR #728, opened 2018-10-05
Update travis config in preparation for making certified defense.

AlexeyKurakin:
I have some problems making code for certified defense to be tested by Travis.
So I split the original change https://github.com/tensorflow/cleverhans/pull/723 into travis config and the code itself.

This is the first part of the change which only include updates in travis config


goodfeli:
Yeah, let's drop support for older versions.

AlexeyKurakin:
So do you want to drop support of both TF 1.4 and TF 1.8? and update everything to TF 1.9?
Or you still want to keep support of one of the older versions?

goodfeli:
yeah, let's do 1.8 and 1.9 for now

npapernot:
Just talked to Ian in real life and he agrees on dropping 1.4 and keeping 1.8

PR #729, opened 2018-10-05
Adds `correctness_and_confidence`

goodfeli:
This is a function that runs on multiple devices. For each example it reports whether the model got the example correct and also the probability the model assigned to that example. This runs on clean examples by default or adversarial examples if an attack is passed in.

### DIFF
--- a/cleverhans/evaluation.py
+++ b/cleverhans/evaluation.py
@@ -43,9 +43,46 @@ def accuracy(sess, model, x, y, batch_size=None, devices=None, feed=None):
   return correct.mean()
 
 
-# Cache for storing output of `batch_eval_multi_worker`'s calls to
-# `graph_factory`, to avoid making the tf graph too big
-_batch_eval_multi_worker_cache = {}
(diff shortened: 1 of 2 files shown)


cleverhans/evaluation.py
```diff
@@ -342,4 +379,61 @@ def __call__(self):
...
+                                   in sorted(attack_params.keys()))
+    self.properties_to_hash = (model, attack, hashable_attack_params)
+
+  def __hash__(self):
+    # Make factory hashable so that no two factories for the
+    # same model will be used to build redundant tf graphs
+    return self.properties_to_hash.__hash__()
+
+  def __eq__(self, other):
+    # Make factory hashable so that no two factories for the
```
npapernot: This comment seems to be copied over from `__hash__`?

goodfeli: It is the same comment but it's still valid. An object needs to implement both `__hash__` and `__eq__` to be hashable. The `__hash__` function determines which bucket it lands in and the `__eq__` function makes it possible to tell it apart from other objects in the same bucket.

PR #734, opened 2018-10-05
Added model to the zoo, using for the CIFAR10 tutorial.

david-berthelot:
Support for #732

### DIFF
--- a/cleverhans_tutorials/cifar10_tutorial_tf.py
+++ b/cleverhans_tutorials/cifar10_tutorial_tf.py
@@ -1,0 +1,218 @@
+"""
+This tutorial shows how to generate adversarial examples using FGSM
+and train a model using adversarial training with TensorFlow.
+It is very similar to mnist_tutorial_keras_tf.py, which does the same
+thing but with a dependence on keras.
+The original paper can be found at:
+https://arxiv.org/abs/1412.6572
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import logging
+import numpy as np
+import tensorflow as tf
+from tensorflow.python.platform import flags
+
+from cleverhans.attacks import FastGradientMethod
+from cleverhans.dataset import CIFAR10
+from cleverhans.loss import CrossEntropy
+from cleverhans.model_zoo.all_convolutional import ModelAllConvolutional
+from cleverhans.train import train
+from cleverhans.utils import AccuracyReport, set_log_level
+from cleverhans.utils_tf import model_eval
+
+FLAGS = flags.FLAGS
+
+NB_EPOCHS = 6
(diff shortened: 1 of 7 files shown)


cleverhans/utils_tf.py
```diff
@@ -516,3 +516,17 @@ def get_available_gpus():
   """
   local_device_protos = device_lib.list_local_devices()
   return [x.name for x in local_device_protos if x.device_type == 'GPU']
+
+
```
goodfeli: maybe put this somewhere like a new `cleverhans.initializers`. I think `utils_tf` should be more about just making tensorflow run, rather than implementing specific ML feature in tensorflow.

cleverhans/utils_tf.py
```diff
@@ -516,3 +516,17 @@ def get_available_gpus():
...
+class HeReLuNormalInitializer(tf.initializers.random_normal):
```
goodfeli: This is also in cleverhans_tutorials.tutorial_models, right? We should remove that copy and turn it into an import

cleverhans/utils_tf.py
npapernot: + 1 

Can we also move the initializer from `cleverhans_tutorials/tutorial_models.py`

npapernot: Nevermind Ian wrote that same comment below

cleverhans/utils_tf.py
npapernot: +1

cleverhans_tutorials/cifar10_tutorial_tf.py
```diff
@@ -0,0 +1,219 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import logging
+import numpy as np
+import tensorflow as tf
+from tensorflow.python.platform import flags
+
+from cleverhans.loss import CrossEntropy
```
npapernot: Can you order the imports alphabetically?

cleverhans_tutorials/cifar10_tutorial_tf.py
```diff
@@ -0,0 +1,219 @@
...
+def cifar10_tutorial(train_start=0, train_end=60000, test_start=0,
+                     test_end=10000, nb_epochs=NB_EPOCHS, batch_size=BATCH_SIZE,
+                     learning_rate=LEARNING_RATE,
+                     clean_train=CLEAN_TRAIN,
+                     testing=False,
+                     backprop_through_attack=BACKPROP_THROUGH_ATTACK,
+                     nb_filters=NB_FILTERS, num_threads=None,
+                     label_smoothing=0.1):
+    """
+    MNIST cleverhans tutorial
```
npapernot: Typo

PR #737, opened 2018-10-05
leave W0101 disabled

goodfeli:


.pylintrc
```diff
@@ -45,6 +45,26 @@ msg-template='{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}'
...
+#      return 0
+#
+#  W0101 complains that the `return 0` is never reached.
+#  However, if we do not have the return statement, we get errors on lines
+#  such as:
+#
+#  x = a.f()
+#
+#  because pylint sees no return statement in f.
+#  We can't have both these checks, and between them, it is better to disable
```
npapernot: Which checks are you referring to here?

goodfeli:
I was able to remove W0101 and have all the tests pass on my laptop. I wonder if maybe one of the travis configs uncovers a different failure.

goodfeli:
Yes, if I run it on Travis, I do get a different check that fails. I've updated the comment in .pylintrc to specifically identify the other check as E1111.

npapernot:
Thanks for updating the comment, it's much more clear now

PR #743, opened 2018-10-06
Fix README typo

WindQAQ:
the path should be `cleverhans/devtools/tests/test_format.py` instead of `cleverhans/devtools/tests_test_format.py`


goodfeli:
Thanks!

PR #753, opened 2018-10-08
deleted contents of certification folder

RAditi:


RAditi:
Changed email address

PR #755, opened 2018-10-08
add max_confidence.py

goodfeli:


cleverhans/max_confidence.py
```diff
@@ -0,0 +1,106 @@
...
+
+from six.moves import range
+import tensorflow as tf
+
+from cleverhans.attacks import Attack
+from cleverhans.model import Model
+from cleverhans.attacks import ProjectedGradientDescent
+
+
+class MaxConfidence(Attack):
```
npapernot: Is there a specific reason for including this attack outside of the file `attacks.py`?

cleverhans/max_confidence.py
```diff
@@ -0,0 +1,106 @@
...
+
+  def __init__(self, model, sess=None):
+    if not isinstance(model, Model):
+      raise TypeError("Model must be cleverhans.model.Model, got " +
+                      str(type(model)))
+
+    super(MaxConfidence, self).__init__(model, sess)
+    self.base_attacker = ProjectedGradientDescent(model, sess=sess)
+    self.structural_kwargs = self.base_attacker.structural_kwargs
+    self.feedable_kwargs = self.feedable_kwargs
```
npapernot: Is this a typo for `self.base_attacker.feedable_kwargs`?

cleverhans/max_confidence.py
goodfeli: No, I originally had it in a separate file because I was developing it in my own repo, but it could be moved into the main attacks.py easily if that is where it goes. I hadn't move it anywhere specific yet because I'm not sure what's going on with attacks.py vs attacks_tf.py.

cleverhans/max_confidence.py
goodfeli: Yes. I guess I must have been running the .pyc the last time I ran this.

cleverhans/max_confidence.py
npapernot: That makes sense. I would say that most new attacks are implemented in `attacks.py` unless they rely on an auxiliary non-tensorflow component or they rely on an existing class that is used by the CleverHans attack class. This is at least how I drew the line so far when merging code. Let me know if you have any suggestions

goodfeli:
OK, it's moved.

PR #775, opened 2018-10-11
Add two more bundling recipes

goodfeli:


goodfeli:
This adds two more bundling recipes, one that uses a fixed amount of computation so that it can make a fair comparison between two models, and another that uses random search so it should work even when there are gradient masking problems. The PR also adds some command line flags that make it possible to choose which recipe to run and where to save the report.

cleverhans/attack_bundling.py
```diff
@@ -156,6 +156,108 @@ def basic_max_confidence_recipe(sess, model, x, y, nb_classes, eps,
...
+
+def random_search_max_confidence_recipe(sess, model, x, y, eps,
+                                        clip_min, clip_max,
+                                        report_path, batch_size=BATCH_SIZE,
+                                        num_noise_points=10000):
+  """Max confidence using random search.
+
+  References:
+  https://openreview.net/forum?id=H1g0piA9tQ
+    Describe the max_confidence procedure used for the bundling in this recipe
```
npapernot: Is this intended as a TODO?

goodfeli: It's a missing 's' typo. I sent the fix, now it says "describes" rather than "describe". The comment is specifying how this reference is related to the code.

PR #786, opened 2018-10-14
add plot_success_fail_curve.py

goodfeli:


cleverhans/plot/success_fail.py
```diff
@@ -0,0 +1,250 @@
...
+
+  fail_optimal = []
+  success_optimal = []
+  fail_upper_bound = []
+  fail_lower_bound = []
+  success_bounded = []
+
+  bounded = False
+
+  while True:
```
carlini: Might be more clear what's going on here to say
`while next_idx < len(all_triplets)`
even if it is redundant.

PR #796, opened 2018-10-16
make_confidence_report now optionally saves adversarial examples

goodfeli:


cleverhans/confidence_report.py
```diff
@@ -42,6 +43,7 @@
 MC_BATCH_SIZE = 16 * num_devices
 NB_ITER = 40
 BASE_EPS_ITER = None # Differs by dataset
+SAVE_ADVX = 1
```
carlini: `True` is probably better if this is a bool.

goodfeli: We've been using integers 0 and 1 to represent bools for command line flags.

The `flags` module is very annoying for bool types.

One of the problems with bools is the following:

For everything other type, you can pass either "--flag value" or "--flag=value" and they both work. For bools, "--flag value" results in `flag` set to True and `value` being appended to `argv`. If the author of the script doesn't check `argv` for unparsed values they silently use the wrong value of the bool.

Because of this we've been using FLAGS.define_integer to handle boolean values. Unfortunately, define_integer raises an error if you pass it `True` as a default value.

carlini: I disagree in the way cleverhans has decided to handle this case then, but definitely it should be consistent.

Issue #795, opened 2018-10-16
Improve the GPU canary


goodfeli:
I have a script in my personal repo called "upper_bound.py".
It's an upper bound on how much it takes to get a canary failure when the GPUs are having whatever sort of problem it is we keep encountering.
This script uses a simple weight decay loss, so I think we could make a model-free canary that just tries to do one step of synchronous gradient descent and sees if all the replicas agree. We'd want to make sure that we don't make a Variable that ends up in any collections. I think we could do this by just calculating synchronous gradients of an expression based on a constant weight tensor rather than weight decay on a Variable.
If we get such a canary working, we can use it to replace the current one. This would be nice because:
- The current canary has to be disabled whenever the loss function is stochastic, so we have no canary guarding jobs that involve training with dropout, training on noisy inputs, etc.
- We could put the new canary in non-training contexts. For example, we could put it the multi-GPU batch eval driver. This way we could have more confidence that our robustness benchmarking results aren't compromised by GPUs returning garbage results.

PR #797, opened 2018-10-16
Better canary

goodfeli:


goodfeli:
I forgot to put it in the commit messages, but this closes #795

carlini:
I know you've been having issues with this a lot, but is this a common enough issue to warrant adding it to cleverhans?

goodfeli:
Anyone in our group using CleverHans for multi-GPU workloads will encounter it somewhere between once per day and once per week. It's hard to tell how much it will affect people in other labs without access to their hardware. If it does affect other group's experiments, it's likely to do so by silently introducing wrong results into their evaluation benchmarks.

carlini:
I see, okay.

PR #810, opened 2018-10-17
Update bundler to make it possible to control step sizes of both PGD attacks

goodfeli:
Previously the second attack was hard-coded to have 25X smaller step size.
When eps is small to start with, this can result in the second eps_iter being absurdly small. I found that on CIFAR-10 the attack is stronger if I set the smaller eps_iter to 1./255 (the smallest variation present in the data) rather than to much smaller than that.

### DIFF
--- a/cleverhans/confidence_report.py
+++ b/cleverhans/confidence_report.py
@@ -108,6 +108,9 @@ def make_confidence_report_bundled(filepath, train_start=TRAIN_START,
   assert x_data.max() <= max_value
   assert x_data.min() >= min_value
 
+  assert eps_iter <= eps
+  assert eps_iter_small <= eps
+
   # Different recipes take different arguments.
   # For now I don't have an idea for a beautiful unifying framework, so
   # we get an if statement.
--- a/cleverhans/confidence_report.py
+++ b/cleverhans/confidence_report.py
@@ -86,14 +86,20 @@ def make_confidence_report_bundled(filepath, train_start=TRAIN_START,
   if 'CIFAR' in str(factory.cls):
     base_eps = 8. / 255.
     base_eps_iter = 2. / 255.
+    base_eps_iter_small = 1. / 255.
   elif 'MNIST' in str(factory.cls):
     base_eps = .3
     base_eps_iter = .1
+    base_eps_iter_small = None
   else:
     raise NotImplementedError(str(factory.cls))
 
   eps = base_eps * value_range
   eps_iter = base_eps_iter * value_range
+  if base_eps_iter_small is None:
+    eps_iter_small = None
+  else:
+    eps_iter_small = base_eps_iter_small * value_range
   nb_iter = 40
   clip_min = min_value
   clip_max = max_value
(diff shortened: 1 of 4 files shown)


cleverhans/confidence_report.py
```diff
@@ -90,14 +90,20 @@ def make_confidence_report_bundled(filepath, train_start=TRAIN_START,
   if 'CIFAR' in str(factory.cls):
     base_eps = 8. / 255.
     base_eps_iter = 2. / 255.
+    base_eps_iter_small = 1. / 255.
   elif 'MNIST' in str(factory.cls):
     base_eps = .3
     base_eps_iter = .1
+    base_eps_iter_small = None
```
nottombrown: Should we have a fixed `eps_iter_small` like in CIFAR here, or is `.1/25` optimal?

scripts/make_confidence_report_bundled.py
```diff
@@ -24,14 +24,17 @@
...
+# pylint: disable=C0413
+silence()
 from cleverhans.confidence_report import make_confidence_report_bundled
 from cleverhans.confidence_report import TRAIN_START, TRAIN_END
 from cleverhans.confidence_report import TEST_START, TEST_END
 from cleverhans.confidence_report import WHICH_SET
 from cleverhans.confidence_report import RECIPE
 from cleverhans.confidence_report import REPORT_PATH
-from cleverhans.utils_tf import silence
-silence()
```
nottombrown: Do we want to include this in the PR?

cleverhans/confidence_report.py
goodfeli: I'm not sure if it's optimal, but it's stronger than .1

I originally made this method testing on MNIST. On MNIST, I started with 40 and .1 because that's what we've usually used in other contexts. When I moved up to 1000 iterations I decided that if I was going to 25X the iteration count I could afford to divide the step size by 25. That way the optimizer can still travel the same total distance in a straight line if it chooses to but it has much more flexibility than if I kept the step size constant. For MNIST, this strategy worked well and made the attack stronger, but it's definitely possible that a stronger attack is out there. Totally coincidentally, it turns out that .1 / 25 = 1.02 / 255, so this is very nearly the same step size as the smallest change in brightness representable by the 8-bit MNIST training data.

For CIFAR, we use a smaller eps to start with (8 / 255) just because CIFAR is so much harder and that's the threat model that's been popular in the literature. On CIFAR, when I tried dividing our original eps_iter of 2. / 255. by 25 the 1000 step attack was actually weaker than the 40 step attack. I think this is because eps_iter became smaller than the minimum storeable difference in the dataset, 1. / 255.

So tl;dr I have a superstition that eps_iter shouldn't be smaller than 1. / 255, and for MNIST we're already satisfying that.

scripts/make_confidence_report_bundled.py
goodfeli: Moving the silence call? Yeah, if you don't mind. I could split it into a separate PR if you want but it seemed small enough to not be worth the effort.

goodfeli:
Ready for re-review

PR #814, opened 2018-10-18
add cleverhans.plot.image and show_images.py

goodfeli:


cleverhans/plot/image.py
```diff
@@ -0,0 +1,146 @@
...
+    assert channels in [1, 3]
+
+  actual_min = ndarray.min()
+  actual_max = ndarray.max()
+
+  if min_val is not None:
+    assert actual_min >= min_val
+    assert actual_max <= max_val
+
+  if 'float' in str(ndarray.dtype):
```
catherio: It might be slightly better to use `np.issubdtype`

```suggestion
  if np.issubdtype(ndarray.dtype, np.floating):
```

goodfeli: Thanks, that's a useful function. I didn't know about it.

goodfeli:
The new suggestions feature is awesome!

PR #817, opened 2018-10-18
change tests to tensorflow style

goodfeli:
(my conversion script before apparently left out the tests)

PR #818, opened 2018-10-18
Check that all kwargs to generate_np were declared

goodfeli:
This PR looks big but actually that is mostly just because it is built on top of https://github.com/tensorflow/cleverhans/pull/817

After that one is merged, this one will be small.


tests_tf/test_attacks.py
```diff
@@ -242,29 +242,6 @@ def test_generate_np_clip_works_as_expected(self):
     self.assertClose(np.min(x_adv), -0.2)
     self.assertClose(np.max(x_adv), 0.1)
 
-  def test_generate_np_caches_graph_computation_for_eps_clip_or_xi(self):
```
AlexeyKurakin: why removing this test?

goodfeli: The test was written wrong, and I'm not sure what the right version of it is.

It's in the test class for `FastGradientMethod`, but it uses a `xi` parameter that `FastGradientMethod` doesn't have.

The test just checks to make sure that some call doesn't cause a new graph to be built. The author presumably intended for one graph to get built and then 2 `xi` values to be fed through it, but all they check for is that a second graph doesn't get built. The test passed because `xi` is just discarded with no error message, and no new graph is built for the new, ignored value of `xi`.

I think that the original author intended to write this test as a method of the test class for `VirtualAdversarialMethod`. When I moved it there, it doesn't work, but not because the test has found a failure in the library, just because the test isn't written right. I don't know enough about that part of the library to write the test correctly in a reasonable amount of time.

Issue #357, opened 2018-02-06
generate_np doesn't support variable dtypes


goodfeli:
The dtypes of most tensors in the `generate_np` caching system are hard-coded to float32. If you pass in a float64 tensor, nothing sensible happens. This is a restriction compared to the `generate` interface.

I have never been a fan of `generate_np` and would be in favor of just removing it since it's one of the more difficult features to maintain and doesn't add any actual adversarial example functionality.

We may also want to investigate whether eager mode + generate provide the functionality that generate_np is meant to provide.

Finally, if we want to keep generate_np, the interface should probably be changed, so that there are multiple allowed dtypes for each argument and the choice of specific dtype is treated as a structural argument.

goodfeli:
cc @npapernot @carlini

npapernot:
I don't use generate_np much so I don't have a strong opinion either way, but transitioning to eager mode would be a good compromise for the use cases I can think of

goodfeli:
SGTM

PR #819, opened 2018-10-18
Fix most of the known problems with generate_np graph caching system

goodfeli:
This fixes issues where we had false cache hits and thus ran the wrong graph when we passed eps=None one one call and then eps=1. or some other value on a later call, or vice versa, or did the same with any other optional parameter, like clip_min, clip_max, etc.

This also fixes issues where generate_np didn't check the type of the arguments / always casted everything to the same dtype. This meant that generate_np and generate would silently behave differently and that it wasn't possible to implement some dtypes (usually float64) with generate_np.

A side benefit of this PR is that it's no longer necessary to declare the type of feedable keyword arguments in the Attack subclass constructor. The types are just inferred from the numpy values when generate_np is called.

The main problem with this PR is that it is a breaking change. Quite a lot of existing client code uses `generate_np` with float64 / int64 variables, relying on undocumented behavior to cast them to float32 / int32.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -798,7 +798,7 @@ class MomentumIterativeMethod(Attack):
       ax = x + utils_tf.clip_eta(ax - x, self.ord, self.eps)
 
       if self.clip_min is not None and self.clip_max is not None:
-        ax = tf.clip_by_value(ax, self.clip_min, self.clip_max)
+        ax = utils_tf.clip_by_value(ax, self.clip_min, self.clip_max)
 
       ax = tf.stop_gradient(ax)
 
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -462,7 +462,7 @@ def fgm(x,
                               "currently implemented.")
 
   # Multiply by constant epsilon
-  scaled_grad = eps * normalized_grad
+  scaled_grad = utils_tf.mul(eps, normalized_grad)
 
   # Add perturbation to original example to obtain adversarial example
   adv_x = x + scaled_grad
@@ -533,8 +533,10 @@ class ProjectedGradientDescent(Attack):
 
     # Initialize loop variables
     if self.rand_init:
-      eta = tf.random_uniform(tf.shape(x), -self.rand_minmax,
-                              self.rand_minmax, dtype=self.tf_dtype)
+      eta = tf.random_uniform(tf.shape(x),
+                              tf.cast(-self.rand_minmax, x.dtype),
+                              tf.cast(self.rand_minmax, x.dtype),
+                              dtype=x.dtype)
     else:
       eta = tf.zeros(tf.shape(x))
     eta = clip_eta(eta, self.ord, self.eps)
@@ -576,7 +578,7 @@ class ProjectedGradientDescent(Attack):
 
       # Clipping perturbation according to clip_min and clip_max
       if self.clip_min is not None and self.clip_max is not None:
-        adv_x = tf.clip_by_value(adv_x, self.clip_min, self.clip_max)
+        adv_x = utils_tf.clip_by_value(adv_x, self.clip_min, self.clip_max)
 
       # Clipping perturbation eta to self.ord norm ball
       eta = adv_x - x
@@ -589,7 +591,7 @@ class ProjectedGradientDescent(Attack):
     adv_x = x + eta
     if self.clip_min is not None or self.clip_max is not None:
       assert self.clip_min is not None and self.clip_max is not None
-      adv_x = tf.clip_by_value(adv_x, self.clip_min, self.clip_max)
+      adv_x = utils_tf.clip_by_value(adv_x, self.clip_min, self.clip_max)
 
     asserts = []
 
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -165,7 +165,7 @@ class Attack(object):
                        " provided")
 
     packed = self.construct_variables(kwargs)
-    fixed, feedable, feedable_types, hash_key = packed
+    fixed, feedable, _feedable_types, hash_key = packed
 
     if hash_key not in self.graphs:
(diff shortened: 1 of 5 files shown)


goodfeli:
This PR includes the commits from the preceding PR: https://github.com/tensorflow/cleverhans/pull/818

So this one will become slightly smaller / slightly easier to review independently if that one is merged first.

cleverhans/attacks.py
```diff
@@ -838,7 +865,7 @@ def __init__(self, model, sess=None, dtypestr='float32', **kwargs):
 
     super(SaliencyMapMethod, self).__init__(model, sess, dtypestr, **kwargs)
 
-    self.feedable_kwargs = {'y_target': self.tf_dtype}
+    self.feedable_kwargs = tuple(['y_target'])
```
AlexeyKurakin: I think you can use `('y_target',)`  which is shorter than `tuple(['y_target'])`

cleverhans/attacks.py
```diff
@@ -2264,3 +2291,34 @@ def attack(self, x, true_y):
   def attack_class(self, x, target_y):
     adv = self.base_attacker.generate(x, y_target=target_y, **self.params)
     return adv
+
+def arg_type(arg_names, kwargs):
+  """
+  Returns a hashable summary of the types of arg_names within kwargs.
```
AlexeyKurakin: Please add more details description of the function.
In particular, please add description of arguments and return value.
It would be useful to know that all values of kwargs has to be numpy arrays or TF tensors

cleverhans/attacks.py
```diff
@@ -158,7 +164,8 @@ def generate_np(self, x_val, **kwargs):
       raise ValueError("Cannot use `generate_np` when no `sess` was"
                        " provided")
 
-    fixed, feedable, hash_key = self.construct_variables(kwargs)
+    packed = self.construct_variables(kwargs)
+    fixed, feedable, _feedable_types, hash_key = packed
```
AlexeyKurakin: Looks like `_feedable_types` is unused.
I though it's more common to just use _ for unused values (i.e. `fixed, feedable, _, hash_key = packed`), but I might be wrong.

goodfeli:
"My main concern is that after this change you have to always explicitely cast all floating point arguments to np.float32, which is inconvenient, error prone and also cause lines to be longer."

-> it isn't quite true that you always have to explicitly cast all floating point arguments to np.float32. You have to either do that *or* write the rest of your code (e.g. the model you're attacking) in a way that accepts something other than float32.

goodfeli:
"My main concern is that after this change you have to always explicitely cast all floating point arguments to np.float32, which is inconvenient, error prone and also cause lines to be longer." -> what specifically is the error you're concerned about for this being error prone? Currently, the error is that you can attempt to write code that runs in float64 and it silently gets downcasted to a less accurate format.

goodfeli:
Do you have an alternate proposal for how to support variable dtypes? Or you're just saying to not support them?

goodfeli:
"I would argue that it's a burden for user to always cast arguments to np.float32. Users will likely forget to do this, moreover it may break existing code (For the reason I described above)." -> if it breaks, they get an error though. The current version silently runs the wrong graph.

goodfeli:
One way that we could avoid breaking existing client code is to change each of the attack implementations. Each subclass could downcast each variable that needs to be downcast (and possibly issue a warning when it does so).

If the problem is that you're not meant to call `Subclass.generate` with some argument that it declares set to `float64`, it feels weird to fix that in a generic `Attack.generate_np`, or to fix it in a part of `Subclass.__init__` that affects only `generate_np`.

cleverhans/attacks.py
goodfeli: Done

cleverhans/attacks.py
goodfeli: Added, but let me know what you think of the new version

cleverhans/attacks.py
goodfeli: Done

goodfeli:
I've pushed an example of how I can get rid of one float32 cast in the numpy calls in the tests. Instead of declaring a type for the argument in self.feedable_kwargs, I rewrite the `generate` method so that it casts `clip_min` and `clip_max` if needed. The advantage of this is that a user who doesn't want to downcast to `float32` doesn't have to. If they write a model and pass an `x` in `float64`, `clip_min` can stay `float64`. It only gets downcast if it's necessary to prevent an exception.

If you think this works, I can redo the rest of the calls.

AlexeyKurakin:
> Do you have an alternate proposal for how to support variable dtypes? Or you're just saying to not support them?

I think, I'm not sure I fully understand usecase for support of different dtypes for attack arguments.
Right now you already can specify dtype for floating point attack arguments (via dtypestr in __init__), and I guess it's safe to assume that all integer type should be int32 (I think all of the integer args specify number of iterations, thus int32 should be enough).
Do you have any use case in mind where you actually need different float dtypes for different arguments to the same attack?

Is it the problem that everything get cast to float32 by default?

goodfeli:
Let's say that you have a defense that is getting a low error rate for the PGD attack. You think that this might just be due to rounding error and your model is small, so you want to try running the attack in float64 on CPU. Currently you can do that by making sure everything you pass to `Attack.generate` is float64. If you use `generate_np` instead you probably expect it to do the same thing, but instead it will silently downcast to float32. The silent downcasting could easily lead to people thinking that their defense works for float64 when it doesn't. There's also no interface that allows the user to turn off the downcasting in `generate_np`; you have to edit the attack code or switch back to using `generate`.

We've had an issue open to support variable dtypes since February: https://github.com/tensorflow/cleverhans/issues/357 This started because a user wrote to cleverhans-dev saying they had attacks that worked in float64 but not float32 and they wanted to run them with generate_np.

goodfeli:
I'm not squashing this because squashing it would make about half of #820 cause merge conflicts

Issue #695, opened 2018-09-30
Possible speedup in `random_crop_and_flip`


goodfeli:
`_pad_img` could be rewritten to operate on a whole batch, and avoid needing a map_fn. Check how this affects runtime of CIFAR-10 training pipeline.

david-berthelot:
Without map_fn, the cropping happens at the same location within a batch. This in turn reduces stochasticity for gradient descent.

PR #822, opened 2018-10-19
Do padding and flipping batch-wise

WindQAQ:
#695 . Not sure if this is still needed. 

In Tensorflow doc, both [tf.image.resize_image_with_crop_or_pad](https://www.tensorflow.org/api_docs/python/tf/image/resize_image_with_crop_or_pad) and [tf.image.random_flip_left_right](https://www.tensorflow.org/api_docs/python/tf/image/random_flip_left_right) can do on a 4D tensor, i.e., a batch of images. Applying these two operations on batch-level can speedup 3x. Here is my [testing code](https://gist.github.com/WindQAQ/71d90117ded4b0a41d9d57c71fa30eab).

### DIFF
--- a/cleverhans/augmentation.py
+++ b/cleverhans/augmentation.py
@@ -41,24 +41,18 @@ def random_crop_and_flip(x, pad_rows=4, pad_cols=4):
   """Augment a batch by randomly cropping and horizontally flipping it."""
   rows = tf.shape(x)[1]
   cols = tf.shape(x)[2]
-
-  def _pad_img(img):
-    """Pad an individual image"""
-    return tf.image.resize_image_with_crop_or_pad(img, rows + pad_rows,
-                                                  cols + pad_cols)
+  channels = x.get_shape()[3]
 
   def _rand_crop_img(img):
     """Randomly crop an individual image"""
-    channels = img.get_shape()[2]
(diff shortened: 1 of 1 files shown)


WindQAQ:
What I do in this PR is pad and crop images batch-wise. The cropping part still uses `tf.map_fn` so each image is cropped with a separate offset. (In fact, I cannot find any idea to crop images on batch-level with seperate offsets without `tf.map_fn`.). It's fine to wait him 😄

david-berthelot:
Yes using tf.map_fn seems to be the way like you pointed.

One thing that I started to change is to go to single item functions like `random_shift`. The batching is handled separately by passing a single item function to `batch_augment`. 

The API you improved is meant to become obsolete since (1) it's a batch API but also (2) since it gives worse results for augmentation (due to how image padding is  done and previously random cropping being single offset - which you addressed).

To see the new API in action, check the cifar10 tutorial:
https://github.com/tensorflow/cleverhans/blob/master/cleverhans_tutorials/cifar10_tutorial_tf.py#L85

AlexeyKurakin:
Talked to @david-berthelot, he thinks it's still useful to merge this PR, even despite the fact that API would be obsolete.

Issue #823, opened 2018-10-19
Adversarial image not clipped for first iteration of PGD (MadryEtAl)


vipinpillai:
As part of the ProjectedGradientDescent attack, when used with default_rand_init=True, _eta_ is initialized with random noise [here](https://github.com/tensorflow/cleverhans/blob/master/cleverhans/attacks.py#L508) and then clipped within L-infinity norm of _eps_.
However, when the the adversarial image is fed to the PGD attack for the very first iteration, [x+e](https://github.com/tensorflow/cleverhans/blob/master/cleverhans/attacks.py#L547) is not clipped to lie in the valid image range.

This shouldn't significantly affect the PGD adversarial attack results though since it is only the first iteration which doesn't provide meaningful gradients.

goodfeli:
Thanks, good catch!

PR #824, opened 2018-10-19
Clip image to be in valid range during first iteration of PGD

vipinpillai:
This is a potential fix for issue [823](https://github.com/tensorflow/cleverhans/issues/823).

_fgm()_ method already does the image clipping before returning the adversarial image, hence I've just moved the clipping to be done before feeding in to the _FGSM.generate()_.

### DIFF
--- a/cleverhans/attacks.py
+++ b/cleverhans/attacks.py
@@ -543,11 +543,13 @@ class ProjectedGradientDescent(Attack):
       return tf.less(i, self.nb_iter)
 
     def body(i, e):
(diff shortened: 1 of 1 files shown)


goodfeli:
OK to go with https://github.com/tensorflow/cleverhans/pull/825 instead? I rewrote the loop body to use adv_x as the persistent variable, which is hopefully less confusing, and added several asserts to make sure we catch similar bugs.

vipinpillai:
Thanks @goodfeli . That's a much cleaner solution.

PR #829, opened 2018-10-21
Fix or disable wrong implementations of ord=1

goodfeli:


tests_tf/test_attacks.py
```diff
@@ -1120,5 +1208,6 @@ def test_keep_pixels_on_image(self):
     self.assertTrue(np.mean(old_labs == new_labs) < 0.3)
 
 
-if __name__ == '__main__':
-  unittest.main()
+
+#if __name__ == '__main__':
```
AlexeyKurakin: why this code is commented?

tests_tf/test_attacks.py
```diff
@@ -250,6 +268,69 @@ def test_generate_np_clip_works_as_expected(self):
...
+    # This test makes sure that `optimize_linear` actually finds the optimal
+    # perturbation for ord=1.
+    # A common misconcpetion is that FGM for ord=1 consists of dividing
+    # the gradient by its L1 norm.
+    # If you do that for the problem in this unit test, you'll get an
+    # objective function value of ~1.667. The optimal result is 2.
+
+    # We need just one example in the batch and two features to show the
+    # common misconception is suboptimal.
+    grad = tf.placeholder(tf.float32, [1, 2])
```
AlexeyKurakin: Here and in the next test, maybe add `shape=`:
```
grad = tf.placeholder(tf.float32, shape=[1, 2])
```

Otherwise it might be confusing to read this line if you don't remember exact API contract for tf.placeholder.

tests_tf/test_attacks.py
```diff
@@ -250,6 +268,69 @@ def test_generate_np_clip_works_as_expected(self):
...
+
+    # Make sure the largest entry of the gradient for the test case is
+    # negative, to catch
+    # the potential bug where we forget to handle the sign of the gradient
+    eta, objective = self.sess.run([eta, objective],
+                                   feed_dict={grad: np.array([[1., -2.]])})
+
+    # Make sure the objective is optimal.
+    # This is the solution obtained by doing the algebra by hand.
+    assert objective == 2., objective
```
AlexeyKurakin:
```
self.assertClose(objective, 2.0)
```

Unit tests should use `self.assert*` methods instead of python assertion. If python assertion fails it will just through AssertException, which will fail unit test, but won't provide very good error messages. 
`self.assert*` methods provide much more intuitive error messages when they fail in the test.

Same applies to asserts below in this method and next method

tests_tf/test_attacks.py
```diff
@@ -250,6 +268,69 @@ def test_generate_np_clip_works_as_expected(self):
...
     self.assertClose(np.max(x_adv), 0.1)
 
+class TestFastGradientMethod(CommonAttackProperties):
+
+  def setUp(self):
+    super(TestFastGradientMethod, self).setUp()
+
+    self.attack = FastGradientMethod(self.model, sess=self.sess)
+
+  def test_optimize_linear_l1(self):
```
AlexeyKurakin: I think it would be better to put `test_optimize_linear_l1_ties` and `test_optimize_linear_l1` into separate test class.
These two tests are testing `optimize_linear` helper function, which is not specific to FGSM attack.
Also these two test do not utilize any methods declared in `CommonAttackProperties`

You can have a class which is called, for examples, `TestAttackHelpers` which would contain tests for `optimize_linear` and potentially other helper methods:

```
class TestAttackHelpers(CleverHansTest)):

  def test_optimize_linear_l1(self):
    ...

  def test_optimize_linear_l1_ties(self):
    ...
```


Also, I think it make sense to add tests for other values of `ord` in the `optimize_linear`.
While they not necessarily have special corner cases, it's useful to test them in case somebody will be doing refactoring and introduce some error in the code.
Direct tests for `optimize_linear` (instead of relying on tests for PGD, FGSM, etc...) will make finding such errors easier.

cleverhans/attacks.py
```diff
@@ -2456,3 +2423,50 @@ def arg_type(arg_names, kwargs):
     dtypes.append(dtype)
   dtypes = tuple(dtypes)
   return (passed, passed_and_not_none, dtypes)
+
+def optimize_linear(grad, eps, ord=np.inf):
+  """
+  Solves for the optimal input to a linear function under a norm constraint.
+
+  Optimal_perturbation =
+    argmax(eta, norm(eta, ord) < 1) dot(eta, grad)
```
AlexeyKurakin: 1) this function solved the problem subject to constraint ||eta|| <= epsilon (rather than <= 1.0)
2) Does people typically use LaTex notation in the CleverHans comments? It might be easier to read (at least for people familiar with latex):

```
 Optimal_perturbation = argmax_{eta, ||eta||_{ord} < eps} dot(eta, grad)
```

tests_tf/test_attacks.py
goodfeli: Done

cleverhans/attacks.py
goodfeli: Done

goodfeli:
Ready for re-review

PR #832, opened 2018-10-22
Fix UnboundLocalError when importing cleverhans without scripts/ folder

cuihaoleo:
Bug fix.


AlexeyKurakin:
PR looks good, but let's wait until it will be tested by Travis

Issue #780, opened 2018-10-12
pgd_attack should have a better name / documentation


goodfeli:
If you look at `pgd_attack` in isolation it looks like it's what you should run if you want to use a PGD attack. The doc should at the very least steer people to the ProjectedGradientDescent class. I'm not 100% sure what's going on with this function, but the doc should specify that `pgd_attack` is mostly intended for implementing SPSA, if that's correct, and if that's not correct the doc should explain what actually is going on. Also, I suspect the name `pgd_attack` ought to be changed. For one thing, it should be different to avoid steering users away from ProjectedGradientDescent, but also I think a large part of the point of this function is that it doesn't need to literally use the gradient, so the "g" seems wrong. Maybe it should just be called `projected_optimizer` to reflect that it's relatively generic and reflect that it's not an `Attack`?

juesato:
I originally implemented `pgd_attack` to allow easily implementing a large family of projected gradient attacks (including PGD, with modifications such as e.g. the type of projection operation, the Optimizer class, steepest descent in L2 vs Linf space, whether to use analytic or finite difference gradients, etc.).

It definitely looks strange in context, because it's only used for SPSA in cleverhans. I think the easiest thing to do would be to just add to the docstring that this is a helper function only used in SPSA, and that `ProjectedGradientDescent` implements the PGD attack. WDYT?

goodfeli:
There are 2 things I think we should improve:

1) Having the word `attack` in the name  is strange when there is an `Attack` class and this is not a direct implementation of one of those classes. In the context of this module, it's especially confusing, because many of the other functions in this model really are just the implementation of some `Attack` class. So I would advocate for finding a new name that doesn't have `attack` in it.

2) This is less of a priority for me, and definitely more pedantic, but having `g` for gradient in the name seems to sell short what this does. A lot of its purpose is to optimize functions where the gradient is not available or isn't useful for optimization. So I would advocate for a name that doesn't have the word `gradient` or a `g` that stands for `gradient` in it.

juesato:
Re 1: That makes sense to me. How do you feel about `pgd_attack_helper`? It is an "attack" in the sense that it takes a clean image Tensor (+ some other stuff) and returns an adversarial image Tensor. But I agree it's confusing, esp given most other methods with `attack` in the name are implementations of an `Attack` class.
Other name options are `pgd_helper` or `make_pgd_adv_image`.

[...]

PR #833, opened 2018-10-22
SPSA confidence report

goodfeli:


cleverhans/attacks_tf.py
```diff
@@ -1842,7 +1872,12 @@ def cond(i, *_):
     # project_perturbation
     check_diff = tf.no_op()
 
-  with tf.control_dependencies([check_diff]):
+  if clip_min is None or clip_max is None:
+    raise NotImplementedError("SPSA only supports clipping for now")
```
AlexeyKurakin: this code is inside pgd_attack function.
Why you referring to SPSA attack in the text of the error?

goodfeli: The `pgd_attack` has a confusing name.

You're proving the point I made in this issue :)  https://github.com/tensorflow/cleverhans/issues/780

PR #845, opened 2018-10-24
Closes #761 (treat callables consistently)

### LINKED ISSUE #761: Treat callables consistently
Many Attacks support passing in a callable rather than a Model. Currently, they are wildly inconsistent.
Many treat `__call__` as probs (incredibly dangerous, can't believe we're still doing this anywhere) while many treat `__call__` as logits.

I think it would be best to just drop support for all these callables rather than try to change the semantics.

goodfeli:


cleverhans/attacks_tfe.py
```diff
@@ -8,7 +8,8 @@
 
 from cleverhans import attacks
 from cleverhans import utils
-from cleverhans.model import CallableModelWrapper
+from cleverhans.compat import reduce_sum
```
npapernot: Is this new import a typo?

goodfeli: I had to rebase to open the PR. When I resolved the merge conflict, I thought that import had been added in the master branch. Now I have no idea where it came from. Deleting it.

PR #858, opened 2018-10-26
Fix generate_np bugs

carlini:


cleverhans/attacks.py
```diff
@@ -592,20 +605,19 @@ def body(i, adv_x):
...
+      asserts.append(tf.assert_less_equal(tf.cast(self.eps_iter, dtype=self.eps.dtype), self.eps))
       if self.ord == np.inf and self.clip_min is not None:
         # The 1e-6 is needed to compensate for numerical error.
         # Without the 1e-6 this fails when e.g. eps=.2, clip_min=.5,
         # clip_max=.7
-        asserts.append(tf.assert_less_equal(self.eps,
-                                            1e-6 + self.clip_max
-                                            - self.clip_min))
+        asserts.append(tf.assert_less_equal(tf.cast(self.eps, self.tf_dtype),
+                                            1e-6 + tf.cast(self.clip_max, self.tf_dtype)
```
npapernot: x.dtype?

cleverhans/attacks.py
```diff
@@ -592,20 +605,23 @@ def body(i, adv_x):
 
     _, adv_x = tf.while_loop(cond, body, [tf.zeros([]), adv_x], back_prop=True)
 
-    asserts = []
 
     # Asserts run only on CPU.
```
goodfeli: It's not necessary to merge this PR, but if you feel like it you could change these lines to just use `utils_tf.assert*` rather than the explicit `with tf.device`

carlini: I have to fix the crash anyway, so I'll just add it to this PR.

tests_tf/test_attacks.py
```diff
@@ -255,7 +255,7 @@ def test_generate_np_can_be_called_with_different_eps(self):
                                       clip_min=-5.0, clip_max=5.0)
 
       delta = np.max(np.abs(x_adv - x_val), axis=1)
-      self.assertClose(delta, eps)
+      self.assertTrue(np.all(delta<=(eps+1e-4)))
```
goodfeli: This looks like it's actually changing the properties of the test.

Can you put a copy of the assertClose version in TestFastGradientMethod? The constraint should aways be tight for that attack.

What happened here is first we wrote the test for FastGradientMethod, then other people subclassed it for other attacks, then I renamed it to intermediate CommonAttackProperties class and pushed things that are specific to fast gradient method into a new TestFastGradientMethod subclass.

goodfeli: In the generic version where the constraint might not be tight, you should use assertLessEqual rather than assertTrue.

tests_tf/test_attacks.py
```diff
@@ -454,119 +454,6 @@ def test_attack_strength_np_batched(self):
     new_labs = np.argmax(self.sess.run(self.model.get_logits(x_adv)), axis=1)
     self.assertLess(np.mean(feed_labs == new_labs), 0.1)
```
goodfeli: It's probably OK, but just so I know what's going on, can you explain why this was removed?

tests_tf/test_attacks.py
carlini: It might be better to have one helper function that generates the adversarial examples and then have FGSM test that it's on the boundary and then PGD test that it's in the range [-eps,eps], not just exactly at the boundary.

tests_tf/test_attacks.py
goodfeli: I think all the methods here were moved to the TestMadryEtAl class. I don't fully understand why they needed to be moved. If that is what happened, I think it's probably not the right thing to do, but maybe I'm missing something. The MadryEtAl class doesn't do anything but initialize ProjectedGradientDescent with the right arguments, so it doesn't make a lot of sense for all the tests to be pushed into that highly specialized subclass.

tests_tf/test_attacks.py
```diff
@@ -454,119 +454,6 @@ def test_attack_strength_np_batched(self):
...
 
-
-class TestBasicIterativeMethod(CommonAttackProperties):
```
goodfeli: If the concern here is just that the name had "BasicIterativeMethod" it's probably better to fix it by renaming this to "TestProjectedGradientDescent".

tests_tf/test_attacks.py
carlini: There were two `class TestBasicIterativeMethod`. I removed one and put the methods into the other.

tests_tf/test_attacks.py
```diff
@@ -974,6 +891,40 @@ def test_generate_np_gives_clipped_adversarial_examples(self):
...
+    x_val = np.random.rand(100, 2)
+    x_val = np.array(x_val, dtype=np.float32)
+
+    x_adv = self.attack.generate_np(x_val, eps=1.0, ord=np.inf,
+                                    clip_min=-5.0, clip_max=5.0,
+                                    nb_iter=10)
+
+    orig_labs = np.argmax(self.sess.run(self.model.get_logits(x_val)), axis=1)
+    new_labs = np.argmax(self.sess.run(self.model.get_logits(x_adv)), axis=1)
+    self.assertTrue(np.mean(orig_labs == new_labs) < 0.1)
```
goodfeli: should be assertLess

tests_tf/test_attacks.py
carlini: There were two `class TestBasicIterativeMethod`.

tests_tf/test_attacks.py
```diff
@@ -974,6 +891,40 @@ def test_generate_np_gives_clipped_adversarial_examples(self):
...
+
+    tf.gradients = fn
+
+    x_adv = self.attack.generate_np(x_val, eps=1.0, ord=np.inf,
+                                    clip_min=-5.0, clip_max=5.0,
+                                    nb_iter=11)
+
+    orig_labs = np.argmax(self.sess.run(self.model.get_logits(x_val)), axis=1)
+    new_labs = np.argmax(self.sess.run(self.model.get_logits(x_adv)), axis=1)
+    self.assertTrue(np.mean(orig_labs == new_labs) < 0.1)
```
goodfeli: should be assertLess

tests_tf/test_attacks.py
carlini: [these were from the copied code and were already tests]

tests_tf/test_attacks.py
carlini: [these were from the copied code and were already tests]

tests_tf/test_attacks.py
```diff
@@ -974,6 +891,40 @@ def test_generate_np_gives_clipped_adversarial_examples(self):
...
+
+    ok = [False]
+    old_grads = tf.gradients
+
+    def fn(*x, **y):
+      ok[0] = True
+      return old_grads(*x, **y)
+
+    tf.gradients = fn
```
goodfeli: If the test crashes inside `generate_np` or something like, will all the other tests get stuck using the wrong gradients?

It might be better to test this by running the attack on a model whose fprop method contains the instrumentation, rather than by changing a shared library.

Another option might be to use a `try`-`finally` construction to make sure that the replacement always happens.

tests_tf/test_attacks.py
goodfeli: That's not actually how I'm reading the diff.
I think you moved all the tests into TestMadryEtAl.

I think what happened was there used to be a BasicIterativeMethod attack class. Then later we added the MadryEtAl attack class. Then later we implemented both using a common class called ProjectedGradientDescent, and made BasicIterativeMethod and MadryEtAl just be empty classes that instantiate ProjectedGradientDescent with the right arguments.

There are now TestBasicIterativeMethod and TestMadryEtAl which are basically just smoke tests to make sure those empty child classes can be called. But TestBasicIterativeMethod caused a problem because that was the same name as all the old tests.

I think the right way to fix it is just don't move anything, but change the name of the old TestBasicIterativeMethod to TestProjectedGradientDescent.

tests_tf/test_attacks.py
carlini: This code all already existed. I was just moving it.

carlini:
Rebase got ugly. Closing and will re-open.

PR #865, opened 2018-10-26
Closes #544 (make sure that overshoot param works)

### LINKED ISSUE #544: overshoot does not work for DeepFool attack
Changing `overshoot` parameter does not have any effect in `DeepFool` attack. I think the problem is with `structural_kwargs` in `attack.py` file.
`over_shoot` should be changed to `overshoot`.

goodfeli:


goodfeli:
`parse_params` takes an `overshoot` param. It also takes `kwargs` so if you pass `over_shoot` no error occurs. Another PR that is open right now is going to deprecate `kwargs` so that problem will go away eventually. `generate_np` doesn't raise an error because all the tests called with `over_shoot` and `over_shoot` was declared as an argument.

Issue #856, opened 2018-10-26
Accept sess=None in attacks


erfannoury:
With the current implementation, every object created from the classes based on the `Attack` class need to provide a `tf.Session` instance in the constructor. However, this limits the usage of the attacks to cases where a `Session` has already been created. For instance, there are problems for use cases in which the attacks are used in the graph construction phase and `tf.Session` will be created later and will not be accessible to the user. This happens when using the Estimator API in which attacks can either be used for training or for evaluation. In both cases the graph will be constructed first and creating the session and executing the graph is handled by the Estimator class. Would it be reasonable to move [this check](https://github.com/tensorflow/cleverhans/blob/master/cleverhans/attacks.py#L51-L52) to where `sess` actually will be used?

goodfeli:
Is `sess` even used anywhere other than `generate_np`? It might make more sense to just take `sess` out of the `Attack` class altogether, and have `generate_np` take a `sess` argument that defaults to tf.get_default_session() or something like that.

AlexeyKurakin:
I looked through all the code of the attacks.
Most of the attack only use sess in the generate_np.
However few attacks don't have proper symbolic implementation (CW, EAD, DeepFool, LBFGS) and rely on sess in generate method. Plus one attack (SaliencyMaps) has a flag which controls whether generate uses pure symbolic attack or uses pyfunc with session.

The main problem with current implementation is that `__init__` method of attack class always assigns `self.sess` even if session was not provided to constructor.


So I would say ideal solution would be to rewrite all remaining attacks to have proper symbolic implementation and then move session argument from the constructor to generate_np.
However modifying these few attacks is quite a lot of efforts.

Less ideal solution could be following. We can allow __init__ to accept None as the sess and remove the code which creates default session from the attack classes. And we put `assert self.sess is not None` into `generate` method of those few attacks which does need session.

erfannoury:
Second solution would be much easier, but I don't know whether it's the preferred method. Is there a way I can help make any of these changes?

AlexeyKurakin:
I think it's fine to do second solution for now.

AlexeyKurakin:
@erfannoury if you willing to make this change it would be great

erfannoury:
@AlexeyKurakin Sure, I can start working on them and will submit a PR as soon as possible.

AlexeyKurakin:
[...]

PR #873, opened 2018-10-30
Accept sess=None in (some) attacks

erfannoury:
This PR is a temporary solution to the issue raised in #856.
I have changed `attacks.py` and all tests for `attacks.py` pass.
Additionally, if this is the intended implementation by @AlexeyKurakin, I will add more tests to cover the new changes.


AlexeyKurakin:
Looks good, thanks for making this change

PR #878, opened 2018-11-02
Fix generate np

carlini:
Re-opens the PR #858 with a rebase.

### DIFF
--- a/tests_tf/test_attacks.py
+++ b/tests_tf/test_attacks.py
@@ -202,10 +202,10 @@ class CommonAttackProperties(CleverHansTest):
                                                  **kwargs):
     x_val, x_adv, delta = self.generate_adversarial_examples_np(ord, eps,
                                                                 **kwargs)
-    self.assertClose(delta, eps)
+    self.assertLess(np.max(np.abs(delta-eps)), 1e-3)
     orig_labs = np.argmax(self.sess.run(self.model.get_logits(x_val)), axis=1)
     new_labs = np.argmax(self.sess.run(self.model.get_logits(x_adv)), axis=1)
-    self.assertTrue(np.mean(orig_labs == new_labs) < 0.5)
+    self.assertLess(np.max(np.mean(orig_labs == new_labs)), .5)
 
   def test_invalid_input(self):
     x_val = -np.ones((2, 2), dtype='float32')
@@ -975,6 +975,10 @@ class TestMomentumIterativeMethod(TestProjectedGradientDescent):
       delta = np.max(np.abs(x_adv - x_val), axis=1)
       self.assertClose(delta, 0.5)
 
+  def test_multiple_initial_random_step(self):
+    # There is no initial random step, so nothing to test here
+    pass
+      
 class TestMadryEtAl(TestProjectedGradientDescent):
   def setUp(self):
     super(TestMadryEtAl, self).setUp()
--- a/tests_tf/test_attacks.py
+++ b/tests_tf/test_attacks.py
@@ -546,25 +546,6 @@ class TestBasicIterativeMethod(CommonAttackProperties):
     self.assertTrue(ok[0])
 
 
-class TestMomentumIterativeMethod(TestBasicIterativeMethod):
-  def setUp(self):
-    super(TestMomentumIterativeMethod, self).setUp()
-
-    self.sess = tf.Session()
-    self.model = SimpleModel()
-    self.attack = MomentumIterativeMethod(self.model, sess=self.sess)
-
-  def test_generate_np_can_be_called_with_different_decay_factor(self):
-    x_val = np.random.rand(100, 2)
-    x_val = np.array(x_val, dtype=np.float32)
-
-    for decay_factor in [0.0, 0.5, 1.0]:
-      x_adv = self.attack.generate_np(x_val, eps=0.5, ord=np.inf,
-                                      decay_factor=decay_factor,
-                                      clip_min=-5.0, clip_max=5.0)
-
-      delta = np.max(np.abs(x_adv - x_val), axis=1)
-      self.assertClose(delta, 0.5)
 
 
 class TestCarliniWagnerL2(CleverHansTest):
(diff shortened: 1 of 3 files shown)


carlini:
A bunch of the tests were written incorrectly and should never have been passing to begin with. I'm going to fix them in this PR so at least they pass, but someone later should go back through and figure out what was intended to happen.

carlini:
Except for this crash, everything passes `terminate called after throwing an instance of 'std::bad_alloc'`

goodfeli:
It looks like it's failing for both tf 1.12 builds but none of the other builds

tests_tf/test_attacks.py
```diff
@@ -454,119 +454,6 @@ def test_attack_strength_np_batched(self):
     new_labs = np.argmax(self.sess.run(self.model.get_logits(x_adv)), axis=1)
     self.assertLess(np.mean(feed_labs == new_labs), 0.1)
 
-
-class TestBasicIterativeMethod(CommonAttackProperties):
```
goodfeli: I think instead of having a big section of diff here, this should just be a name change to "TestProjectedGradientDescent"

tests_tf/test_attacks.py
```diff
@@ -454,119 +454,6 @@ def test_attack_strength_np_batched(self):
...
-    orig_labs = np.argmax(self.sess.run(self.model.get_logits(x_val)), axis=1)
-    new_labs = np.argmax(self.sess.run(self.model.get_logits(x_adv)), axis=1)
-    self.assertTrue(np.mean(orig_labs == new_labs) < 0.1)
-
-    tf.gradients = old_grads
-
-    self.assertTrue(ok[0])
-
-
-class TestMomentumIterativeMethod(TestBasicIterativeMethod):
```
goodfeli: I think this shouldn't have a big diff, just the top should have the parent class change to TestProjectedGradientDescent

tests_tf/test_attacks.py
```diff
@@ -927,14 +814,44 @@ def test_generate_np_gives_clipped_adversarial_examples(self):
     self.assertTrue(np.max(x_adv) < .301)
 
 
-class TestMadryEtAl(CleverHansTest):
+class TestProjectedGradientDescent(CommonAttackProperties):
```
goodfeli: I think this shouldn't be part of the diff. It's just checking that the empty child class can call the parent class with the right defaults and not crash.

tests_tf/test_attacks.py
```diff
@@ -927,14 +814,44 @@ def test_generate_np_gives_clipped_adversarial_examples(self):
...
   def setUp(self):
-    super(TestMadryEtAl, self).setUp()
+    super(TestProjectedGradientDescent, self).setUp()
 
     self.sess = tf.Session()
     self.model = SimpleModel()
-    self.attack = MadryEtAl(self.model, sess=self.sess)
+    self.attack = ProjectedGradientDescent(self.model, sess=self.sess)
+
+  def test_generate_np_gives_adversarial_example_linfinity(self):
```
goodfeli: I think these methods here should be in TestProjectedGradientDescent above

tests_tf/test_attacks.py
```diff
@@ -1255,3 +1227,7 @@ def test_keep_pixels_on_image(self):
     new_labs = np.argmax(self.sess.run(self.model.get_logits(x_adv)), axis=1)
     print(np.mean(old_labs == new_labs))
     self.assertTrue(np.mean(old_labs == new_labs) < 0.3)
+
+if __name__ == "__main__":
```
goodfeli: I thought we were removing these statements because they prevent nosetest from correctly supporting "nosetests foo.py:bar" syntax to run only `bar`

tests_tf/test_attacks.py
```diff
@@ -1018,6 +970,26 @@ def test_multiple_initial_random_step(self):
     pass
 
 
+class TestMomentumIterativeMethod(TestBasicIterativeMethod):
```
goodfeli: It would be better if this class didn't move, so that the git history is less confusing

tests_tf/test_attacks.py
carlini: Done

tests_tf/test_attacks.py
carlini: Done

tests_tf/test_attacks.py
carlini: This is just a change of names. I made the big method called TestProjectedGradientDescent, so I had to define TestMadryEtAl

tests_tf/test_attacks.py
carlini: Done

tests_tf/test_attacks.py
carlini: Done

tests_tf/test_attacks.py
carlini: Done, but git diff still gets confused.

carlini:
Rebase broke the history.

The diff still looks somewhat unhappy because I had to merge the two methods together, but it looks somewhat nicer now.

carlini:
Again this issue "terminate called after throwing an instance of 'std::bad_alloc". Going to retry.

carlini:
The test is now getting killed, but everything is passing.

I don't know that much about travis do debug this one. Any suggestions @npapernot @goodfeli?

npapernot:
I restarted the test to see if it passes. Travis sometimes encounters issues reporting back the result of tests

carlini:
It failed again. Interestingly it is consistently just 2.7 on TF 1.8 that is failing here.

npapernot:
Looks like the test commented out was the faulty one

carlini:
Okay so this is very confusing. By removing a test that I never called, the tests pass. Let me put those lines back in and see if it works again.

carlini:
I am 90% sure this is just a travis issue. I installed TF 1.8 on Python 2.7 and I can run the tests successfully. 

If it's okay with you @npapernot @goodfeli I'm going to comment out this test again so we can merge this commit, and then investigate further.

carlini:
I have removed some duplicated work that was going on in the tests. This should allow the tests to pass, but in the future we will need to find a way to make the tests faster or this will happen again.

npapernot:
Thanks, hopefully we will make progress towards migrating to CI with GPUs

goodfeli:
This broke `generate`...

carlini:
What does it break?

goodfeli:
The new assert only works if `eps` is numpy/tensorflow. If it's a float, the new assert causes an AttributeError while trying to read its dtype.

PR #891, opened 2018-11-12
fix unpickling

goodfeli:
Some pickle files for ConfidenceReports are broken because OrderedDict wants to use this constructor argument to unpickle itself. I actually haven't figured out what causes it to use this path so I haven't been able to make a unit test for it.


goodfeli:
OK, it looks like I actually do have a unit test.

I submitted another PR with a unit test that I thought did not catch this problem. It turns out it does catch the problem, but only for a subset of the Travis builds:
https://github.com/tensorflow/cleverhans/pull/890

PR #898, opened 2018-11-20
This is a test commit

carlini:
This commit does nothing to check if travis is broken on Python 2.7 with TF 1.8.


carlini:
It looks like it might be an issue of timing: the new tests might be slow and so travis is failing them.

PR #906, opened 2018-11-21
Fix bugs in SPSA early stopping

goodfeli:
There were two bugs here:
1 If the early stopping threshold was set to 0., early stopping was disabled because this was treated as a value of False. (0. is probably the most important threshold, because the loss is a hinge and a threshold of 0. guarantees misclassification)
2) The early stopping loss was calculated before doing the projection, so it could result in early stopping using a loss that wasn't actually attainable

### DIFF
--- a/cleverhans/attacks_tf.py
+++ b/cleverhans/attacks_tf.py
@@ -1485,7 +1485,7 @@ class TensorOptimizer(object):
   behaviors when being assigned multiple times within a single sess.run()
   call, particularly in Distributed TF, so this avoids thinking about those
   issues. These are helper classes for the `projected_optimization`
-  method. Apart from not using Variables, they follow the an interface very
+  method. Apart from not using Variables, they follow an interface very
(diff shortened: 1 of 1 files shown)


goodfeli:
This PR includes commits from https://github.com/tensorflow/cleverhans/pull/903 . The diff will be smaller after that is merged

cleverhans/attacks_tf.py
```diff
@@ -1485,8 +1485,8 @@ class TensorOptimizer(object):
   behaviors when being assigned multiple times within a single sess.run()
   call, particularly in Distributed TF, so this avoids thinking about those
   issues. These are helper classes for the `projected_optimization`
-  method. Apart from not using Variables, they follow the same interface as
-  tf.Optimizer.
+  method. Apart from not using Variables, they follow the an interface very
```
npapernot:
```suggestion
  method. Apart from not using Variables, they follow an interface very
```

PR #920, opened 2018-12-12
Del class from kwargs in CIFAR

catherio:


catherio:
This was messy for no good reason, trying again

PR #921, opened 2018-12-12
Move the del fix to CIFAR

catherio:


carlini:
We'll merge when tests pass

PR #923, opened 2018-12-12
WIP: reduce amount of new graph nodes created while running certification

goodfeli:
I pushed this WIP branch to keep track of where I was during the hackathon. Adding github comments below to document what's going on

### DIFF
--- a/cleverhans/experimental/certification/dual_formulation.py
+++ b/cleverhans/experimental/certification/dual_formulation.py
@@ -213,7 +213,6 @@ class DualFormulation(object):
     """Function that projects the input dual variables onto the feasible set.
 
     Returns:
-      projected_dual: Feasible dual solution corresponding to current dual
       projected_certificate: Objective value of feasible dual
     """
     type_check(self.lambda_pos) 
--- a/cleverhans/experimental/certification/dual_formulation.py
+++ b/cleverhans/experimental/certification/dual_formulation.py
@@ -20,8 +20,10 @@ class DualFormulation(object):
     Args:
       dual_var: dictionary of dual variables containing a) lambda_pos
         b) lambda_neg, c) lambda_quad, d) lambda_lu
-      neural_net_param_object: NeuralNetParam object created for
-        the network under consideration
+        TODO: where should one look to see the definitions of these?
+              what are their types?
+                they mostly seem to be Variables. lambda_lu can also be a non-var tensor
+      neural_net_param_object: NeuralNetParam object created for the network under consideration
       test_input: clean example to certify around
       true_class: the class label of the test input
       adv_class: the label that the adversary tried to perturb input to
@@ -36,12 +38,7 @@ class DualFormulation(object):
     self.input_minval = tf.convert_to_tensor(input_minval, dtype=tf.float32)
     self.input_maxval = tf.convert_to_tensor(input_maxval, dtype=tf.float32)
     self.epsilon = tf.convert_to_tensor(epsilon, dtype=tf.float32)
-    self.final_linear = (self.nn_params.final_weights[adv_class, :]
-                         - self.nn_params.final_weights[true_class, :])
-    self.final_linear = tf.reshape(self.final_linear,
-                                   shape=[tf.size(self.final_linear), 1])
-    self.final_constant = (self.nn_params.final_bias[adv_class]
-                           - self.nn_params.final_bias[true_class])
+    self.final_linear, self.final_constant = final_linear_and_constant(self.nn_params, adv_class, true_class)
 
     # Computing lower and upper bounds
     # Note that lower and upper are of size nn_params.num_hidden_layers + 1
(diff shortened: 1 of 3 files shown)


cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -35,12 +38,7 @@ def __init__(self, dual_var, neural_net_param_object, test_input, true_class,
     self.input_minval = tf.convert_to_tensor(input_minval, dtype=tf.float32)
     self.input_maxval = tf.convert_to_tensor(input_maxval, dtype=tf.float32)
     self.epsilon = tf.convert_to_tensor(epsilon, dtype=tf.float32)
-    self.final_linear = (self.nn_params.final_weights[adv_class, :]
-                         - self.nn_params.final_weights[true_class, :])
-    self.final_linear = tf.reshape(self.final_linear,
-                                   shape=[tf.size(self.final_linear), 1])
-    self.final_constant = (self.nn_params.final_bias[adv_class]
-                           - self.nn_params.final_bias[true_class])
+    self.final_linear, self.final_constant = final_linear_and_constant(self.nn_params, adv_class, true_class)
```
goodfeli: This caches these graph nodes, so multiple DualFormulation objects with the same arguments don't build multiple copies of final_linear / final_constant

goodfeli: Currently line 49 fails (it creates a new graph node). I haven't yet investigated whether that line is feasible to cache.

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -66,7 +64,13 @@ def __init__(self, dual_var, neural_net_param_object, test_input, true_class,
       self.lower.append(current_lower)
       self.upper.append(current_upper)
 
+
+
     # Computing the optimization terms
+    type_check(dual_var['lambda_pos'])
```
goodfeli: For my own codebase reverse engineering purposes, I verified that these are all Variables

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -209,45 +213,76 @@ def project_dual(self):
     """Function that projects the input dual variables onto the feasible set.
 
     Returns:
-      projected_dual: Feasible dual solution corresponding to current dual
```
goodfeli: Updated the doc. This isn't actually returned

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -209,45 +213,76 @@ def project_dual(self):
...
       projected_certificate: Objective value of feasible dual
     """
-    # TODO: consider whether we can use shallow copy of the lists without
```
goodfeli: The MNIST certification example still works with this changed to the shallow copy. Haven't run all the tests though.

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -209,45 +213,76 @@ def project_dual(self):
...
+    type_check(self.lambda_pos) 
+    projected_lambda_pos = list(self.lambda_pos)
+    projected_lambda_neg = list(self.lambda_neg)
+    projected_lambda_quad = list(self.lambda_quad)
+    projected_lambda_lu = list(self.lambda_lu)
+    projected_nu = self.nu
 
     # TODO: get rid of the special case for one hidden layer
     # Different projection for 1 hidden layer
     if self.nn_params.num_hidden_layers == 1:
```
goodfeli: cached all of this if statement, so it won't get re-run

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -271,7 +307,7 @@ def project_dual(self):
 
     projected_dual_var = {'lambda_pos': projected_lambda_pos,
                           'lambda_neg': projected_lambda_neg,
-                          'lambda_lu': projected_lambda_lu,
+                          'lambda_lu': new_projected_lambda_lu,
                           'lambda_quad': projected_lambda_quad,
                           'nu': projected_nu}
```
goodfeli: I suspect that the recursive call to DualFormulation here might make it impossible to cache everything without a refactor. I was working through caching as many things as possible and seeing whether or not I run into a wall.

cleverhans/experimental/certification/optimization.py
```diff
@@ -241,6 +240,13 @@ def run_one_step(self, eig_init_vec_val, eig_num_iter_val,
     Returns:
      found_cert: True is negative certificate is found, False otherwise
     """
+
```
goodfeli: This causes a crash when the optimizer tries to change the graph inside the loop. I added it to automatically find places where we add nodes when we shouldn't. We can take it back out when we're done with the refactor.

cleverhans/experimental/certification/utils.py
```diff
@@ -45,6 +45,9 @@ def initialize_dual(neural_net_params_object, init_dual_file=None,
   lambda_quad = []
   lambda_lu = []
```
goodfeli: Here I made some changes so `nu` will be a Variable, rather than a reshape of a Variable, just so the type checking will be easier to write

PR #936, opened 2018-12-21
Change double precision to single precision, for TPU compabitility

catherio:


npapernot:
@goodfeli could you double-check? I think I recall you setting this to be float64

goodfeli:
I think this is fine to merge because:
- this is really just a sanity check that the user is calling with good arguments. you probably won't get good optimization performance if you use eps=.1 and eps_iter=.1000000000000001 but there's not real any reason that that would be a substantially worse idea than using eps=.1 and eps_iter=.1.
- we don't really have any use cases that depend on float64 precision of these variables in practice

That being said, we might actually be able to get this to run on TPU without a downcast.
I'm not familiar with writing TPU code yet. Would it work to just stick the whole thing in a "with tf.device("/CPU:0")" block so the casts run on CPU too? The asserts are already CPU-only.

catherio:
It seems easiest to me to pull this in for now, and then revisit later. I'm going to eventually want to pass over all this code to check that it's performant, and scrutinize which parts are running on CPU/TPU and whether that's appropriate.... but for now getting it to run at all is a bigger priority.

PR #937, opened 2018-12-27
'Update MNIST PyTorch Tutorial to work with PyTorch 1.0.0'

teja5832:
Made a few changes to `cleverhans_tutorials/mnist_tutorial_pytorch.py` which was otherwise running into a `TypeError` as below:

```
  File "mnist_tutorial_pytorch.py", line 184, in <module>
    tf.app.run()
  File "/nobackup/python35/lib/python3.6/site-packages/tensorflow/python/platform/app.py", line 125, in run
    _sys.exit(main(argv))
  File "mnist_tutorial_pytorch.py", line 173, in main
    learning_rate=FLAGS.learning_rate)
  File "mnist_tutorial_pytorch.py", line 110, in mnist_tutorial
    correct += (np.argmax(preds_np, axis=1) == ys).sum()
TypeError: eq() received an invalid combination of arguments - got (numpy.ndarray), but expected one of:
 * (Tensor other)
      didn't match because some of the arguments have invalid types: (numpy.ndarray)
 * (Number other)
      didn't match because some of the arguments have invalid types: (numpy.ndarray)
```

### DIFF
--- a/cleverhans_tutorials/mnist_tutorial_pytorch.py
+++ b/cleverhans_tutorials/mnist_tutorial_pytorch.py
@@ -107,7 +107,7 @@ def mnist_tutorial(nb_epochs=NB_EPOCHS, batch_size=BATCH_SIZE,
       optimizer.step()  # update gradients
 
       preds_np = preds.cpu().detach().numpy()
-      correct += (np.argmax(preds_np, axis=1) == ys.detach().numpy()).sum()
+      correct += (np.argmax(preds_np, axis=1) == ys.cpu().detach().numpy()).sum()
       total += train_loader.batch_size
       step += 1
       if total % 1000 == 0:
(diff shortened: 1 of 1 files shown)


npapernot:
@david-berthelot could you take a look when you get a chance?

teja5832:
@david-berthelot 
* I have no particular reason for `.cpu().detach)` vs. `.detach().cpu()`
*  `preds_np = preds.detach().numpy()` would work if the `detach()` automatically brings stuff into CPU - numpy does not work with GPUs automatically AFAIK: https://stackoverflow.com/questions/49605231/does-numpy-automatically-detect-and-use-gpu

david-berthelot:
Okay thanks, in light of yours answers, I assume you should replace:
`ys.detach().numpy()` with `ys.cpu().detach().numpy()`

Does that make sense?

teja5832:
@david-berthelot  I've made that change. Please let me know if there's something else.

david-berthelot:
Looks good, ready for merge.

PR #961, opened 2019-01-23
Refactored code for speedup.

shreyashankar:
* removed extraneous calls to functions
* avoided reconstructing graph nodes several times
* deleted unnecessary files

TODO:

* decompose the eigenvalue computation
* delete the testing code (random seeds, logging time)
* incorporate https://github.com/tensorflow/cleverhans/pull/957

### DIFF
--- a/cleverhans/experimental/certification/utils.py
+++ b/cleverhans/experimental/certification/utils.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import numpy as np
 import tensorflow as tf
 
 
@@ -17,6 +18,83 @@ def diag(diag_elements):
     tf matrix with diagonal entries as diag_elements
   """
   return tf.diag(tf.reshape(diag_elements, [-1]))
+
+
+def initialize_dual(neural_net_params_object, init_dual_file=None,
+                    random_init_variance=0.01, init_nu=200.0):
+  """Function to initialize the dual variables of the class.
+
+  Args:
+    neural_net_params_object: Object with the neural net weights, biases
+      and types
+    init_dual_file: Path to file containing dual variables, if the path
+      is empty, perform random initialization
+      Expects numpy dictionary with
+      lambda_pos_0, lambda_pos_1, ..
+      lambda_neg_0, lambda_neg_1, ..
+      lambda_quad_0, lambda_quad_1, ..
+      lambda_lu_0, lambda_lu_1, ..
+      random_init_variance: variance for random initialization
+    init_nu: Value to initialize nu variable with
+
+  Returns:
+    dual_var: dual variables initialized appropriately.
+  """
+  lambda_pos = []
+  lambda_neg = []
+  lambda_quad = []
+  lambda_lu = []
+
+  if init_dual_file is None:
+    for i in range(0, neural_net_params_object.num_hidden_layers + 1):
+      initializer = (np.random.uniform(0, random_init_variance, size=(
+          neural_net_params_object.sizes[i], 1))).astype(np.float32)
+      lambda_pos.append(tf.get_variable('lambda_pos_' + str(i),
+                                        initializer=initializer,
+                                        dtype=tf.float32))
+      initializer = (np.random.uniform(0, random_init_variance, size=(
+          neural_net_params_object.sizes[i], 1))).astype(np.float32)
+      lambda_neg.append(tf.get_variable('lambda_neg_' + str(i),
+                                        initializer=initializer,
+                                        dtype=tf.float32))
+      initializer = (np.random.uniform(0, random_init_variance, size=(
+          neural_net_params_object.sizes[i], 1))).astype(np.float32)
+      lambda_quad.append(tf.get_variable('lambda_quad_' + str(i),
+                                         initializer=initializer,
+                                         dtype=tf.float32))
+      initializer = (np.random.uniform(0, random_init_variance, size=(
+          neural_net_params_object.sizes[i], 1))).astype(np.float32)
+      lambda_lu.append(tf.get_variable('lambda_lu_' + str(i),
+                                       initializer=initializer,
+                                       dtype=tf.float32))
+    nu = tf.get_variable('nu', initializer=init_nu)
+    nu = tf.reshape(nu, shape=(1, 1))
+  else:
+    # Loading from file
+    dual_var_init_val = np.load(init_dual_file).item()
+    for i in range(0, neural_net_params_object.num_hidden_layers + 1):
+      lambda_pos.append(
+          tf.get_variable('lambda_pos_' + str(i),
+                          initializer=dual_var_init_val['lambda_pos'][i],
+                          dtype=tf.float32))
+      lambda_neg.append(
+          tf.get_variable('lambda_neg_' + str(i),
+                          initializer=dual_var_init_val['lambda_neg'][i],
+                          dtype=tf.float32))
+      lambda_quad.append(
+          tf.get_variable('lambda_quad_' + str(i),
+                          initializer=dual_var_init_val['lambda_quad'][i],
+                          dtype=tf.float32))
+      lambda_lu.append(
+          tf.get_variable('lambda_lu_' + str(i),
+                          initializer=dual_var_init_val['lambda_lu'][i],
+                          dtype=tf.float32))
+    nu = tf.get_variable('nu', initializer=1.0*dual_var_init_val['nu'])
+    nu = tf.reshape(nu, shape=(1, 1))
+  dual_var = {'lambda_pos': lambda_pos, 'lambda_neg': lambda_neg,
+              'lambda_quad': lambda_quad, 'lambda_lu': lambda_lu, 'nu': nu}
+  return dual_var
+
 
 def eig_one_step(current_vector, learning_rate, vector_prod_fn):
   """Function that performs one step of gd (variant) for min eigen value.


cleverhans/experimental/certification/optimization.py
```diff
@@ -171,7 +260,7 @@ def run_one_step(self, eig_init_vec_val, eig_num_iter_val,
     # Project onto feasible set of dual variables
     if self.current_step % self.params['projection_steps'] == 0:
 
-      projected_certificate = self.dual_object.project_dual()
+      projected_certificate = self.projected_dual_object.compute_certificate()
```
AlexeyKurakin: `self.projected_dual_object.compute_certificate()` will add a few more ops to the TesnorFlow graph. It means that additional ops will be added to TensorFlow graph on each step because `run_one_step` is called on every step.

Is it possible to do this only once instead of every step?

cleverhans/experimental/certification/optimization.py
```diff
@@ -181,31 +264,39 @@ def run_one_step(self, eig_init_vec_val, eig_num_iter_val,
...
+
+    [
+        _, self.current_total_objective, self.current_unconstrained_objective,
+        self.current_eig_vec_val, self.current_eig_val_estimate
+    ] = self.sess.run([
+        self.opt_one_step, self.total_objective,
+        self.dual_object.unconstrained_objective, self.eig_vec_estimate,
+        self.eig_val_estimate
+    ],
+                      feed_dict=step_feed_dict)
```
AlexeyKurakin: same as mentioned in a few other file - weird indent of argument `feed_dict`

cleverhans/experimental/certification/optimization.py
```diff
@@ -181,31 +264,39 @@ def run_one_step(self, eig_init_vec_val, eig_num_iter_val,
...
+          'min_eig_val_estimate':
+              float(self.current_eig_val_estimate)
+      }
       tf.logging.debug('Current inner step: %d, optimization stats: %s',
                        self.current_step, stats)
       if self.params['stats_folder'] is not None:
         stats = json.dumps(stats)
-        with tf.gfile.Open((self.params['stats_folder']+ '/' +
-                            str(self.current_step) + '.json'))as file_f:
+        with tf.gfile.Open((self.params['stats_folder'] + '/' + str(
```
AlexeyKurakin: to fit `with tf.gfile.Open(...)` into single line you can introduce `filename` variable and then just use `with tf.gfile.Open(filename, 'w'):`

also please use `os.path.join(a, b)` instead of `a + '/' + b`

cleverhans/experimental/certification/optimization.py
```diff
@@ -216,7 +313,7 @@ def run_optimization(self):
       True if certificate is found
       False otherwise
     """
-    self.prepare_one_step()
+    self.prepare_for_optimization()
```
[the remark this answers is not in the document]
AlexeyKurakin: What I meant is to move call to `prepare_for_optimization` into init function of `Optimization` class. This won't change how user interact with this class (`prepare_for_optimization` will be called either when class is created or when `run_optimization` is called), in both cases user will do following:
```python
opt = Optimization(...)
result =  opt.run_optimization()
```

The main reason why I was thinking to put call to self.prepare_for_optimization into `__init__` function because you already creating some ops in the `__init__` function (creating bunch of placeholders and call to `project_dual`).
It's not super important, but generally it's nice for readability to keep all the code which construct ops in one place.
Otherwise reader have to look at multiple methods to figure our what ops are being constructed.

Back to your concern, I would suggest you to think how users are supposed to use Optimization class:

* is it a one time object which is created every time you run optimization? If answer is yes, then it does not matter too much where exactly you put call to `self.prepare_for_optimization`
* Are you going to reuse `Optimization` class and call `run_optimization` several times? if yes, then probably it's a good idea to avoid constructing ops in the `run_optimization` method. Also in this case it worth thinking about how subsequent calls to `run_optimization` would differ from each other (i.e. right now no arguments are passed to `run_optimization`, so even if you call it several times, each subsequent call will do the same thing)

I'll let you think and make a decision regarding what to do

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -187,142 +189,64 @@ def get_psd_product(self, vector):
...
       else:
-        current_col_elems.append(utils.diag(self.lambda_lu[i] +
-                                            self.lambda_quad[i]))
+        current_col_elems.append(
+            utils.diag(self.lambda_lu[i] + self.lambda_quad[i]))
       if i < self.nn_params.num_hidden_layers:
-        current_col_elems.append((
-            (tf.matmul(utils.diag(-1*self.lambda_quad[i+1]),
-                       self.nn_params.weights[i]))))
+        current_col_elems.append(((tf.matmul(
```
AlexeyKurakin: too many unnecessary parenthesis, should be:
```python
current_col_elems.append(tf.matmul(
    utils.diag(-1 * self.lambda_quad[i + 1]),
    self.nn_params.weights[i]))
```

cleverhans/experimental/certification/nn.py
```diff
@@ -0,0 +1,157 @@
...
+  """NeuralNetwork is a class that interfaces the verification code with
+  the neural net parameters (weights).
+  """
+
+  def __init__(self, net_weights, net_biases, net_layer_types):
+    """Function to initialize NeuralNetParams class.
+
+    Args:
+      net_weights: list of numpy matrices of weights of each layer
+       [convention: x[i+1] = W[i] x[i]
```
AlexeyKurakin: why there is `[` in the beginning, I think it should be space instead

cleverhans/experimental/certification/nn.py
```diff
@@ -0,0 +1,157 @@
...
+        weights of layer i; 'bias_var' is the name of tf
+        variable for bias of layer i; 'is_transpose' is set to
+        True if the weights have to be transposed as per
+        convention Note that last layer is always feedforward
+        (verification operates at the layer below the final
+        softmax for more numerical stability)
+
+    Returns:
+      net_weights: list of numpy matrices of weights of each
+      layer
```
AlexeyKurakin: There are still some issues with formatting of comment:
* `layer` should have proper indentation (need to add two spaces). Also I think `layer` probably can be moved to the previous line and you still should be within line length limit
* unnecessary `[` in front of `convertion` (next line), and also this line should have proper indentation.

cleverhans/experimental/certification/nn.py
```diff
@@ -0,0 +1,157 @@
...
+    net_layer_types: type of each layer ['ff' or 'ff_relu' or 'ff_conv'
+      or 'ff_conv_relu']
+      'ff': Simple feedforward layer with no activations
+      'ff_relu': Simple feedforward layer with ReLU activations
+      'ff_conv': Convolution layer with no activation
+      'ff_conv_relu': Convolution layer with ReLU activation
+
+    Raises:
+      ValueError: If layer_types are invalid or variable names
+      not found in
```
AlexeyKurakin: please fix indentation (add two spaces), also please combine this and next line together (i.e. `not found in checkpoint` in the same line)

cleverhans/experimental/certification/optimization.py
```diff
@@ -88,75 +176,76 @@ def get_min_eig_vec_proxy(self, use_tf_eig=False):
...
+        tf.transpose(self.stopped_eig_vec_estimate),
+        self.dual_object.get_psd_product(self.stopped_eig_vec_estimate))
     # Penalizing negative of min eigen value because we want min eig value
     # to be positive
-    self.total_objective = (self.dual_object.unconstrained_objective +
-                            0.5*(tf.square(
-                                tf.maximum(-1*self.penalty_placeholder*
-                                           self.eig_val_estimate, 0))))
+    self.total_objective = (
+        self.dual_object.unconstrained_objective + 0.5 * (tf.square(
```
AlexeyKurakin: I think it would be better for readability to move `0.5*tf.square` onto next line.
Also parenthesis around tf.square are not necessary.
And usually you can replace `-1 * x` with `-x`, this will help to save few additional characters


```python
   self.total_objective = (
        self.dual_object.unconstrained_objective
        + 0.5 * tf.square(
            tf.maximum(-self.penalty_placeholder * self.eig_val_estimate, 0)))
```

cleverhans/experimental/certification/optimization.py
```diff
@@ -181,31 +268,38 @@ def run_one_step(self, eig_init_vec_val, eig_num_iter_val,
...
-          self.eig_val_estimate], feed_dict=step_feed_dict)
+    [
+        _, self.current_total_objective, self.current_unconstrained_objective,
+        self.current_eig_vec_val, self.current_eig_val_estimate
+    ] = self.sess.run([
+        self.opt_one_step, self.total_objective,
+        self.dual_object.unconstrained_objective, self.eig_vec_estimate,
+        self.eig_val_estimate
+    ],
+                      feed_dict=step_feed_dict)
```
AlexeyKurakin: feed dict is still misaligned compared to the remaining of the code

cleverhans/experimental/certification/optimization.py
```diff
@@ -181,31 +268,38 @@ def run_one_step(self, eig_init_vec_val, eig_num_iter_val,
...
+          'min_eig_val_estimate':
+              float(self.current_eig_val_estimate)
+      }
       tf.logging.debug('Current inner step: %d, optimization stats: %s',
                        self.current_step, stats)
       if self.params['stats_folder'] is not None:
         stats = json.dumps(stats)
-        with tf.gfile.Open((self.params['stats_folder']+ '/' +
-                            str(self.current_step) + '.json'))as file_f:
+        filename = os.path.join(self.params['stats_folder'], str(
```
AlexeyKurakin: maybe following would be better:
```python
filename = os.path.join(self.params['stats_folder'],
                        str(self.current_step) + '.json')
```

cleverhans/experimental/certification/utils.py
```diff
@@ -10,10 +10,8 @@
 
 def diag(diag_elements):
   """Function to create tensorflow diagonal matrix with input diagonal entries.
-
   Args:
```
AlexeyKurakin: Usually we leave blank space between function description and list of arguments, as well as between list of arguments and return value.
But it seems like most of the blank lines separating logical blocks in the comments were removed in this file

As an example of how function comments should be formatted in python, take a look at this piece of TensorFlow code: https://github.com/tensorflow/tensorflow/blob/7285f3fbddf2054c3971973f62f998579d00d8ec/tensorflow/python/layers/convolutional.py#L122

Overall idea is following:
```python
def some_function(arg1, arg2, arg3, ...):
  """Short one-line description.

  Optional multi-line description

  Args:
    arg1: description of arg1
    arg2: decsription of arg2, if you need to split this description
      into multiple lines then each subsequent line should have proper
      indentation
    arg3: ...
    ...

  Returns:
    description of return value

  Raises:
    ValueError: description in what case ValueError is raised
    SomeOtherError: ...
    ...
  """
  # function body

```

Issue #895, opened 2018-11-14
Use Keras' custom losses in mnist_tutorial_keras_tf.py


batzner:
I was reading the `mnist_tutorial_keras_tf.py` and wondered if it could be simplified by using Keras' custom losses. For example, the following code runs as fast as the tutorial in both CPU and GPU mode:

```python
model = Sequential([
    ...
])

def loss(labels, logits):
    # Compute the cross-entropy on the legitimate examples
    cross_ent = tf.losses.softmax_cross_entropy(labels, logits)

    # Compute the adversarial examples
    gradients, = tf.gradients(cross_ent, model.input)
    inputs_adv = tf.stop_gradient(model.input + 0.3 * tf.sign(gradients))

    # Compute the cross-entropy on the adversarial examples
    logits_adv = model(inputs_adv)
    cross_ent_adv = tf.losses.softmax_cross_entropy(labels, logits_adv)

    return 0.5 * cross_ent + 0.5 * cross_ent_adv

model.compile(optimizer='adam', loss=loss)
model.fit(x_train, y_train, ...)
```

Is there any downside to using custom losses that I am missing? I think it would make the tutorial a bit more accessible to Keras users, as they could stick to the `compile` and `fit` functions of Keras models. I'd be happy to submit a pull request if you think it is appropriate.

goodfeli:
That seems OK to me.

npapernot:
Thanks for looking into this. At the time where we wrote the tutorial, it was difficult to use Keras' fit function because it then required that we pass Keras' backend learning_phase to each sess.run in CleverHans (for instance when calling model_eval or batch_eval). If this is not the case anymore, it would be good to make the tutorial more Keras-like so it is more accessible to Keras users.

npapernot:
@batzner  would you still like to make a PR for this or should I close this issue for now?

batzner:
@npapernot please excuse the slow reply - I have opened a PR here: https://github.com/tensorflow/cleverhans/pull/962

PR #962, opened 2019-01-23
Keras tutorial using .fit() and .evaluate()

batzner:
As noted in https://github.com/tensorflow/cleverhans/issues/895, using CleverHans' functionality together with Keras' custom losses allows Keras users to keep using `.fit()` and `.evaluate()` with their models. Providing a tutorial showing how to do this would enable a quick start for Keras / tf.keras users. 

For example, the accuracy on adversarial examples can be implemented using a custom metric:

```python
def adv_acc(y_true, y_pred):
  # Generate adversarial examples
  x_adv = fgsm.generate(model.input, **fgsm_params)
  # Consider the attack to be constant
  x_adv = tf.stop_gradient(x_adv)

  # Accuracy on the adversarial examples
  y_adv = model(x_adv)
  return keras.metrics.categorical_accuracy(y_true, y_adv)
``` 

Training and evaluating a model using this metric thus requires minimal changes to the usual Keras workflow:

```python
model.compile(optimizer='adam', 
              loss='categorical_crossentropy', 
              metrics=['accuracy', adv_acc])
model.fit(x_train, y_train, epochs=nb_epochs, validation_data=(x_test, y_test))

# Evaluate the accuracy on legitimate and adversarial test examples
loss, acc, adv_acc = model.evaluate(x_test, y_test)
```

## Code

In `cleverhans_tutorials/mnist_tutorial_keras.py`, I modified `cleverhans_tutorials/mnist_tutorial_keras_tf.py` and replaced
- `cleverhans.train.train` with `model.fit()`
- `cleverhans.utils_tf.model_eval` with `model.evaluate()`

Since a tutorial showcasing the use of the CleverHans functions for training and evaluation is very helpful, I decided to put the new implementation into its own module for now. If you find it more appropriate to replace the existing tutorial I am happy to do so. 

`tests_tf/test_mnist_tutorial_keras.py` contains tests for the tutorial, just like `tests_tf/test_mnist_tutorial_keras_tf.py`

## Metrics

To make sure that the new implementation still achieves the same accuracies as the existing one I ran both tutorials with 8 different random seeds on a GPU. Below are the means and standard deviations of the test accuracies in the `cleverhans.utils.AccuracyReport`:

### Original Implementation
- `clean_train_clean_eval`: 99.33% ± 0.02%
- `clean_train_adv_eval`: 8.28% ± 1.53%
- `adv_train_clean_eval`: 99.07% ± 0.06%
- `adv_train_adv_eval`: 92.13% ± 0.67%

### New Implementation
- `clean_train_clean_eval`: 99.31% ± 0.06%
- `clean_train_adv_eval`: 8.57% ± 2.44%
- `adv_train_clean_eval`: 99.11% ± 0.07%
- `adv_train_adv_eval`: 92.09% ± 1.08%


npapernot:
Thank you for the PR, this looks mostly good to merge, I just have a few suggestions: 

* In the code, you refer to `y_` as the model predictions on some inputs (e.g., on adv inputs), could you change that to `preds_` to be consistent with other parts of the library? It makes it easier to distinguish labels from model predictions

* The tests are failing because of the tolerance.

* @goodfeli do you have a preference towards making this a new tutorial or updating the existing tutorial? The workflow are different enough that I see some values of having the two versions coexist (but we could also include both in the same file to limit code duplication at the expense of readability probably)

batzner:
@npapernot thanks for the quick feedback. I renamed `y_` variables to `preds_`.

The [TF 1.12 build](https://travis-ci.com/tensorflow/cleverhans/jobs/172435442#L2505) failed because Travis did not receive any output for 10 minutes and thought the build was stalled. After changing `nosetests --nologcapture` to `nosetests --nocapture`, the build passes, since Travis now also sees `print()` outputs from the tutorials. Another option would be to use `travis_wait` or to change all `print()` statements in the tutorials to `_logger.info()`. I'm happy to do that if you prefer.

The [TF 1.8 build](https://travis-ci.com/tensorflow/cleverhans/jobs/172968975#L1761) fails because of the tolerance, as you mentioned. Interestingly, it only fails when the test is called as part of `nosetests --nocapture -v --stop tests_tf;`. To verify that, I temporarily added 

`nosetests --nocapture -v --stop tests_tf/test_mnist_tutorial_keras.py;`

to the travis config, which results in the exact same accs in `report` and `report_2` ([line 1150](https://travis-ci.com/tensorflow/cleverhans/jobs/172968975#L1150) and [line 1176](https://travis-ci.com/tensorflow/cleverhans/jobs/172968975#L1176)). However, the subsequent call of 

`nosetests --nocapture -v --stop tests_tf;`

results in different accs in `report` and `report_2` ([line 1720](https://travis-ci.com/tensorflow/cleverhans/jobs/172968975#L1720) and [line 1746](https://travis-ci.com/tensorflow/cleverhans/jobs/172968975#L1746)). This is irritating, since both times the exact same code is executed and it only happens on TF 1.8. It might have to do with some code that is executed in a previous test in `tests_tf`. I hope I will be able to find the reason.

npapernot:
I believe we had a similar problem in the past. What is likely happening is that adding the new tutorial makes running tests on tutorials too long and so Travis crashes. A possibility would be to move the existing Keras tutorial to the examples folder and only keep yours in the tutorials folder. Given that keras layers are now part of tensorflow, I don't think having a proper keras model and training it with tensorflow is a common use case we need to support. Users could directly use tf.keras.layers on TF tensors if they want to use keras to define the model but train using TF.

npapernot:
Thanks I only made one change (decrease the number of epochs for the test) and will merge once tests pass again

PR #949, opened 2019-01-18
fix casting bug + add a test

goodfeli:

PR #963, opened 2019-01-23
fix casting bug + add a test

goodfeli:


goodfeli:
This is a second attempt at #949 . Not sure what went wrong with #949 but it had a lot of trouble with the build erroring and none of the other branches have that problem so I'm starting over and applying the same diff on top of a clean master.

PR #966, opened 2019-01-24
add dknn example

npapernot:


goodfeli:
Would it make sense to put this in `cleverhans.model_zoo` rather than `examples`?

npapernot:
Good suggestion, I just moved the script to the model_zoo

cleverhans/model_zoo/deep_k_nearest_neighbors/dknn.py
```diff
@@ -0,0 +1,429 @@
...
+      params_cp.lsh_family = falconn.LSHFamily.CrossPolytope
+      params_cp.distance_function = falconn.DistanceFunction.EuclideanSquared
+      params_cp.l = self.nb_tables
+      params_cp.num_rotations = 2  # for dense set it to 1; for sparse data set it to 2
+      params_cp.seed = 5721840
+      # we want to use all the available threads to set up
+      params_cp.num_setup_threads = 0
+      params_cp.storage_hash_table = falconn.StorageHashTable.BitPackedFlatHashTable
+
+      # we build 18-bit hashes so that each table has
```
carlini: Maybe worth parameterizing this if it's not meant to be MNIST/CIFAR specific

cleverhans/model_zoo/deep_k_nearest_neighbors/dknn.py
```diff
@@ -0,0 +1,429 @@
...
+          knns_ind[layer][i, :] = query_res
+        except:  # pylint: disable-msg=W0702
+          knns_ind[layer][i, :len(query_res)] = query_res
+          knn_errors += knns_ind[layer].shape[1] - len(query_res)
+
+      # Find labels of neighbors found in the training data.
+      knns_labels[layer] = np.zeros((nb_data, self.neighbors), dtype=np.int32)
+      for data_id in range(nb_data):
+        knns_labels[layer][data_id,
+                           :] = self.train_labels[knns_ind[layer][data_id]]
```
carlini: This line-split looks like it's meant for 80 characters. I believe we increased it?

cleverhans/model_zoo/deep_k_nearest_neighbors/dknn.py
```diff
@@ -0,0 +1,429 @@
...
+
+    cali_knns_not_in_class = self.nonconformity(cali_knns_labels)
+    cali_knns_not_in_l = np.zeros(self.nb_cali, dtype=np.int32)
+    for i in range(self.nb_cali):
+      cali_knns_not_in_l[i] = cali_knns_not_in_class[i, cali_labels[i]]
+    cali_knns_not_in_l_sorted = np.sort(cali_knns_not_in_l)
+    try:
+      last_zero = int(np.argwhere(cali_knns_not_in_l_sorted == 0)[-1])
+    except:  # pylint: disable-msg=W0702
+      last_zero = 0
```
carlini: How is this distinguishable from the case where the last 0 is actually at index 0 in cali_knns_not_in_l_sorted?

cleverhans/model_zoo/deep_k_nearest_neighbors/dknn.py
npapernot: Good catch, thanks

cleverhans/model_zoo/deep_k_nearest_neighbors/dknn.py
npapernot: Yes, that's a good idea

cleverhans/model_zoo/deep_k_nearest_neighbors/dknn.py
npapernot: I replaced this with a call to `np.trim_zeros`, which is probably more readable.

PR #971, opened 2019-02-06
Confidence

goodfeli:


goodfeli:
This fixes 2 things that were interfering with `make_confidence_report_spsa.py`:
1) The behavior of `safe_zip` changed from returning a list to returning an iterator when we moved from python 2 to python 3, and this broke a lot of code that called it. I changed `safe_zip` to always return a list.
2) `make_confidence_report_spsa.py` supports a batch size argument, but the argument isn't actually passed through all the way to the functions implementing the script. I made it get passed through successfully.

PR #977, opened 2019-02-09
Making certification work for CNNs

shreyashankar:
Introducing functionality to certify CNNs (tested on very small CNN)

Todo:

- Make nn test pass
- Test on Kolter + Wong CNN


shreyashankar:
Made nn tests pass

cleverhans/experimental/certification/nn.py
```diff
@@ -74,22 +126,51 @@ def forward_pass(self, vector, layer_index, is_transpose=False, is_abs=False):
...
+      vector = tf.reshape(vector, self.input_shapes[layer_index])
+      if(layer_type in {'ff', 'ff_relu'}):
+        return_vector = tf.matmul(self.weights[layer_index], vector)
+      elif (layer_type in {'conv', 'conv_relu'}):
+        return_vector = tf.nn.conv2d(vector, self.weights[layer_index],
+                                     strides=[1, STRIDE, STRIDE, 1],
+                                     padding=PADDING)
+      else:
+        raise NotImplementedError('Unsupported layer type: {0}'.format(layer_type))
+      return tf.reshape(return_vector, (self.sizes[layer_index + 1], 1))
```
AlexeyKurakin: There are few things in this function which could be improved:
* Support of `is_abs`  and `transpose` at the same time. Right now code won't work properly if both is_abs and transpose are specified
* Reduce code duplication. Currently you call tf.matmul or tf.conv in multiple times.

Few ideas on how to improve the function:
1. you store weight in some local variable and if `is_abs` true then take absolute value of the variable:
    ```
    weight = self.weights[layer_index]
    if is_abs:
      weight = tf.abs(weight)
    ```

2. Based on type of op and whether transpose is needed you choose a function to perform operation:
    ```
    if layer_type in {'ff', 'ff_relu'}:
      return_vector = tf.matmul(weight, vector, transpose_a=is_transpose)
    elif layer_type in {'conv', 'conv_relu'}:
      if is_transpose:
        ...
      else:
        ...
    return tf.reshape(return_vector, ...)
    ```

cleverhans/experimental/certification/nn.py
```diff
@@ -43,22 +47,70 @@ def __init__(self, net_weights, net_biases, net_layer_types):
...
+            [1, current_num_rows, current_num_columns, current_num_channels])
+
+        # For conv networks, unraveling the bias terms
+        small_bias = tf.convert_to_tensor(net_biases[i], dtype=tf.float32)
+        large_bias = tf.tile(tf.reshape(small_bias, [-1, 1]),
+                             [current_num_rows*current_num_columns, 1])
+        self.biases.append(large_bias)
+
+    # Last layer shape: always ff
+    if self.has_conv:
```
AlexeyKurakin: Most of the code in this if statement is duplicated, except for computation of size.
So you can refactor it in a following way:

```
if self.has_conv:
  final_dim = ...
else:
  final_dim = ...
self.sizes.append(final_dim)
...
```

cleverhans/experimental/certification/nn.py
```diff
@@ -43,22 +47,70 @@ def __init__(self, net_weights, net_biases, net_layer_types):
...
-    self.sizes.append(int(np.shape(net_weights[self.num_hidden_layers - 1])[0]))
-    self.final_weights = tf.convert_to_tensor(
-        net_weights[self.num_hidden_layers], dtype=tf.float32)
-    self.final_bias = tf.convert_to_tensor(
-        net_biases[self.num_hidden_layers], dtype=tf.float32)
+      if(self.layer_types[i] in {'ff', 'ff_relu'}):
+        self.sizes.append(int(shape[1]))
+        # For feedforward networks, no unraveling the bias terms
+
+        small_bias = tf.convert_to_tensor(net_biases[i], dtype=tf.float32)
```
AlexeyKurakin: why it's called small_bias?

cleverhans/experimental/certification/nn.py
```diff
@@ -43,22 +47,70 @@ def __init__(self, net_weights, net_biases, net_layer_types):
...
+      current_num_rows = input_shape[0]
+      current_num_columns = input_shape[1]
+      current_num_channels = input_shape[2]
+
     # Setting the sizes of the layers of the network
     # sizes[i] contains the size of x_i
     for i in range(self.num_hidden_layers):
       shape = np.shape(net_weights[i])
-      self.sizes.append(int(shape[1]))
+      # self.sizes.append(int(shape[1]))
```
AlexeyKurakin: If this code if not needed then remove it completely instead of commenting.
Same for few lines below

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -127,8 +125,9 @@ def __init__(self, sess, dual_var, neural_net_param_object, test_input, true_cla
 
     # Construct objectives, matrices, and certificate
     self.set_differentiable_objective()
-    self.get_full_psd_matrix()
-    self.construct_certificate()
+    if not self.nn_params.has_conv:
```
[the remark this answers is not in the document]
shreyashankar: I haven't implemented the conv->FF functionality that is required to make it such that you can get the full psd matrix for conv networks. Currently, with conv networks, I never explicitly compute M or H; thus I can't call `self.get_full_psd_matrix()`.

cleverhans/experimental/certification/nn.py
```diff
@@ -148,8 +230,11 @@ def load_network_from_checkpoint(checkpoint, model_json):
     # We want weights W such that x^{i+1} = W^i x^i + b^i
     # Can think of a hack involving matching shapes but if shapes are equal
     # it can be ambiguous
-    if layer_model_var['is_transpose']:
+    # TODO(shankarshreya): delete this
```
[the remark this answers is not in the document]
shreyashankar: deleted-- yes

cleverhans/experimental/certification/nn.py
shreyashankar: untiled version of bias. aditi had this

cleverhans/experimental/certification/certify.py
```diff
@@ -63,17 +63,34 @@
...
+flags.DEFINE_integer('num_columns', 28,
+                     'Number of columns in image')
+flags.DEFINE_integer('num_channels', 1,
+                     'Number of channels in image')
+flags.DEFINE_integer('stride', 2,
+                     'Stride for convolution')
+flags.DEFINE_string('padding', 'SAME',
+                    'Type of padding for convolution')
 
+dataset = 'MNIST'
```
AlexeyKurakin: dataset is not used anymore

cleverhans/experimental/certification/dual_formulation.py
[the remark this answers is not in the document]
AlexeyKurakin: what I was thinking - can you completely get rid of call `get_full_psd_matrix()`?

If you have implementation for conv without explicitely constructing M or H, then can you use the same implementation for fully connected networks?

cleverhans/experimental/certification/nn.py
```diff
@@ -11,23 +11,21 @@
...
+  def __init__(self, net_weights, net_biases, net_layer_types, input_shape=None, cnn_params=None):
     """Function to initialize NeuralNetParams class.
 
     Args:
       net_weights: list of numpy matrices of weights of each layer
        [convention: x[i+1] = W[i] x[i]
       net_biases: list of numpy arrays of biases of each layer
       net_layer_types: type of each layer ['ff' or 'ff_relu' or 'ff_conv' or
-        'ff_conv_relu']
+        'ff_conv_relu']      vector = tf.reshape(vector, self.input_shapes[layer_index])
```
AlexeyKurakin: Is `vector = tf.reshape(vector, self.input_shapes[layer_index])` added accidentally?

cleverhans/experimental/certification/nn.py
```diff
@@ -11,23 +11,21 @@
...
-STRIDE = 2
-PADDING = 'SAME'
-
 class NeuralNetwork(object):
   """NeuralNetwork is a class that interfaces the verification code with
   the neural net parameters (weights).
   """
 
-  def __init__(self, net_weights, net_biases, net_layer_types, input_shape=None):
+  def __init__(self, net_weights, net_biases, net_layer_types, input_shape=None, cnn_params=None):
```
AlexeyKurakin: please add description of cnn_params into list of arguments

cleverhans/experimental/certification/nn.py
```diff
@@ -54,15 +52,16 @@ def __init__(self, net_weights, net_biases, net_layer_types, input_shape=None):
       current_num_rows = input_shape[0]
       current_num_columns = input_shape[1]
       current_num_channels = input_shape[2]
+    if(cnn_params is not None):
```
AlexeyKurakin: The problem with this approach is that stride and padding will be the same for all convolutions in the network, which in practice not the case for real neural networks.
Padding and stride should be defined per convolution layer

Another suggestions is following.
Having tuple of values with different meaning could be not very clear for readers of the code.
Alternatives are either to use [namedtuple](https://docs.python.org/2/library/collections.html#collections.namedtuple) or use two different arguments, one for stride and one for padding.

cleverhans/experimental/certification/nn.py
```diff
@@ -130,47 +126,36 @@ def forward_pass(self, vector, layer_index, is_transpose=False, is_abs=False):
...
-      else:
-        raise NotImplementedError('Unsupported layer type: {0}'.format(layer_type))
-      return tf.reshape(return_vector, (self.sizes[layer_index + 1], 1))
+      raise NotImplementedError('Unsupported layer type: {0}'.format(layer_type))
+    if is_transpose:
+      return tf.reshape(return_vector, (self.sizes[layer_index], 1))
+    return tf.reshape(return_vector, (self.sizes[layer_index + 1], 1))
 
-def load_network_from_checkpoint(checkpoint, model_json, input_shape=None):
+def load_network_from_checkpoint(checkpoint, model_json, input_shape=None, cnn_params=None):
```
AlexeyKurakin: It would be more flexible to read cnn parameters (stride and padding) from JSON file and also make them defined individually for each layer.

Since we define format of the JSON file we can just add necessary fields (stride and padding) into conv layers.

cleverhans/experimental/certification/dual_formulation.py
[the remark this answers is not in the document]
shreyashankar: yes -- i think this is better done in another diff though, because in some cases for debugging we'd like to have `get_full_psd_matrix()`

PR #986, opened 2019-02-19
move tutorial model

npapernot:


cleverhans/model_zoo/simple_cnn.py
```diff
@@ -0,0 +1,42 @@
...
+
+import functools
+
+import tensorflow as tf
+
+from cleverhans import initializers
+from cleverhans.model import Model
+
+
+class ModelBasicCNN(Model):
```
goodfeli: Maybe rename the class SimpleCNN to match the filename?

npapernot: Good catch, I changed the filename to match the existing name to avoid confusing users of the current name.

PR #987, opened 2019-02-20
Implemented Lanczos in TF.

shreyashankar:
Instead of using scipy to compute min eig, this code uses Lanczos to compute min eig and corresponding eigenvector.

Next steps:
* Refactor compute certificate to use Lanczos
* Write tests for Lanczos


shreyashankar:
https://github.com/tensorflow/cleverhans/pull/977 should be merged first

cleverhans/experimental/certification/optimization.py
```diff
@@ -108,6 +109,61 @@ def _vector_prod_fn(x):
...
+
+  def get_lanczos_eig_vec(self):
+    """Computes the min eigen value and corresponding vector of matrix M
+    using the Lanczos algorithm.
+
+    Returns:
+      eig_vec: Corresponding eigen vector
+      eig_val: Minimum absolute eigen value
+    """
+    alpha, beta, _ = self.sess.run([self.alpha, self.beta, self.W])
```
AlexeyKurakin: computation of W does not have any side effects (i.e. changing variables, etc...), so you can just do following:
```
alpha, beta = sess.run([self.alpha, self.beta])
```

cleverhans/experimental/certification/utils.py
```diff
@@ -95,6 +95,67 @@ def initialize_dual(neural_net_params_object, init_dual_file=None,
...
+    curr_alpha = tf.reshape(tf.reduce_sum(tf.multiply(v, Q[:, i+1])), [1,])
+    alpha = tf.concat([alpha, curr_alpha], axis=0)
+    v = v-beta[-1]*Q[:, i]-alpha[-1]*Q[:, i+1]
+    curr_beta = tf.reshape(tf.norm(v), [1,])
+    beta = tf.concat([beta, curr_beta], axis=0)
+    curr_norm = tf.reshape(v/(beta[-1]+1e-8), [n, 1])
+    Q = tf.concat([Q, curr_norm], axis=1)
+
+  alpha = tf.slice(alpha, begin=[1], size=[-1])
+  return alpha, beta, Q
```
AlexeyKurakin: The first element of alpha, beta and Q is zero.
You remove if from alpha vector in previous line, however for beta and Q it's getting removed in the code which calls this function.
From the point of view of designing a good API it would be much better to remove the first element in this function rather than in the code which calls it.

cleverhans/experimental/certification/utils.py
```diff
@@ -95,6 +95,67 @@ def initialize_dual(neural_net_params_object, init_dual_file=None,
...
+  Q = tf.concat([Q, v], axis=1)
+
+  # diagonals of the tridiagonal matrix
+  beta = tf.reshape(tf.constant(0, dtype=tf.float32), [1,])
+  alpha = tf.reshape(tf.constant(0, dtype=tf.float32), [1,])
+
+  for i in range(k):
+    v = vector_prod_fn(tf.reshape(Q[:, i+1], [n, 1])) - tf.scalar_mul(scalar, tf.reshape(Q[:, i+1], [n, 1]))
+    v = tf.reshape(v, [n,])
+    curr_alpha = tf.reshape(tf.reduce_sum(tf.multiply(v, Q[:, i+1])), [1,])
```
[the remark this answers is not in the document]
shreyashankar: 'module has no attribute dot'?

PR #988, opened 2019-02-20
Allow 'y' to be np.int64 for SPSA

carlini:


goodfeli:
I suspect something breaks if you use int64, but not 100% sure.

Is there a test that's easy to tweak to exercise it in the int64 case?

Are you pretty sure the bit depth is treated as a structural argument? If you're not totally sure it's probably worth having a test that calls it twice with different bit depths and makes sure the cache contains two entries.

carlini:
I'll add some tests -- is there anything in particular you expect would break other than the structural argument?

carlini:
(The reason for this is that if you run np.argmax it returns int64 by default, so it's somewhat more convenient to have it automatically handle. If you would feel safer, I'd be fine changing the code so that if it's an int64 it automatically casts to int32.)

goodfeli:
One other thing that might go wrong is turning float32 variables into float64 somewhere downstream. This could result in some ops getting moved from GPU to CPU and causing a slowdown. Not sure if that actually happens or not. Unfortunately it's hard to catch with nosetests because Travis doesn't have GPUs.

npapernot:
@carlini what's the status here?

carlini:
Oh, forgot this PR existed. I need to write some tests to make sure this doesn't break anything unexpected -- I'll do it this week.

PR #989, opened 2019-02-21
Make all_convolutional picklable

### LINKED ISSUE #930: Make model_zoo.all_convolutional picklable
This might be as simple as just making it inherit from NoRefModel rather than Model

anant90:
Fixes #930.


npapernot:
Thank you for the PR!

PR #991, opened 2019-02-21
Moving Lanczos code to dual class.

shreyashankar:
Moving the Lanczos code to the dual class so we can use it to compute the certificate, speeding up things / eliminating use of Scipy.


cleverhans/experimental/certification/certify.py
```diff
@@ -127,6 +127,10 @@ def main(_):
         'has_conv': nn_params.has_conv,
         'lanczos_steps': FLAGS.lanczos_steps
     }
+    lzs_params = {
+        'min_iter': 5,
```
AlexeyKurakin: Make 5 named constant, i.e. `MIN_LANCZOS_ITER = 5`

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -50,6 +58,9 @@ def __init__(self, sess, dual_var, neural_net_param_object, test_input, true_cla
     self.input_minval = tf.convert_to_tensor(input_minval, dtype=tf.float32)
     self.input_maxval = tf.convert_to_tensor(input_maxval, dtype=tf.float32)
     self.epsilon = tf.convert_to_tensor(epsilon, dtype=tf.float32)
+    self.lzs_params = lzs_params
+    if not self.lzs_params:
+      self.lzs_params = DEFAULT_LZS_PARAMS
```
AlexeyKurakin: It's better to make copy of DEFAULT_LZS_PARAMS here, i.e. `self.lzs_params = DEFAULT_LZS_PARAMS.copy()`

The reason is following. DEFAULT_LZS_PARAMS is not a real constant, it's just a convention that DEFAULT_LZS_PARAMS should be unchanged. So if value of `self.lzs_params` will be changed in this class then value of DEFAULT_LZS_PARAMS also will change.
It does not happen in your code, but it's possible that change of `self.lzs_params` will happen in some future modifications, which can lead to hard-to-find bugs.
To prevent such hard-to-find bugs, it's better to make a copy of DEFAULT_LZS_PARAMS

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -50,6 +58,9 @@ def __init__(self, sess, dual_var, neural_net_param_object, test_input, true_cla
     self.input_minval = tf.convert_to_tensor(input_minval, dtype=tf.float32)
     self.input_maxval = tf.convert_to_tensor(input_maxval, dtype=tf.float32)
     self.epsilon = tf.convert_to_tensor(epsilon, dtype=tf.float32)
+    self.lzs_params = lzs_params
```
AlexeyKurakin: shorter syntax (which won't require `if` statement):
`self.lzs_params = lzs_params or DEFAULT_LZS_PARAMS.copy()`

In python `A or B` will return A if A is evaluated to true (i.e. not None and not empty) and B otherwise.

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -303,31 +357,88 @@ def get_full_psd_matrix(self):
         axis=0)
     return self.matrix_h, self.matrix_m
 
-  def compute_certificate(self):
+  def make_M_psd(self):
```
AlexeyKurakin: it's better to call this function `make_m_psd` or `make_m_matrix_psd`
Usually functions in Python are names using `all_lower_case_with_underscore' or 'CamelCase' naming convention. and naming convention should be consistent across the codebase.

Take a look at Python style naming convention as well: https://www.python.org/dev/peps/pep-0008/#naming-conventions

It looks like Cleverhans linter does not enforce naming convention very well

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -344,14 +455,30 @@ def np_vector_prod_fn_h(np_vector):
...
 
-    dual_feed_dict.update(zip(self.lambda_lu, new_lambda_lu_val))
-    dual_feed_dict.update(zip(self.lambda_neg, new_lambda_neg_val))
-    scalar_f = self.sess.run(self.scalar_f, feed_dict=dual_feed_dict)
-    vector_g = self.sess.run(self.vector_g, feed_dict=dual_feed_dict)
-    x, _ = lgmres(linear_operator_h, vector_g)
-    x = x.reshape((x.shape[0], 1))
-    second_term = np.matmul(np.transpose(vector_g), x) + 0.05
+    # Assign new lambda
+    self.sess.run([tf.assign(var, val) for var, val in zip(self.lambda_lu, new_lambda_lu_val)])
```
AlexeyKurakin: As we discussed before, temporary reassigning variables is not the best way to do it.

Better way to do it is to have a loop inside TensorFlow graph which would iterate through various values.

Please add TODO to resolve it in the future

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -303,31 +357,88 @@ def get_full_psd_matrix(self):
...
+      mid_nu = (lower_nu + upper_nu) / 2
+      self.sess.run(tf.assign(self.nu, mid_nu))
+      _, min_eig_val_m = self.get_lanczos_eig()
+      if min_eig_val_m - TOL < 0:
+        lower_nu = mid_nu
+      else:
+        upper_nu = mid_nu
+
+    # Add 0.05 to final nu to account for numerical instability
+    return original_nu, final_nu + 0.05
```
AlexeyKurakin: Does it make sense to reassign value `original_nu` to the `self.nu`?
If not please update description of the function to mention that it has side effect by modifying nu

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -303,31 +357,88 @@ def get_full_psd_matrix(self):
...
+
+  def get_lanczos_eig(self, M=True):
```
AlexeyKurakin: what is M and why default to true?
Please add description of the argument to the function comment.

Also name of the argument is very confusing, please rename it.
By looking only on the name, one can think that M is supposed to be value of the matrix, however it's apparently not. Generally you should try to make variable/argument names to be self explanatory

Also it's recommended to name variables and arguments using `lowercase_with_underscore`
See python style guide about naming: https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -344,14 +455,30 @@ def np_vector_prod_fn_h(np_vector):
...
-    second_term = np.matmul(np.transpose(vector_g), x) + 0.05
+    # Assign new lambda
+    self.sess.run([tf.assign(var, val) for var, val in zip(self.lambda_lu, new_lambda_lu_val)])
+    self.sess.run([tf.assign(var, val) for var, val in zip(self.lambda_neg, new_lambda_neg_val)])
+
+    # Make matrix M PSD
+    scalar_f = self.sess.run(self.scalar_f)
+    old_nu, second_term = self.make_M_psd()
+
+    computed_certificate = scalar_f + 0.5*(second_term)
```
AlexeyKurakin: why parenthesis around second term

cleverhans/experimental/certification/dual_formulation.py
shreyashankar: i have a comment at the bottom of compute certificate for this

cleverhans/experimental/certification/dual_formulation.py
shreyashankar: oops forgot to remove parentheses

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -344,14 +457,33 @@ def np_vector_prod_fn_h(np_vector):
...
+      if min_eig_val_m - TOL > 0:
+        tf.logging.info('Found certificate of robustness!')
+        return True
+
+    # Reset values of the variables to their original values before projection
+    # If we don't do this, the optimizer will use the projected values for variables
+    # instead of the true values
+    self.sess.run([tf.assign(var, val) for var, val in zip(self.lambda_lu, lambda_lu_val)])
+    self.sess.run([tf.assign(var, val) for var, val in zip(self.lambda_neg, lambda_neg_val)])
+    self.sess.run(tf.assign(self.nu, old_nu))
```
AlexeyKurakin: As we discussed offline, it's better to recover old value of nu inside `make_m_psd` so this function won't have side effects

PR #999, opened 2019-03-09
Update keras wrapper fprop.

HaojieYuan:
Make it compatible with combined keras model.

The context when problem could happen is here:
https://stackoverflow.com/questions/55073785/disconnected-graph-when-concatenating-two-models

To summarized, layer.ouput returns the first output node of a layer,  and when you try to combine two existing models in keras, the second model will have 2 output nodes and wrapper will give error message of disconnected graph.

By changing layer.output to layer.get_output_at(-1), we can avoid problems above, without having effect on those normal models with only 1 output node.

### DIFF
--- a/cleverhans/attacks/lbfgs.py
+++ b/cleverhans/attacks/lbfgs.py
@@ -52,6 +52,15 @@ class LBFGS(Attack):
     self.parse_params(**kwargs)
 
     _, nb_classes = self.get_or_guess_labels(x, kwargs)
+    
+    # target label unspecified will cause error while calculating
+    # softmax_cross_entropy_with_logits()
+    # here we set second most-likely category as target
+    if self.y_target is None:
+        probs = self.model.get_probs(x)
+        values, indices = tf.nn.top_k(probs, 2, sorted=True)
+        indices_2nd = indices[:, 1]
+        self.y_target = tf.one_hot(indices_2nd, probs.shape[-1])
 
     attack = LBFGS_impl(
         self.sess, x, self.model.get_logits(x), self.y_target,
--- a/cleverhans/attacks/lbfgs.py
+++ b/cleverhans/attacks/lbfgs.py
@@ -52,15 +52,6 @@ class LBFGS(Attack):
     self.parse_params(**kwargs)
 
     _, nb_classes = self.get_or_guess_labels(x, kwargs)
-    
-    # target label unspecified will cause error while calculating
-    # softmax_cross_entropy_with_logits()
-    # here we set second most-likely category as target
-    if self.y_target is None:
-        probs = self.model.get_probs(x)
-        values, indices = tf.nn.top_k(probs, 2, sorted=True)
-        indices_2nd = indices[:, 1]
-        self.y_target = tf.one_hot(indices_2nd, probs.shape[-1])
 
     attack = LBFGS_impl(
         self.sess, x, self.model.get_logits(x), self.y_target,
--- a/cleverhans/utils_keras.py
+++ b/cleverhans/utils_keras.py
@@ -191,7 +191,7 @@ class KerasModelWrapper(Model):
       new_input = self.model.get_input_at(0)
 
       # Make a new model that returns each of the layers as output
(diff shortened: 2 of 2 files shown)


npapernot:
Would the model have two heads in this case? Or are the models stacked?

If the model has two heads and only one of them is selected, it would be better to throw an exception and explain to users that the attacks require a single head. Otherwise, the default behavior could be surprising.

HaojieYuan:
The case is: we got `model1` and `model2` predefined and trained. 
Then we create a merged model by 
`model_merge = Model(inputs=model1.input, output=model2(model1.output))`
In this case the last layer of `model_merge` will have 2 output nodes and `layer.output` can not deal with it.

Is there a way to define a stacked model without introducing 2 output nodes?
If not, it would be better to throw warning messages instead exceptions.

npapernot:
Thanks for the clarification (I don't use Keras that much so I wasn't sure what you meant by combining the two models). If the two models are being stacked (i.e., the input goes through the first model and the output of the first model is used as the input of the second model, and the output of the second model is the final prediction of the stacked model). 

In that case, yes it would make sense to go with your change after you add the warning. Thanks!

npapernot:
Sorry I just realized that the fix is taking the last output of *each* layer in the model, and that could lead to potentially confusing behavior. Do all layers have multiple outputs when you stack two models or is it only for a specific layer (presumably the one which merges the two models)?

HaojieYuan:
When models are stacked,  the model stacked on others will be abstracted as 1 layer:
```
model1 = Sequential([
    Conv2D(2, (3,3), padding='same', input_shape=(6,6,1)),
    Activation('relu')
])
model2 = Sequential([
    Conv2D(3, (3,3), padding='same', input_shape=(6,6,2)),
    Activation('softmax')
])
model_merge = Model(inputs=model1.input, outputs=model2(model1.output))
model_merge.summary()
```
```
_________________________________________________________________
Layer (type)                 Output Shape              Param #   
=================================================================
conv2d_input (InputLayer)    (None, 6, 6, 1)           0         
_________________________________________________________________
conv2d (Conv2D)              (None, 6, 6, 2)           20        
_________________________________________________________________
activation (Activation)      (None, 6, 6, 2)           0         
_________________________________________________________________
sequential_1 (Sequential)    (None, 6, 6, 3)           57        
=================================================================
Total params: 77
Trainable params: 77
Non-trainable params: 0
_________________________________________________________________

```
So in this case **each** layer of model2 will have 2 outputs, because it was viewed as a single layer.
But for layers in model1, they still have 1 output.

I know set every layer's last output as default sounds risky and seems will lead to potentially confusing behavior. I will try to figure out a way to detect stacked model and only set those sequential layers' output to last node.

HaojieYuan:
Another case, if model2 was not defined by `Model()` instead of `Sequential()`
```
model1 = Sequential([
    Conv2D(2, (3,3), padding='same', input_shape=(6,6,1)),
    Activation('relu')
])

inputs = Input(shape=(6,6,2))
c1 = Conv2D(3, (3, 3), padding='same')(inputs)
a1 = Activation('softmax')(c1)
model2 = Model(inputs=inputs, outputs=a1)

model_merge = Model(inputs=model1.input,  outputs=model2(model1.output))
model_merge.summary()
```
```
_________________________________________________________________
Layer (type)                 Output Shape              Param #   
=================================================================
conv2d_input (InputLayer)    (None, 6, 6, 1)           0         
_________________________________________________________________
conv2d (Conv2D)              (None, 6, 6, 2)           20        
_________________________________________________________________
activation (Activation)      (None, 6, 6, 2)           0         
_________________________________________________________________
model (Model)                (None, 6, 6, 3)           57        
=================================================================
Total params: 77
Trainable params: 77
Non-trainable params: 0
```
The second model is also abstracted as 1 layer.
To my knowledge there are no other ways to build a keras model besides `Model()` and `Sequential()`, I will try to detect these layers.

npapernot:
Thanks for investigating this and finding a solution that only takes the last output on abstract layers

PR #1007, opened 2019-03-14
[WIP] ALP

goodfeli:


AlexeyKurakin:
One suggestion regarding this - maybe have a separate sub-directory under experimental and put code there. For example put this into "cleverhans/experimental/alp" directory.
Otherwise we might end up with adding more and more files right into experimental directory and it might become confusing

goodfeli:
Yeah, this is WIP. By the time it's ready to merge I expect not to have anything in experimental

PR #1011, opened 2019-03-21
Changing code to use TF lanczos

shreyashankar:
Remove old lanczos implementation and change code to use new implementation.

### DIFF
--- a/cleverhans/experimental/certification/dual_formulation.py
+++ b/cleverhans/experimental/certification/dual_formulation.py
@@ -154,7 +154,7 @@ class DualFormulation(object):
     """
     # Using autograph to automatically handle
     # the control flow of minimum_eigen_vector
-    self.min_eigen_vec = autograph.to_graph(utils.lanczos_decomp)
+    self.min_eigen_vec = autograph.to_graph(utils.tf_lanczos_smallest_eigval)
 
     def _m_vector_prod_fn(x):
       return self.get_psd_product(x)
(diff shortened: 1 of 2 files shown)


cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -154,35 +154,22 @@ def construct_lanczos_params(self):
...
-                                                             self.lzs_params['max_iter'])
-
-    self.alpha_h_hat, self.beta_h_hat, self.Q_h_hat = self.min_eigen_vec(_h_vector_prod_fn,
-                                                                         self.eig_max_placeholder,
-                                                                         self.matrix_m_dimension-1,
-                                                                         self.lzs_params['max_iter'])
+    self.m_min_eig, self.m_min_vec = self.min_eigen_vec(_m_vector_prod_fn,
+                                                        self.matrix_m_dimension,
+                                                        self.lzs_params['max_iter'],
+                                                        dtype=tf.float64)
```
AlexeyKurakin: is using float64 helpful here?
I think if you want to alter dtype for lanczos, it would be good to actually have it stored in the class:
```
class DualFormulation(object):

  def __init__(...):
    # ...
    self.lzs_dtype = tf.float64  # dtype for Lanczos algorithm

  def construct_lanczos_params(...):
    # ...
    self.m_min_eig, self.m_min_vec = self.min_eigen_vec(
        _m_vector_prod_fn,
        self.matrix_m_dimension,
        self.lzs_params['max_iter'],
        dtype=self.lzs_dtype)
```
This way you we can easily switch between dtypes

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -402,40 +389,23 @@ def make_m_psd(self):
...
     return original_nu, final_nu + 0.05
 
-  def get_lanczos_eig(self, compute_m=True):
+  def get_lanczos_eig(self, compute_m=True, feed_dict=None):
     """Computes the min eigen value and corresponding vector of matrix M or H
     using the Lanczos algorithm.
-
     Args:
       compute_m: boolean to determine whether we should compute eig val/vec
         for M or for H. True for M; False for H.
```
AlexeyKurakin: Please add description of feed_dict to the list of arguments in the comment

cleverhans/experimental/certification/utils.py
```diff
@@ -289,7 +222,7 @@ def tf_lanczos_smallest_eigval(vector_prod_fn,
 
   # Iteration 0 of Lanczos
   q_vectors = q_vectors.write(0, w)
-  w_ = vector_prod_fn(w)
+  w_ = tf.cast(vector_prod_fn(tf.cast(w, tf.float32)), tf.float64)
```
AlexeyKurakin: There are two issues with this code:

* It completely ignores dtype which is provided as an argument. So if won't work if the caller wanted to use float32 dtype. In particular this causes test failures
* It makes very specific assumptions about dtype used by vector_prod_fn, which makes it inconvenient to change code in the future

There are two obvious ways to fix it:

1. add another argument to this function `dtype_of_vector_prod_fn` and do casting using `dtype` and `dtype_of_vector_prod_fn`
2. don't do any casting in this function and instead rely on `vector_prod_fn` using the correct dtype

Approach 1 will create more clutter in the code.
Approach 2 is cleaner and more concise. Also if you consider API of `tf_lanczos_smallest_eigval` as a generic utility function then one would expect that caller of `tf_lanczos_smallest_eigval` has to provide both `dtype` and `vector_prod_fn` which is compatible with given dtype.

If you follow approach 2, then you remove any casting from here and you modify some of the functions `_m_vector_prod_fn` / `_h_vector_prod_fn` / `get_psd_product` / `get_h_product` in `dual_formulation` to do proper cast.

cleverhans/experimental/certification/dual_formulation.py
shreyashankar: yes float64 is more precise

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -282,11 +288,13 @@ def get_psd_product(self, vector):
       result_product: Matrix product of M and vector
     """
     # For convenience, think of x as [\alpha, \beta]
+    vector = tf.cast(vector, self.nn_dtype)
     alpha = tf.reshape(vector[0], shape=[1, 1])
     beta = vector[1:]
     # Computing the product of matrix_h with beta part of vector
     # At first layer, h is simply diagonal
     h_beta = self.get_h_product(beta)
+    h_beta = tf.cast(h_beta, self.nn_dtype)
```
AlexeyKurakin: Is this line needed? 
`self.get_h_produce(beta)` already should cast everything to `nn_dtype`

Also you can do following:
`h_beta = self.get_h_product(beta, dtype=self.nn_dtype)`

PR #1012, opened 2019-03-22
Implemented warm starts for Lanczos algorithm

shreyashankar:
Run for fewer # iterations by storing the previous eigenvector

### DIFF
--- a/cleverhans/experimental/certification/dual_formulation.py
+++ b/cleverhans/experimental/certification/dual_formulation.py
@@ -90,8 +90,8 @@ class DualFormulation(object):
       lo_plus_up = self.nn_params.forward_pass(self.lower[i] + self.upper[i], i)
       lo_minus_up = self.nn_params.forward_pass(self.lower[i] - self.upper[i], i, is_abs=True)
       up_minus_lo = self.nn_params.forward_pass(self.upper[i] - self.lower[i], i, is_abs=True)
-      current_lower = 0.5 * (lo_plus_up + lo_minus_up + self.nn_params.biases[i])
-      current_upper = 0.5 * (lo_plus_up + up_minus_lo + self.nn_params.biases[i])
+      current_lower = 0.5 * (lo_plus_up + lo_minus_up) + self.nn_params.biases[i]
+      current_upper = 0.5 * (lo_plus_up + up_minus_lo) + self.nn_params.biases[i]
       self.pre_lower.append(current_lower)
       self.pre_upper.append(current_upper)
       self.lower.append(tf.nn.relu(current_lower))
--- a/cleverhans/experimental/certification/dual_formulation.py
+++ b/cleverhans/experimental/certification/dual_formulation.py
@@ -154,7 +154,7 @@ class DualFormulation(object):
     """
     # Using autograph to automatically handle
     # the control flow of minimum_eigen_vector
-    self.min_eigen_vec = autograph.to_graph(utils.lanczos_decomp)
+    self.min_eigen_vec = autograph.to_graph(utils.tf_lanczos_smallest_eigval)
 
     def _m_vector_prod_fn(x):
       return self.get_psd_product(x)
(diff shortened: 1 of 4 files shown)


cleverhans/experimental/certification/utils.py
```diff
@@ -183,6 +183,7 @@ def minimum_eigen_vector(x, num_steps, learning_rate, vector_prod_fn):
 
 def tf_lanczos_smallest_eigval(vector_prod_fn,
                                matrix_dim,
+                               b,
```
AlexeyKurakin: rename it to something more self-explanatory, like `initial_vector`

cleverhans/experimental/certification/utils.py
```diff
@@ -216,8 +218,11 @@ def tf_lanczos_smallest_eigval(vector_prod_fn,
   q_vectors = tf.TensorArray(
       dtype, size=1, dynamic_size=True, element_shape=(matrix_dim, 1))
 
-  # Create random vector with Euclidean norm 1
-  b = tf.random_normal(shape=(matrix_dim, 1), dtype=dtype)
+  # If start vector is all zeros, make it a random normal vector and run for 1000 iter
+  if tf.reduce_sum(b) == 0:
```
AlexeyKurakin: summing up and comparing to zero could be fragile. First of all it's possible that legitimate non-zero b will sum up to 0. Also generally speaking most of floating point computations introduce errors. I'm not sure whether summing up bunch of zeros could be non zero, but better to avoid it just in case.

I would recommend you to do one of the following:

1. compare `b` to None, if it's None then you create new random vector. If it's non None then you use it directly.
   The disadvantage is that you have to call `tf_lanczos_smallest_eigval` twice, one time with b=None and another with b not None, thus you will have two subgraphs.

2. Provide boolean flag to this function and based on value of this flag re-initialize b:
  
```
def tf_lanczos_smallest_eigval(... , use_warm_start, ...):
  
  if use_warm_start == False:   # may need to try few variations of the condition to make autograph happy
    b = tf.random_normal(...)
```

Keep in mind that such `use_warm_start` flag could be Tensor of type `tf.bool`, i.e. `tf.placeholder(tf.bool, shape=())`

3. If you want to actually compare norm of `b` with something rather than having another boolean argument or None, then you should use `tf.norm(b) < TOLERANCE`.

cleverhans/experimental/certification/utils.py
```diff
@@ -216,8 +218,11 @@ def tf_lanczos_smallest_eigval(vector_prod_fn,
...
+    b = tf.random_normal(shape=(matrix_dim, 1), dtype=dtype)
+    max_iter = 1000
```
AlexeyKurakin: This value should be an argument of the function.

If you decide to go route 1 (from the comment above) then you don't need this at all.

If you decide to go route 2 (from the comment above) then you can do one of the following
* provide two arguments with number of iterations: `num_iter_full`, `num_iter_warm_start` and choose proper one inside the function
* have only one argument `max_iter`, but when calling this function make it equal to something like `tf.cond(use_warm_starts, lambda: tf.constant(100), lambda: tf.constant(1000))`

If you decide to go route 3, then probably you should have two arguments `num_iter_full`, `num_iter_warm_start`

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -425,7 +432,13 @@ def compute_certificate(self, current_step):
     lambda_neg_val = self.sess.run(self.lambda_neg)
     lambda_lu_val = self.sess.run(self.lambda_lu)
 
-    _, min_eig_val_h = self.get_lanczos_eig(compute_m=False)
+    # Use 1000 iterations for Lanczos in compute certificate
```
AlexeyKurakin: 1000 should be in self.lzs_params
We may want to adjust specific number

cleverhans/experimental/certification/utils.py
shreyashankar: I will do #3 because if the norm is anyways less than a tolerance, then we should not be using that vector

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -90,8 +90,8 @@ def __init__(self, sess, dual_var, neural_net_param_object, test_input, true_cla
       lo_plus_up = self.nn_params.forward_pass(self.lower[i] + self.upper[i], i)
       lo_minus_up = self.nn_params.forward_pass(self.lower[i] - self.upper[i], i, is_abs=True)
       up_minus_lo = self.nn_params.forward_pass(self.upper[i] - self.lower[i], i, is_abs=True)
-      current_lower = 0.5 * (lo_plus_up + lo_minus_up + self.nn_params.biases[i])
-      current_upper = 0.5 * (lo_plus_up + up_minus_lo + self.nn_params.biases[i])
+      current_lower = 0.5 * (lo_plus_up + lo_minus_up) + self.nn_params.biases[i]
+      current_upper = 0.5 * (lo_plus_up + up_minus_lo) + self.nn_params.biases[i]
```
AlexeyKurakin: What is this change about?

Is it the bug in bounds computation which you mentioned before? If yes - it's not really related to this PR and it's better to put it into separate own PR.

cleverhans/experimental/certification/utils.py
```diff
@@ -219,11 +221,11 @@ def tf_lanczos_smallest_eigval(vector_prod_fn,
       dtype, size=1, dynamic_size=True, element_shape=(matrix_dim, 1))
 
   # If start vector is all zeros, make it a random normal vector and run for 1000 iter
-  if tf.reduce_sum(b) == 0:
-    b = tf.random_normal(shape=(matrix_dim, 1), dtype=dtype)
-    max_iter = 1000
+  if tf.norm(initial_vector) < collapse_tol:
```
AlexeyKurakin: Please update comment above. It's no longer 1000 iterations, but number specified by an argument

cleverhans/experimental/certification/dual_formulation.py
shreyashankar: yes this was the bug. ok

PR #1014, opened 2019-03-22
Remove unnecessary sess.run calls in certification

shreyashankar:
Creates a projected dual object and runs compute certificate on the projected object rather than running a lot of tf.assigns.

### DIFF
--- a/cleverhans/experimental/certification/dual_formulation.py
+++ b/cleverhans/experimental/certification/dual_formulation.py
@@ -90,8 +90,8 @@ class DualFormulation(object):
       lo_plus_up = self.nn_params.forward_pass(self.lower[i] + self.upper[i], i)
       lo_minus_up = self.nn_params.forward_pass(self.lower[i] - self.upper[i], i, is_abs=True)
       up_minus_lo = self.nn_params.forward_pass(self.upper[i] - self.lower[i], i, is_abs=True)
-      current_lower = 0.5 * (lo_plus_up + lo_minus_up + self.nn_params.biases[i])
-      current_upper = 0.5 * (lo_plus_up + up_minus_lo + self.nn_params.biases[i])
+      current_lower = 0.5 * (lo_plus_up + lo_minus_up) + self.nn_params.biases[i]
+      current_upper = 0.5 * (lo_plus_up + up_minus_lo) + self.nn_params.biases[i]
       self.pre_lower.append(current_lower)
       self.pre_upper.append(current_upper)
       self.lower.append(tf.nn.relu(current_lower))
--- a/cleverhans/experimental/certification/dual_formulation.py
+++ b/cleverhans/experimental/certification/dual_formulation.py
@@ -154,7 +154,7 @@ class DualFormulation(object):
     """
     # Using autograph to automatically handle
     # the control flow of minimum_eigen_vector
-    self.min_eigen_vec = autograph.to_graph(utils.lanczos_decomp)
+    self.min_eigen_vec = autograph.to_graph(utils.tf_lanczos_smallest_eigval)
 
     def _m_vector_prod_fn(x):
       return self.get_psd_product(x)
@@ -162,27 +162,14 @@ class DualFormulation(object):
       return self.get_h_product(x)
 
     # Construct nodes for computing eigenvalue of M
-    self.alpha_m, self.beta_m, self.Q_m = self.min_eigen_vec(_m_vector_prod_fn,
-                                                             0,
-                                                             self.matrix_m_dimension,
-                                                             self.lzs_params['min_iter'])
-
-    self.eig_max_placeholder = tf.placeholder(tf.float32, shape=[])
-    self.alpha_m_hat, self.beta_m_hat, self.Q_m_hat = self.min_eigen_vec(_m_vector_prod_fn,
-                                                                         self.eig_max_placeholder,
-                                                                         self.matrix_m_dimension,
-                                                                         self.lzs_params['max_iter'])
-
-    # Construct nodes for computing eigenvalue of H
-    self.alpha_h, self.beta_h, self.Q_h = self.min_eigen_vec(_h_vector_prod_fn,
-                                                             0,
-                                                             self.matrix_m_dimension-1,
-                                                             self.lzs_params['max_iter'])
-
-    self.alpha_h_hat, self.beta_h_hat, self.Q_h_hat = self.min_eigen_vec(_h_vector_prod_fn,
-                                                                         self.eig_max_placeholder,
-                                                                         self.matrix_m_dimension-1,
-                                                                         self.lzs_params['max_iter'])
+    self.m_min_eig, self.m_min_vec = self.min_eigen_vec(_m_vector_prod_fn,
+                                                        self.matrix_m_dimension,
+                                                        self.lzs_params['max_iter'],
+                                                        dtype=tf.float64)
+    self.h_min_eig, self.h_min_vec = self.min_eigen_vec(_h_vector_prod_fn,
+                                                        self.matrix_m_dimension-1,
+                                                        self.lzs_params['max_iter'],
+                                                        dtype=tf.float64)
 
   def set_differentiable_objective(self):
     """Function that constructs minimization objective from dual variables."""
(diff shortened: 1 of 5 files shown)


cleverhans/experimental/certification/optimization.py
```diff
@@ -52,6 +56,51 @@ def __init__(self, dual_formulation_object, sess, optimization_params):
     # Create graph for optimization
     self.prepare_for_optimization()
 
+  def project_dual(self):
```
AlexeyKurakin: This function is essentially does not depend on anything declared in this class except for dual_object, and it creates another version of dual object. So it looks like DualFormulation would be a better place for this function.
In such case you slighlty shorten lines in this function (because you type `self` instead of `self.dual_object`), also you don't need to add `min_eig_val_h` (which is not dual variable) into `dual_var` dictionary:

```
class DualFormulation(object):
  
    def make_projected_dual(self):
      projected_nu = ...
      min_eig_h = ...
      projected_lambda_pos = [tf.identity(x) for x in self.lambda_pos]
      ...
      projected_dual_object = DualFormulation(...)
      projected_dual_object.min_eig_val_h = min_eig_h
      return projected_dual_object

```

cleverhans/experimental/certification/optimization.py
```diff
@@ -282,6 +326,20 @@ def run_one_step(self, eig_init_vec_val, eig_num_iter_val, smooth_val,
...
+    # Project onto feasible set of dual variables
+    if self.current_step % self.params['projection_steps'] == 0 and self.current_unconstrained_objective < 0:
+      nu = self.sess.run(self.dual_object.nu)
+      feed_dict = {
+          self.projected_dual_object.nu: nu,
+          self.dual_object.h_min_vec_ph: self.dual_object.h_min_vec_estimate,
+          self.projected_dual_object.m_min_vec_ph: self.projected_dual_object.m_min_vec_estimate
+      }
+      _, min_eig_val_h_lz = self.dual_object.get_lanczos_eig(compute_m=False, feed_dict=feed_dict)
+      feed_dict.update({self.min_eig_val_h: min_eig_val_h_lz})
```
AlexeyKurakin: I think this class does not have `min_eig_val_h` member, it's declared in `projected_dual_object`

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -426,61 +423,43 @@ def get_lanczos_eig(self, compute_m=True, feed_dict=None):
...
-    # TODO(shankarshreya): remove the assign ops and do this with new variables
-    self.sess.run([tf.assign(var, val) for var, val in zip(self.lambda_lu, new_lambda_lu_val)])
-    self.sess.run([tf.assign(var, val) for var, val in zip(self.lambda_neg, new_lambda_neg_val)])
-
-    # Make matrix M PSD
-    scalar_f = self.sess.run(self.scalar_f)
-    _, second_term = self.make_m_psd(feed_dict)
-
-    computed_certificate = scalar_f + 0.5*second_term
+    # Add 0.05 to final nu to account for numerical instability
```
AlexeyKurakin: This comment looks irrelevant to surrounding lines

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -426,61 +423,43 @@ def get_lanczos_eig(self, compute_m=True, feed_dict=None):
...
     tf.logging.info('Inner step: %d, current value of certificate: %f',
                     current_step, computed_certificate)
 
     # Sometimes due to either overflow or instability in inverses,
     # the returned certificate is large and negative -- keeping a check
-    if LOWER_CERT_BOUND < computed_certificate < -1:
-      _, min_eig_val_m = self.get_lanczos_eig()
+    if LOWER_CERT_BOUND < computed_certificate < 0:
+      _, min_eig_val_m = self.get_lanczos_eig(feed_dict=feed_dict)
+      tf.logging.info("min eig val from lanczos: " + str(min_eig_val_m))
```
AlexeyKurakin: please update quote style to use `'` to be consistent with the rest of the codebase.
Also consider removing this logging line if this logging is not needed.

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -360,52 +362,47 @@ def get_full_psd_matrix(self):
         axis=0)
     return self.matrix_h, self.matrix_m
 
-  def make_m_psd(self, feed_dict):
+  def make_m_psd(self, original_nu, feed_dict):
```
AlexeyKurakin: This and few other functions modify `feed_dict` which could be used somewhere else, ex:
```
feed_dict = { .... }
make_m_psd(nu, feed_dict)
sess.run(something_else, feed_dict=feed_dict)
```

As a result you have to track what is in feed_dict and what was modified. Also it may lead to hard to find bugs when function modified feed_dict while you didn't expect this.

So I would recommend to make a copy of feed dict if function is intended to modify it:
```
def make_m_psd(self, original_nu, feed_dict):
  feed_dict = feed_dict.copy()
  ...
```

Please check the rest of the code it applies to other functions as well

cleverhans/experimental/certification/optimization.py
```diff
@@ -282,6 +326,20 @@ def run_one_step(self, eig_init_vec_val, eig_num_iter_val, smooth_val,
...
         with tf.gfile.Open(filename) as file_f:
           file_f.write(stats)
+
+    # Project onto feasible set of dual variables
+    if self.current_step % self.params['projection_steps'] == 0 and self.current_unconstrained_objective < 0:
+      nu = self.sess.run(self.dual_object.nu)
+      feed_dict = {
+          self.projected_dual_object.nu: nu,
+          self.dual_object.h_min_vec_ph: self.dual_object.h_min_vec_estimate,
+          self.projected_dual_object.m_min_vec_ph: self.projected_dual_object.m_min_vec_estimate
```
AlexeyKurakin: it looks like `m_min_vec_estimate` is never updated and always zero in `self.projected_dual_object`

I looked at the code where `m_min_vec_estimate` is used (not in this PR) and it looks like it might better be part of `Optimization` class

cleverhans/experimental/certification/optimization.py
```diff
@@ -282,6 +326,20 @@ def run_one_step(self, eig_init_vec_val, eig_num_iter_val, smooth_val,
                                 str(self.current_step) + '.json')
         with tf.gfile.Open(filename) as file_f:
           file_f.write(stats)
+
+    # Project onto feasible set of dual variables
+    if self.current_step % self.params['projection_steps'] == 0 and self.current_unconstrained_objective < 0:
+      nu = self.sess.run(self.dual_object.nu)
+      feed_dict = {
```
AlexeyKurakin: You first compute `self.dual_object.get_lanczos_eig`, but at the same time you populate feed_dict with a bunch of stuff from `projected_dual_object`. I would suggest you to prepare two feed_dicts, one for `self.dual_object.get_lanczos_eig` which only refers to `self.dual_object` and then another one for `self.projected_dual_object.compute_certificate` which would only refer to `self.projected_dual_object`

cleverhans/experimental/certification/optimization.py
AlexeyKurakin: As we discussed offline, `m_min_vec_ph` intended to be zero in this case.
To achieve this I would suggest you to use `tf.placeholder_with_default`:

```
# in dual formulation code
class DualFormulation:

  def __init__():
    ...
    self.m_min_vec_ph = tf.placeholder_with_default(tf.zeros(...), ...)
```
Then you don't need to pass it into feed dict

cleverhans/experimental/certification/optimization.py
```diff
@@ -282,6 +326,21 @@ def run_one_step(self, eig_init_vec_val, eig_num_iter_val, smooth_val,
...
+    # Project onto feasible set of dual variables
+    if self.current_step % self.params['projection_steps'] == 0 and self.current_unconstrained_objective < 0:
+      nu = self.sess.run(self.dual_object.nu)
+      dual_feed_dict = {
+          self.dual_object.h_min_vec_ph: self.dual_object.h_min_vec_estimate
+      }
+      _, min_eig_val_h_lz = self.dual_object.get_lanczos_eig(compute_m=False, feed_dict=dual_feed_dict)
+      projected_dual_feed_dict = {
+          self.projected_dual_object.nu: nu,
+          self.project_dual.min_eig_val_h: min_eig_val_h_lz
```
AlexeyKurakin: I think it's called `self.projected_dual_object`

cleverhans/experimental/certification/dual_formulation.py
```diff
@@ -152,6 +152,55 @@ def __init__(self, sess, dual_var, neural_net_param_object, test_input, true_cla
...
+        'lambda_lu': projected_lambda_lu,
+        'lambda_quad': projected_lambda_quad,
+        'nu': projected_nu,
+    }
+    projected_dual_object = DualFormulation(
+        self.sess, projected_dual_var, self.nn_params,
+        self.test_input, self.true_class,
+        self.adv_class, self.input_minval,
+        self.input_maxval, self.epsilon,
+        self.lzs_params, False)
```
AlexeyKurakin: could you specifically spell name of the last argument, i.e.:
```
    projected_dual_object = DualFormulation(
        self.sess, projected_dual_var, self.nn_params,
        self.test_input, self.true_class,
        self.adv_class, self.input_minval,
        self.input_maxval, self.epsilon,
        self.lzs_params,
        project_dual=False)
```

Also it would be more readable if each argument of the function is on the separate line

PR #1018, opened 2019-04-04
V4 structure

npapernot:
This is the proposed structure for working on the v4 of cleverhans. When the corresponding milestone is complete, we will reorganize the repository to have the structure defined in the main README


npapernot:
cc @goodfeli

david-berthelot:
Quick question: for tutorials there are no tf2/jax/torch folders, is it intentional?

michaelshiyu:
@david-berthelot looks like there is?

npapernot:
thanks I was missing the init files that's why

Issue #1020, opened 2019-04-04
implement PGD in PyTorch


npapernot:

michaelshiyu:
@npapernot I can take this one. 

I do have a question, though. Since PyTorch is dynamic in the sense that it does not explicitly build a graph before each run, are we ok with the potential inconsistencies in the API if we go for a fully-PyTorch implementation? 

Some inconsistencies I can think of include, for example, when the user uses a PyTorch implementation, the “x” being fed into attack.generate would be the actual data in contrast to a placeholder when the user uses a TF implementation of the same attack.

Thanks.

npapernot:
We thought about this and are not expecting a unified interface across frameworks. 

Instead, we are expecting this kind of API: 

``` 
def pgd(model_fn, input, eps=0.2, iterations=1000):
  return adv_x
```

where `input` and `adv_x` could be numpy arrays. 

cc @david-berthelot who may have some thoughts as well or can correct me if I am not reflecting  properly some of the discussions we had offline

iArunava:
This is a possible implementation of the generate function.
```
    def generate(self, x, **kwargs):
        """
        Generate symbolic graph for adversarial examples and return.
        Args:
            x: The model's symbolic inputs.
            kwargs: See `parse_params`
        """

        assert self.parse_params(**kwargs)

        asserts = []

        # If a data range was specified, check that the input was in that range
        if self.clip_min is not None:
            asserts.append(x.any() >= self.clip_min)

        if self.clip_max is not None:
            asserts.append(x.any() <= self.clip_max)

        # Initialize loop variables
        if self.rand_init:
            eta = torch.FloatTensor(*x.shape).uniform_(-self.minmax, self.minmax)
        else:
            eta = torch.zeros_like(x)

        # Clip eta
        eta = clip_eta(eta, self.ord, self.eps)
        adv_x = x + eta

        if self.clip_min is not None or self.clip_max is not None:
            adv_x = torch.clamp(adv_x, self.clip_max, self.clip_min)

        if self.y_target is not None:
            y = self.y_target
            targeted = True
        elif self.y is not None:
            y = self.y
            targeted = False
        else:
            model_preds = self.model.get_probs(x)
            preds_max = reduce_max(model_preds, 1)
            y = torch.equals(model_preds, preds_max).float()
            y.requires_grad = False
            targeted = False
            del model_preds

        y_kwarg = 'y_target' if targeted else 'y'
        fgm_params = {
[...]

PR #1028, opened 2019-04-06
[WIP] Implement PGD in PyTorch

michaelshiyu:
#1020


tutorials/future/torch/mnist_tutorial.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def main(_):
+  device = 'cuda' if torch.cuda.is_available() else 'cpu'
+
+  # Load training and test data
+  data = ld_mnist()
+
+  # Instantiate model, loss, and optimizer for training
+  net = LeNet5(1, 2)
```
david-berthelot: I think we lost the code to use the GPU when available.
```
if device == 'cuda': 
  net = net.cuda()
```

tutorials/future/torch/mnist_tutorial.py
```diff
@@ -29,149 +34,81 @@ def forward(self, x):
...
+  optimizer = torch.optim.Adam(net.parameters(), lr=1e-4, weight_decay=0)
+
+  # This will be called later to report accuracy on clean and adversarial examples
+  def eval():
+    net.eval()
+    total = 0
+    total_correct = 0
+    total_correct_fgm = 0
+    total_correct_pgd = 0
+    for x, y in data.test:
```
michaelshiyu: If using GPU, should perhaps do

x, y = x.to(device), y.to(device)

Otherwise, PyTorch throws an error if the model is already on GPU.

tutorials/future/torch/mnist_tutorial.py
```diff
@@ -29,149 +34,81 @@ def forward(self, x):
...
-      x_adv_pgd = pgd.generate(
-        x,
-        y=y,
-        eps=.3,
-        eps_iter=.01,
-        nb_iter=40,
-        ord=np.inf
-        )
+  for epoch in range(FLAGS.nb_epochs):
+    for __, (x, y) in enumerate(data.train):
```
michaelshiyu: x, y = x.to(device), y.to(device)

tutorials/future/torch/mnist_tutorial.py
michaelshiyu: or just

net.to(device)

tutorials/future/torch/mnist_tutorial.py
```diff
@@ -29,149 +34,81 @@ def forward(self, x):
...
+  # Load training and test data
+  data = ld_mnist()
+
+  # Instantiate model, loss, and optimizer for training
+  net = LeNet5(1, 2)
+  loss_fn = torch.nn.CrossEntropyLoss(reduction='mean')
+  optimizer = torch.optim.Adam(net.parameters(), lr=1e-4, weight_decay=0)
+
+  # This will be called later to report accuracy on clean and adversarial examples
+  def eval():
```
michaelshiyu: Maybe put the rest of this method under

with torch.no_grad():

to prevent any gradient from being computed/saved

npapernot: This makes the attack fail

michaelshiyu: @npapernot sorry I missed the fact that you initiated the attacks in the eval method, which requires gradient computations.

cleverhans/utils_pytorch.py
```diff
@@ -183,7 +183,8 @@ def optimize_linear(grad, eps, ord=np.inf):
 
     max_abs_grad, _ = torch.max(abs_grad.view(grad.size(0), -1), 1)
     max_mask = abs_grad.eq(max_abs_grad.view(ori_shape)).to(torch.float)
-    num_ties = torch.sum(max_mask, red_ind, keepdim=True)
```
michaelshiyu: @npapernot I know this has already been merged but out of curiosity, could you please let me know what was the test case that revealed a bug in the original implementation? I've been testing with a few hand-crafted grad matrices but couldn't find a bug in the original implementation, so I guess I must have missed some corner cases. Thanks.

cleverhans/utils_pytorch.py
```diff
@@ -183,7 +183,8 @@ def optimize_linear(grad, eps, ord=np.inf):
...
+    for red_scalar in red_ind:
+      num_ties = torch.sum(num_ties, red_scalar, keepdim=True)
     optimal_perturbation = sign * max_mask / num_ties
```
michaelshiyu: @npapernot I know this has already been merged but out of curiosity, could you please let me know what was the test case that revealed a bug in the original implementation? I've been testing with a few hand-crafted grad matrices but couldn't find a bug in the original implementation, so I guess I must have missed some corner cases. Thanks.

PR #1029, opened 2019-04-10
Implementing the Noise Attack PyTorch

iArunava:
Ongoing implementation of the Noise Attack. 

- [x] Implementation


iArunava:
@npapernot can you please take a look? Thanks :)

iArunava:
Are there any relevant tests that needs to be added ? 
Also, I want to add a tutorial or just how to run this attack and since there is already a tutorial for the PGD attack, I was thinking of modifying 
https://github.com/tensorflow/cleverhans/blob/f01d21deada2f835c759323ecc58981304054c05/tutorials/future/torch/mnist_tutorial.py#L62-L64
And the take in arguments from the command line as to which attack to use and users can use the attack.

Some inputs will be valuable.! Thanks! 
P.S. I updated the noise attack API as requested. Please let me know your views.

cleverhans/future/torch/attacks/noise.py
```diff
@@ -0,0 +1,42 @@
...
+    :param clip_min: (optional float) Minimum input component value
+    :param clip_max: (optional float) Maximum input component value
+    """
+
+    if order != np.inf: raise NotImplementedError(ord)
+    
+    eta = torch.FloatTensor(*x.shape, device=x.device) \
+               .uniform_(-eps, eps)
+
+    adv_x = x.float() + eta
```
npapernot: What is the purpose of casting with `.float()` here? Is it required for the attack to work? (This question may be silly, I'm new to PyTorch)

cleverhans/future/torch/attacks/noise.py
```diff
@@ -0,0 +1,42 @@
...
+
+    if clip_min is not None and clip_max is not None:
```
npapernot: The logic is slightly different from the original attack here. Could you change the if statement to be an `or` and then assert that both clip_min and clip_max are set?

cleverhans/future/torch/attacks/noise.py
iArunava: Just in case the `x` passed is of some other type like `Byte` it will throw an error.
So, its safe to do a `.float()`
```
RuntimeError: expected backend CPU and dtype Float but got backend CPU and dtype Byte
```

cleverhans/future/torch/attacks/noise.py
iArunava: Done. Can you tell me one thing though? Where is the paper that talks about this attack?
I have played around this with attack a lot. And seems like:
- When the attack is performed on a normalized (0-1) image this attack fails.
- But when the attack is performed on not normalized (0-255) image this attack works every time. (the image is normalized and then passed into the model)
Am I getting something wrong?

And thanks for getting back :) Feels great to hear from you :)

iArunava:
Are there any relevant tests or tutorials that needs to be added along with this?

cleverhans/future/torch/attacks/noise.py
npapernot: @david-berthelot could you take a look?

cleverhans/future/torch/attacks/noise.py
npapernot: This attack is more of a test to make sure that a defense is not performing [gradient masking](http://www.cleverhans.io/security/privacy/ml/2017/02/15/why-attacking-machine-learning-is-easier-than-defending-it.html). It was recently advocated for in https://arxiv.org/abs/1802.00420 and https://openreview.net/forum?id=H1g0piA9tQ

This depends on the threat model considered, but in general the attack should not be very successful unless you run it for a large number of times.

npapernot:
This should be good to merge once the `.float()` issue has been cleared by @david-berthelot.

cleverhans/future/torch/attacks/noise.py
iArunava: Ohkay. Got it. Thanks :)

cleverhans/future/torch/attacks/noise.py
david-berthelot: I wouldn't recommending casting x to float, my personal preference is to error out rather than do some magic under the hood that can be hard to debug. Ideally I'd suggest returning a tensor of the same type as was provided in input.

iArunava: I agree with you. That's a nice point, its better to throw an Exception than to change the type of the passed input internally. I will make the changes and push asap.

iArunava: Okay! Changed. You guys were right, I designed my tests in a way such that a `ByteTensor` was sent as the input, which is actually an edge case, as most of the time the inputs will be in `float` and the `eta` is in `float` already, so there is no need to cast. Sorry, my bad.

PR #1030, opened 2019-04-11
Add Boundary Attack++

Jianbo-Lab:
Add a decision-based attack, Boundary Attack++, from the preprint: https://arxiv.org/abs/1904.02144.

### DIFF
--- a/cleverhans/attacks/bapp.py
+++ b/cleverhans/attacks/bapp.py
@@ -77,15 +77,24 @@ class BoundaryAttackPlusPlus(Attack):
           'Require a target image for targeted attack.'
       _check_first_dimension(self.image_target, 'image_target')
 
-    bapp = BAPP(self.sess, self.model, shape, self.constraint,
-                self.initial_num_evals, self.max_num_evals,
-                self.stepsize_search, self.num_iterations, self.gamma,
-                self.batch_size, self.clip_min,
-                self.clip_max, verbose=self.verbose)
+    # Set shape and d.
+    self.shape = shape
+    self.d = int(np.prod(shape))
+
+    # Set binary search threshold.
+    if self.constraint == 'l2':
+      self.theta = self.gamma / np.sqrt(self.d)
+    else:
+      self.theta = self.gamma / self.d
+
+    # Construct input placeholder and output for decision function.
+    self.input_ph = tf.placeholder(
+        tf_dtype, [None] + list(self.shape), name='input_image')
+    self.logits = self.model.get_logits(self.input_ph)
 
     def bapp_wrap(x, target_label, target_image):
       """ Wrapper to use tensors as input and output. """
-      return np.array(bapp.attack(x, target_label, target_image),
+      return np.array(self._bapp(x, target_label, target_image),
                       dtype=self.np_dtype)
 
     if self.y_target is not None:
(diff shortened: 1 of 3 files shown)


cleverhans/attacks/bapp.py
```diff
@@ -0,0 +1,573 @@
...
+    self.gamma = gamma
+    self.constraint = constraint
+    self.batch_size = batch_size
+    self.clip_min = clip_min
+    self.clip_max = clip_max
+    self.verbose = verbose
+
+
+def _check_first_dimension(x, tensor_name):
+  message = "Require {} has batch_size of 1.".format(tensor_name)
```
npapernot:
```suggestion
  message = "Tensor {} should have batch_size of 1.".format(tensor_name)
```

cleverhans/attacks/bapp.py
```diff
@@ -0,0 +1,573 @@
...
+  if x.get_shape().as_list()[0] is None:
+    check_batch = utils_tf.assert_equal(tf.shape(x)[0], 1, message=message)
+    with tf.control_dependencies([check_batch]):
+      x = tf.identity(x)
+  elif x.get_shape().as_list()[0] != 1:
+    raise ValueError(message)
+
+
+class BAPP(object):
```
npapernot: What is the reason for having two separate classes, with the first class basically (if I understand correctly) wrapping the second class?

cleverhans/attacks/bapp.py
```diff
@@ -0,0 +1,573 @@
...
+      dist = compute_distance(perturbed, sample, self.constraint)
+      if self.verbose:
+        print('iteration: {:d}, {:s} distance {:.4E}'.format(
+            j+1, self.constraint, dist))
+
+    perturbed = np.expand_dims(perturbed, 0)
+    return perturbed
+
+
+def clip_image(image, clip_min, clip_max):
```
npapernot: You can probably reuse: https://github.com/tensorflow/cleverhans/blob/f70ca7e000dadd6ace5aeff15bba0e960e8c1384/cleverhans/utils_tf.py#L540

cleverhans/attacks/bapp.py
```diff
@@ -0,0 +1,573 @@
...
+    fval = fval - np.mean(fval)
+    gradf = np.mean(fval * rv, axis=0)
+
+  # Get the gradient direction.
+  gradf = gradf / np.linalg.norm(gradf)
+
+  return gradf
+
+
+def project(original_image, perturbed_images, alphas, shape, constraint):
```
npapernot: You can probably reuse https://github.com/tensorflow/cleverhans/blob/f70ca7e000dadd6ace5aeff15bba0e960e8c1384/cleverhans/utils_tf.py#L358

cleverhans/attacks/bapp.py
```diff
@@ -0,0 +1,573 @@
...
+        perturbed_images,
+        original_image - alphas,
+        original_image + alphas
+    )
+  return projected
+
+
+def binary_search_batch(original_image, perturbed_images, decision_function,
+                        shape, constraint, theta):
+  """ Binary search to approach the boundar. """
```
npapernot:
```suggestion
  """ Binary search to approach the boundary. """
```

tests_tf/test_attacks.py
```diff
@@ -1281,3 +1283,137 @@ def test_keep_pixels_on_image(self):
...
+    self.assertTrue(np.mean(orig_labs == new_labs) < 0.1)
+
+  def test_generate_untargeted_linf(self):
+
+    x_val = np.random.rand(50, 2)
+    x_val = np.array(x_val, dtype=np.float32)
+
+    orig_labs = np.argmax(self.sess.run(self.model.get_logits(x_val)), axis=1)
+
+    # Require input has batchsize 1.
```
npapernot:
```suggestion
    # Requires input to have batchsize 1.
```

cleverhans/attacks/bapp.py
Jianbo-Lab: clip_by_value only applies to clipping by 0-dim tensor. clip_image is capable of clipping by a tensor with the same shape as the image. Also, as the wrapper is used, it might be more consistent to not use tf operators in the function.

cleverhans/attacks/bapp.py
Jianbo-Lab: 'clip_eta' is not applicable here. 'clip_eta' with ord=2 first calculates the norm of a tensor, and then divides the tensor by the norm. Thus clip_eta is equivalent to projection onto the unit sphere centered at the origin. 'project' projects a tensor to a sphere centered at a given point (not necessarily origin) with a given radius (not necessarily its norm). 'project' seems to be a more general version of 'clip_eta'.

cleverhans/attacks/bapp.py
Jianbo-Lab: I tried to follow the CleverHans code of C&W attack when a wrapper is used: https://github.com/tensorflow/cleverhans/blob/master/cleverhans/attacks/carlini_wagner_l2.py
But I agree with your suggestion that the extra class is redundant. I've removed it in the updated version. Thanks Nicolas!

npapernot:
Thanks for the PR! Feel free to add your name to the author list in the README @Jianbo-Lab

Jianbo-Lab:
Cool! I've created another PR for updating README.md. @npapernot

PR #1034, opened 2019-04-20
Semantic Attack Implementation in PyTorch

iArunava:


cleverhans/future/torch/attacks/semantic.py
```diff
@@ -0,0 +1,38 @@
+"""
+Semantic adversarial Examples
+"""
+
+from cleverhans.attacks.attack import Attack
+
+class Semantic(Attack):
```
npapernot: We are trying to get rid of classes. The code looks good but could you turn it into a simple function? Basically the content of `generate()` below. See an example signature here: https://github.com/tensorflow/cleverhans/blob/f5f92007249b6fc80315ee1bdfcc21931e568bda/cleverhans/future/torch/attacks/fast_gradient_method.py#L8

iArunava: Yes! My bad! I forgot to update this PR.

PR #1037, opened 2019-04-22
Issue 1021 PGD TF2

LorenzHW:
Implementation for #1021 
Tested with Tensorflow 2.0.0-alpha0.
Let me know what you think. I would be happy to contribute more.


cleverhans/future/tf2/attacks/fast_gradient_method.py
```diff
@@ -0,0 +1,135 @@
...
+    # We don't currently support one-sided clipping
+    assert clip_min is not None and clip_max is not None
+    adv_x = tf.clip_by_value(adv_x, clip_min, clip_max)
+
+  if sanity_checks:
+    assert np.all(asserts)
+  return adv_x
+
+
+@tf.function
```
david-berthelot: Our plan for CleverHans 4 is not to use tf.function inside the library and let the callers/users wrap the attack like in `my_attack = tf.function(pgd)` in their code.

cleverhans/future/tf2/attacks/fast_gradient_method.py
```diff
@@ -0,0 +1,135 @@
...
+def train_step(model_fn, x, y, targeted):
```
david-berthelot: Is this function really necessary or was its main purpose to enable tf.function wrapping? If it's not necessary, let's put its code inside `fast_gradient_method` otherwise let's rename it to `compute_gradients` for example.

cleverhans/future/tf2/attacks/fast_gradient_method.py
```diff
@@ -0,0 +1,135 @@
...
+  :param x: input tensor.
+  :param y: (optional) Tensor with true labels. If targeted is true, then provide the
+            target label. Otherwise, only provide this parameter if you'd like to use true
+            labels when crafting adversarial samples. Otherwise, model predictions are used
+            as labels to avoid the "label leaking" effect (explained in this paper:
+            https://arxiv.org/abs/1611.01236). Default is None.
+  :param targeted: (optional) bool. Is the attack targeted or untargeted?
+            Untargeted, the default, will try to make the label incorrect.
+  """
+  loss_fn = tf.losses.SparseCategoricalCrossentropy(from_logits=True)
```
david-berthelot: For stateless objects, we prefer to use functional APIs typically from `tf.nn.*` such as `tf.nn.softmax_cross_entropy_with_logits`.

cleverhans/future/tf2/attacks/fast_gradient_method.py
```diff
@@ -0,0 +1,135 @@
...
+  Optimal_perturbation = argmax_{eta, ||eta||_{ord} < eps} dot(eta, grad)
+
+  :param grad: tf tensor containing a batch of gradients
+  :param eps: float scalar specifying size of constraint region
+  :param ord: int specifying order of norm
+  :returns:
+    tf tensor containing optimal perturbation
+  """
+
+  # In Python 2, the `list` call in the following line is redundant / harmless.
```
david-berthelot: We are dropping Python2 support in CleverHans 4.

cleverhans/future/tf2/attacks/fast_gradient_method.py
```diff
@@ -0,0 +1,135 @@
...
+  # In Python 3, the `list` call is needed to convert the iterator returned by `range` into a list.
+  red_ind = list(range(1, len(grad.get_shape())))
+  avoid_zero_div = 1e-12
+  if ord == np.inf:
+    # Take sign of gradient
+    optimal_perturbation = tf.sign(grad)
+    # The following line should not change the numerical results.
+    # It applies only because `optimal_perturbation` is the output of
```
david-berthelot: We moved from 80 to 120 characters per line.

cleverhans/future/tf2/attacks/fast_gradient_method.py
LorenzHW: I changed it. However, if I use `tf.function(pgd)` with adversarial training in the tutorial it gets really slow. It is slower than without `tf.function`.

david-berthelot: Interesting, in that case since there is a performance penalty, I would recommend to put back the tf.function decorator like you initially did.

cleverhans/future/tf2/attacks/__init__.py
```diff
@@ -0,0 +1,2 @@
+from cleverhans.future.tf2.attacks.fast_gradient_method import fast_gradient_method
+from cleverhans.future.tf2.attacks.projected_gradient_descent import projected_gradient_descent
```
david-berthelot: May be put a newline at the end of the file.

cleverhans/future/tf2/utils_tf.py
```diff
@@ -0,0 +1,722 @@
+"""Utility functions for writing TensorFlow code"""
```
david-berthelot: I didn't realize this enormous file was added. From the look of it, it seems that you copied an existing file into future. I would say keep utils_tf.py to its bare minimum (we thought several times of not having a utils file since ends collecting all kind of unrelated stuff), like remove all the coded that's not needed by your current functions.

cleverhans/future/tf2/attacks/fast_gradient_method.py
LorenzHW: Ok, I moved it into a separate function again. Maybe let's try it again when the final version of TF2.0 is released.

PR #1043, opened 2019-04-25
fix corner cases

Jianbo-Lab:
Fix two corner cases:
1. Add clip after update to avoid accidental large step sizes in the initial steps.
2. Add assertion for the case where random initialization fails.

Use Sign of gradient for L-inf Boundary Attack++.


Jianbo-Lab:
As I git commit from a remote machine, it seems the Contributor License didn't recognize me. I will close the pull request and create a duplicate one.

PR #1049, opened 2019-05-08
Add tutorial of CIFAR10 in TF2

### LINKED ISSUE #306: add CIFAR10 tutorial in TF2
This might require making some changes to the library to support this new mode.

souravsingh:
Fixes #306

### DIFF
--- a/tutorials/future/tf2/cifar10_tutorial.py
+++ b/tutorials/future/tf2/cifar10_tutorial.py
@@ -1,0 +1,76 @@
+import numpy as np
+import tensorflow as tf
+import tensorflow_datasets as tfds
+from absl import app, flags
+from easydict import EasyDict
+from tensorflow.keras import Model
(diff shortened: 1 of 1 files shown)


npapernot:
What's the accuracy before and after running the attack? The epsilon seems high (at least the default value of 0.3)

souravsingh:
I will have to check, but I believe accuracy before the attack was around 75%

npapernot:
This seems quite low. @david-berthelot do you think we could use your all convolutional model to get a better baseline?

david-berthelot:
Yes reaching more than 90% should be relatively easy even with a small number of filters (say 16 or 32 for runtime reasons) for the all convolutional model already present in the current CleverHans. The code can be found here:
https://github.com/tensorflow/cleverhans/blob/master/cleverhans/model_zoo/all_convolutional.py

npapernot:
I'm closing this PR because the code does not run (it is incomplete and needs to much work to be completed so starting from scratch in #1076)

PR #1053, opened 2019-05-11
Deepfool attack in PyTorch

iArunava:


iArunava:
@npapernot can you please check it once? I have tested it in my local, and the attack works flawlessly. Although, it doesn't yet work for batches of images.

cleverhans/future/torch/attacks/deepfool.py
```diff
@@ -0,0 +1,86 @@
...
+      The input tensor
+  net : nn.Module
+        The net.
+  eta : float
+        Eta as used in the paper.
+  tnc : int
+       The number of classes to consider.
+  miter : int
+          The maximum number of iterations.
+  mret : bool
```
iArunava: DeepFool also says that the total perturbation it causes to an image can be used to measure the robustness of the model, so I have allowed this extra parameter `mret` which if true (defaults to `False`) returns that metric along with the adversarial image.

cleverhans/future/torch/attacks/deepfool.py
```diff
@@ -0,0 +1,86 @@
...
+import numpy as np
+import copy
+
+from torch.autograd.gradcheck import zero_gradients as zgrad
+
+
+__all__ = ['deepfool']
+
+
+def deepfool(x, net:nn.Module, eta:float=0.02, tnc:int=10, miter:int=50,
```
npapernot: For the other attacks we thought about specifying types but then decided to remain pythonic and leave it to the user to decide whether they want to pass a float or a tensor containing `eta` for instance.

cleverhans/future/torch/attacks/deepfool.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def deepfool(x, net:nn.Module, eta:float=0.02, tnc:int=10, miter:int=50, 
+             mret=False):
+  """
+  Implementation of the DeepFool Attack.
+
+  Arguments
+  ---------
+  x : torch.Tensor
```
npapernot: Why are the comments split over multiple lines?

cleverhans/future/torch/attacks/deepfool.py
npapernot: Empirically measuring robustness in this way is not specific to deepfool so I think to remain consistent with other attacks, we'll only return the adversarial example and if users want to measure robustness in this way they can diff the original and adversarial inputs

npapernot:
The authors pointed out to me in private correspondence that their own implementation is in PyTorch, which I did not know. It makes sense to at least compare how the attacks perform on one or two datasets and perhaps to even follow their original implementation (with the appropriate credit in the docstring)

iArunava:
Hi! I am aware that the original implementation of DeepFool is in PyTorch (and in Matlab). I have referred to their implementation have checked that everything is according. About running on 2 datasets. I have checked it on MNIST it gets an accuracy of 1.2% on the test set of MNIST.
I will check it on ILSVRC 2012 (else CIFAR10 if Colab takes too much time) and let you know.

Also, the original implementation is not vectorized, it can't take batches of images. For MNIST on Lenet it takes 2hours on Colab. I have vectorized the entire implementation [here](https://github.com/iArunava/scratchai/blob/master/scratchai/attacks/attacks/deepfool.py) You may check it out. With the vectorized implementation it takes about 1-2 mins.

Also, I will go through your entire review and make relevant updates accordingly as soon as possible.

P.S. I have contacted Seyed myself and let him know about my implementation but haven't received feedback yet.

npapernot:
Thanks for the contribution @iArunava : did you get a chance to wrap up this PR?

iArunava:
Oh my, I am really sorry for the delay, I am caught up with a few projects. I will try to get this to shape as soon as possible. Thanks for understanding.

npapernot:
Ping @iArunava ?

jonasguan:
Closing this PR for now due to inactivity, since we're undergoing a major revamp. Happy to reopen if it's updated later.

PR #1054, opened 2019-05-13
train script to construct imperceptible adversarial examples

yaq007:
This is the code for ICML 2019 "Imperceptible, Robust and Targeted Adversarial Examples for Automatic Speech Recognition."


yaq007:
To whom it may concern,

I signed it. Please verify it.

Best,
Yao

PR #1058, opened 2019-06-01
change name, stepsize, default search, add author

Jianbo-Lab:
@npapernot Please merge the changes on the previous Boundary Attack++, resulting from changes and improvements in the preprint. The changes in the code are
1. Following suggestions from my collaborators, we decided to change the name of Boundary Attack++ to HopSkipJumpAttack. 
2. A new author, Martin J. Wainwright, is added to the paper.
3. Update the choice of stepsize based on a newly established theorem. 
4. Change default stepsize search to "geometric progression".


npapernot:
Thanks, this looks good. Could you add a deprecation warning of 6 months for the old attack name and have it call the new attack class so that the code is backwards compatible?

Jianbo-Lab:
I have added a new function named as BoundaryAttackPlusPlus, calling the class HopSkipJumpAttack with an extra warning. Does it meet the requirements?

PR #1076, opened 2019-06-27
make improvements to TF2 tutorial on MNIST and add tutorial on CIFAR10

npapernot:


npapernot:
Now also addressing #306

PR #1079, opened 2019-06-28
JAX PGD support

npapernot:
Works for L_inf and L_2


npapernot:
Closes #1019

cleverhans/future/jax/attacks/projected_gradient_descent.py
```diff
@@ -0,0 +1,80 @@
...
+  if clip_min is not None or clip_max is not None:
+    adv_x = np.clip(adv_x, a_min=clip_min, a_max=clip_max)
+
+  if y is None:
+    # Using model predictions as ground truth to avoid label leaking
+    x_labels = np.argmax(model_fn(x), 1)
+    y = one_hot(x_labels, 10)
+
+  i = 0
+  while i < nb_iter:
```
david-berthelot: Why not simply doing?
```
for i in range(nb_iter):
   ...
```

npapernot: Good catch!

Issue #1077, opened 2019-06-27
Rename ord argument to norm since ord is a reserved keyword for Python


david-berthelot:

michaelshiyu:
Can I take this issue if nobody else is working on it?

PR #1080, opened 2019-06-29
[MRG] Change ord to norm

michaelshiyu:
#1077


michaelshiyu:
Hi @npapernot, could you please review this PR? Many thanks.

npapernot:
My bad the title of the issue was ambiguous, could you only change the name of the argument for the 3 new languages supported in cleverhans v4? (jax eager and pytorch) 

We don't need to update the existing APIs that will get deprecated when we increment the version number to v4. Basically, the only files that need to be modified are those in the `future` folders. 

Let me know if you need more details!

michaelshiyu:
Gotcha! Thanks for the details, @npapernot! 

Just to make sure, the three new languages you mentioned are jax, tf2, and pytorch, right? Thanks.

npapernot:
Yes, all of their code should be in future folders

michaelshiyu:
> Yes, all of their code should be in future folders

Hi @npapernot, I fixed the PR accordingly. In addition, I did some refactoring and moved all future-related pytorch code into ```future```. Could you please take another look at this PR? Many thanks.

PR #1083, opened 2019-07-02
Implementing two attacks using tensorflow 2.0

MyRespect:
1. adding CarliniWagnerL2 attack using tensorflow2.0 according to the authors' original source code.
2. adding DeepFool attack using tensorflow2.0 according to the authors' original source code.

### DIFF
--- a/cleverhans/future/tf2/attacks/carlini_wagner_l2.py
+++ b/cleverhans/future/tf2/attacks/carlini_wagner_l2.py
@@ -1,0 +1,210 @@
+""" The CarliniWagnerL2 attack """
+
+import numpy as np
+import tensorflow as tf
+from tensorflow.keras.models import Sequential
+from tensorflow.keras import datasets, optimizers
+
+BINARY_SEARCH_STEPS = 9  # number of times to adjust the constant with binary search
+MAX_ITERATIONS = 1000   # number of iterations to perform gradient descent
+ABORT_EARLY = True       # if we stop improving, abort gradient descent early
+LEARNING_RATE = 1e-2     # larger values converge faster to less accurate results
+TARGETED = True          # should we target one specific class? or just be wrong?
+CONFIDENCE = 0           # how strong the adversarial example should be
+INITIAL_CONST = 1e-3     # the initial constant c to pick as a first guess
+
+
+class CarliniL2:
+  """
+  This attack was originally proposed by Carlini and Wagner. It is an
+  iterative attack that finds adversarial examples on many defenses that
+  are robust to other attacks.
+  Paper link: https://arxiv.org/abs/1608.04644
+
+  At a high level, this attack is an iterative attack using Adam and
+  a specially-chosen loss function to find adversarial examples with
+  lower distortion than other attacks. This comes at the cost of speed,
+  as this attack is often much slower than others.
+
+  :param model: network (input: images, output: values of activation **BEFORE** softmax).
+  :param shape: (width, height, channels)
+  :param num_classes: default 10
+
+  Returns adversarial examples for the supplied model.
+
+  :param confidence: Confidence of adversarial examples: higher produces examples
+              that are farther away, but more strongly classified as adversarial.
+  :param batch_size: Number of attacks to run simultaneously.
+  :param targeted: True if we should perform a targetted attack, False otherwise.
+  :param learning_rate: The learning rate for the attack algorithm. Smaller values
+              produce better results but are slower to converge.
+  :param binary_search_steps: The number of times we perform binary search to
+              find the optimal tradeoff-constant between distance and confidence. 
+  :param max_iterations: The maximum number of iterations. Larger values are more
+              accurate; setting too small will require a large learning rate and will
+              produce poor results.
+  :param abort_early: If true, allows early aborts if gradient descent gets stuck.
+  :param initial_const: The initial tradeoff-constant to use to tune the relative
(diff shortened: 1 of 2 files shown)


cleverhans/future/tf2/attacks/carlini_wagner_l2.py
```diff
@@ -0,0 +1,210 @@
...
+BINARY_SEARCH_STEPS = 9  # number of times to adjust the constant with binary search
+MAX_ITERATIONS = 1000   # number of iterations to perform gradient descent
+ABORT_EARLY = True       # if we stop improving, abort gradient descent early
+LEARNING_RATE = 1e-2     # larger values converge faster to less accurate results
+TARGETED = True          # should we target one specific class? or just be wrong?
+CONFIDENCE = 0           # how strong the adversarial example should be
+INITIAL_CONST = 1e-3     # the initial constant c to pick as a first guess
+
+
+class CarliniL2:
```
carlini: More correct to rename to CarliniWagnerL2

cleverhans/future/tf2/attacks/deepfool.py
```diff
@@ -0,0 +1,84 @@
...
+
+      if pert_k < pert:
+        pert = pert_k
+        w = w_k
+
+    r_i = (pert + 1e-4) * w / np.linalg.norm(w)
+    r_tot = np.float32(r_tot + r_i)
+    pert_image = image_norm + (1 + overshoot) * r_tot
+
+    x = tf.Variable(pert_image)
```
carlini: This creates a new Variable inside of a loop up to a large number of iterations. Non-rhetorical question: will this potentially cause out-of-memory errors if run for long enough?

cleverhans/future/tf2/attacks/deepfool.py
```diff
@@ -0,0 +1,84 @@
+""" The DeepFool attack """
+
+import copy
+import numpy as np
+import tensorflow as tf
+
+
+def deepfool(image_norm, model, num_classes=10, overshoot=0.02, max_iter=50):
```
carlini: High-level comment: it would be helpful to add some comments to the code to describe what it's doing.

cleverhans/future/tf2/attacks/carlini_wagner_l2.py
```diff
@@ -0,0 +1,210 @@
...
+          # if targetted, optimize for making the other class most likely
+          loss1 = tf.maximum(0.0, other - real + self.CONFIDENCE)
+        else:
+          # if untargeted, optimize for making this class least likely.
+          loss1 = tf.maximum(0.0, real - other + self.CONFIDENCE)
+
+        loss2 = tf.reduce_sum(l2dist)
+        loss1 = tf.reduce_sum(const * loss1)
+        loss = loss1 + loss2
+      optimizer = optimizers.Adam(self.LEARNING_RATE)
```
jvmancuso: not entirely familiar with the new OptimizerV2, but I think this line will create a new optimizer at each train step, which would be a bug

jvmancuso: confirmed with this script:
```
import tensorflow as tf

def train_step(modifier, i):
  x = tf.constant([1., 2, 3], shape=(1,3), dtype=tf.float32)
  c = tf.reduce_sum(10 * x)
  with tf.GradientTape() as tape:
    y = x @ modifier
    loss = tf.norm(c - y)
  optimizer = tf.keras.optimizers.Adam(1)
  grad = tape.gradient(loss, [modifier])
  if i % 5 == 4:
    print(optimizer.iterations)
  optimizer.apply_gradients(zip(grad, [modifier]))
  mmm = optimizer.weights

  return loss, mmm


mod = tf.Variable(tf.ones(shape=(3,1), dtype=tf.float32))
for i in range(15):
  l, m = train_step(mod, i)
```
which returns:
```
<tf.Variable 'iter:0' shape=() dtype=int64, numpy=0>
<tf.Variable 'iter:0' shape=() dtype=int64, numpy=0>
<tf.Variable 'iter:0' shape=() dtype=int64, numpy=0>
```

cleverhans/future/tf2/attacks/carlini_wagner_l2.py
```diff
@@ -0,0 +1,210 @@
...
+      if self.TARGETED:
+        return x == y
+      else:
+        return x != y
+
+    # @tf.function # there are some problems to be fixed, open issues in Tensorflow2.0 beta
+    def train_step(modifier, timg, tlab, const):
+      with tf.GradientTape() as tape:
+        newimg = tf.tanh(modifier + timg) * self.boxmul + self.boxplus
+        output = model.predict(newimg)
```
jvmancuso: the object `model` will not exist here -- it needs to be stashed as an attribute during the `__init__` and referenced as an attribute here (or referenced some other way after a refactor)

npapernot:
Thanks for the feedback @jvmncs

jvmncs:
@npapernot no problem. I actually have a fresh TF 2.0 implementation of this on my machine. It's similar to the original 1.x code, but fully vectorized and stays in-graph as much as possible (although I've only tested it with eager mode so far). If there's interest, I can open a PR for review.

npapernot:
That would be great yes!

npapernot:
@jvmncs would you be able to make the pull request?

PR #1086, opened 2019-07-11
Implement SPSA in PyTorch

cassidylaidlaw:
I mostly translated from the Tensorflow code and implemented a similar interface. I also added tests based on the existing PyTorch tests.

### DIFF
--- a/cleverhans/future/torch/attacks/spsa.py
+++ b/cleverhans/future/torch/attacks/spsa.py
@@ -110,7 +110,7 @@ def spsa(model_fn, x, eps, nb_iter, clip_min=-np.inf, clip_max=np.inf, y=None,
     if early_stop_loss_threshold is not None and loss < early_stop_loss_threshold:
       break
 
-  adv_x = (x + perturbation).detach()
+  adv_x = torch.clamp((x + perturbation).detach(), clip_min, clip_max)
 
   asserts.append(torch.all(torch.abs(adv_x - x) <= eps + 1e-6))
   asserts.append(torch.all(adv_x >= clip_min))
--- a/cleverhans/future/torch/attacks/spsa.py
+++ b/cleverhans/future/torch/attacks/spsa.py
@@ -1,0 +1,187 @@
+"""The SPSA attack."""
+import numpy as np
+import torch
+from torch import optim
+
+
+def spsa(model_fn, x, eps, nb_iter, clip_min=-np.inf, clip_max=np.inf, y=None,
+         targeted=False, early_stop_loss_threshold=None, learning_rate=0.01, delta=0.01,
+         spsa_samples=128, spsa_iters=1, is_debug=False, sanity_checks=True):
+  """
+  This implements the SPSA adversary, as in https://arxiv.org/abs/1802.05666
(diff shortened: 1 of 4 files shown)


cleverhans/future/torch/attacks/spsa.py
```diff
@@ -0,0 +1,187 @@
...
+  asserts = []
+
+  # If a data range was specified, check that the input was in that range
+  asserts.append(torch.all(x >= clip_min))
+  asserts.append(torch.all(x <= clip_max))
+
+  if is_debug:
+    print("Starting SPSA attack with eps = {}".format(eps))
+
+  perturbation = (torch.rand_like(x) * 2 - 1) * eps
```
npapernot: Does this imply the attack assumes that the inputs are centered around 0?

cleverhans/future/torch/attacks/spsa.py
```diff
@@ -0,0 +1,187 @@
...
+  asserts.append(torch.all(adv_x >= clip_min))
+  asserts.append(torch.all(adv_x <= clip_max))
+
+  if sanity_checks:
+    assert np.all(asserts)
+
+  return adv_x
+
+
+def _project_perturbation(perturbation, epsilon, input_image, clip_min=-np.inf,
```
npapernot: This assumes the adversary uses an L_inf norm. Is it easy to generalize the implementation to L_inf and L2?

It would also make sense to reuse the function here to compute `clipped_perturbation` : https://github.com/tensorflow/cleverhans/blob/921cbdb73bb625c36bf749e4f2bda4f017ec8f09/cleverhans/future/torch/utils.py#L8

cleverhans/future/torch/attacks/spsa.py
cassidylaidlaw: No, because perturbation is eventually added to x (see line 113)

cleverhans/future/torch/attacks/spsa.py
cassidylaidlaw: The original implementation uses an L_inf threat model, but it should be easy to generalize to L2 as well. I'll work on it!

npapernot: Excellent, thanks! If this is simpler, we can merge the current PR and then make a follow-up adding L2 support. If that's better for you, it would be great to use the `clip_eta` util function and then I'll merge this PR. We can also wait for L2

npapernot:
Do you want us to merge this as a L_inf implementation and later make a separate PR for other norms?

cassidylaidlaw:
I switched to use `clip_eta` so it should support norms other than L_inf. I kept the default as L_inf since that's what the original paper used.

One other question—should I make the learning_rate parameter consistent with the other implementations, even if they're wrong? Jonathan Uesato mentioned I should add a "lr" parameter that is correct but keep learning_rate so that existing code continues to work as is. This implementation currently uses a learning_rate parameter that is correct, but I could switch it to be consistent across implementations

adam-dziedzic:
Is this code ready? Can it be merged?

rumschuettel:
I would be interested in this as well. Can it be merged? :)

PR #1088, opened 2019-07-12
Bug fix in random noise attack

shoaibahmed:
The previous legacy device function call was only applicable for the CPU device. Updated according to the most recent API.


npapernot:
Thanks for the PR! Would you be able to apply the fix to other PyTorch related files as well?

shoaibahmed:
> Thanks for the PR! Would you be able to apply the fix to other PyTorch related files as well?

Other function calls are still applicable to the latest version. Only the noise attack was not working with a CUDA enabled device. Let me check if the other ones are using the latest interface.

shoaibahmed:
> Thanks for the PR! Would you be able to apply the fix to other PyTorch related files as well?

I have checked. All other PyTorch related files in future are fine at this point.

PR #1090, opened 2019-07-17
Implement DeepFool in PyTorch

cassidylaidlaw:
I know somebody else has an open PR for DeepFool in PyTorch but I think my implementation is better in a few ways:
 * It correctly handles stopping on batches—the algorithm is supposed to stop as soon as it reaches misclassification for each sample separately, and this implementation does that.
 * It doesn't use NumPy, so everything can take place on the GPU and data doesn't have to be copied back and forth to the CPU, resulting in a big speedup.
 * It's tested against the original implementation by the authors and performs exactly the same on a nontrivial network with five classes.
 * The arguments more closely mirror other attacks in the library.
Let me know what you think!

### DIFF
--- a/cleverhans/future/torch/attacks/deepfool.py
+++ b/cleverhans/future/torch/attacks/deepfool.py
@@ -1,0 +1,161 @@
+"""The Deepfool attack."""
+import numpy as np
+import torch
+from torch.autograd.gradcheck import zero_gradients
+from cleverhans.future.torch.utils import clip_eta
+
+
+def deepfool(model_fn, x, clip_min=-np.inf, clip_max=np.inf,
+             y=None, targeted=False, eps=None, norm=None,
+             num_classes=10, overshoot=0.02, max_iter=50,
+             is_debug=True, sanity_checks=False):
+  """
+  PyTorch implementation of DeepFool (https://arxiv.org/pdf/1511.04599.pdf).
+  :param model_fn: A callable that takes an input tensor and returns the model logits.
+  :param x: Input tensor.
+  :param clip_min: If specified, the minimum input value.
+  :param clip_max: If specified, the maximum input value.
+  :param y: (optional) Tensor with true labels. If targeted is true, then provide the
+            target label. Otherwise, only provide this parameter if you'd like to use true
+            labels when crafting adversarial samples. Otherwise, model predictions are used
+            as labels to avoid the "label leaking" effect (explained in this paper:
+            https://arxiv.org/abs/1611.01236). Default is None.
+  :param targeted: (optional) bool. Is the attack targeted or untargeted? Untargeted, the
+            default, will try to make the label incorrect. Targeted will instead try to
+            move in the direction of being more like y.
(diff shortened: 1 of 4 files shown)


npapernot:
@iArunava what do you think?

cleverhans/future/torch/attacks/deepfool.py
```diff
@@ -0,0 +1,175 @@
...
+    all_perturbation_updates[live] = smallest_perturbation_updates
+    perturbations.add_(all_perturbation_updates)
+
+    perturbations *= (1 + overshoot)
+    if eps is not None:
+      perturbations = clip_eta(perturbations, norm, eps)
+    perturbations = torch.clamp(x + perturbations, clip_min, clip_max) - x
+    perturbations /= (1 + overshoot)
+
+#   perturbations *= (1 + overshoot)
```
npapernot: Why is this code commented?

cleverhans/future/torch/utils.py
```diff
@@ -21,22 +21,7 @@ def clip_eta(eta, norm, eps):
...
-      norm = torch.sqrt(torch.max(
-          avoid_zero_div,
-          torch.sum(eta ** 2, dim=reduc_ind, keepdim=True)
-      ))
-    factor = torch.min(
-        torch.tensor(1., dtype=eta.dtype, device=eta.device),
-        eps / norm
-        )
-    eta *= factor
+    eta = torch.renorm(eta, p=norm, dim=0, maxnorm=eps)
```
npapernot: Does this avoid div by zero? This is important for numerical stability in our case.

cleverhans/future/torch/attacks/deepfool.py
```diff
@@ -0,0 +1,175 @@
+"""The Deepfool attack."""
+import numpy as np
+import torch
+from torch.autograd.gradcheck import zero_gradients
+from cleverhans.future.torch.utils import clip_eta
+
+
+def deepfool(model_fn, x, clip_min=-np.inf, clip_max=np.inf,
```
npapernot: To be clear, are the advantages of this implementation that it can handle batches of inputs and early stop if all batch inputs have been misclassified?

cassidylaidlaw: It does early stopping separately for each element of the batch, which mirrors the original DeepFool implementation. Like the first element of the batch could be early stopped after 2 iterations and the second could be early stopped after 10. I think the other implementation would keep iterating the whole batch until all of them had succeeded.

Other than that I think the pure torch implementation is a plus for being a lot faster on GPU.

cleverhans/future/torch/attacks/deepfool.py
cassidylaidlaw: Sorry, that shouldn't be there actually. I'll remove it

cleverhans/future/torch/utils.py
cassidylaidlaw: In what case do you imagine div by zero could be an issue? I've tested and it works fine if the norm of eta is zero and fine if eps is 0

PR #1106, opened 2019-09-10
Update utils_keras.py

LaurensonR:
Added a VGG-type CNN baseline model that has increased resiliency to two categories of adversarial attacks compared to the existing baseline CNN model.


LaurensonR:
I signed it

PR #1127, opened 2019-10-22
Correct the code of confidence formulation in DkNN

Buhua-Liu:


Buhua-Liu:
@googlebot I signed it!

PR #1137, opened 2019-12-19
compute average distortion

BUB97:
I think that when computing the average distortion, we shouldn't include the targets we didn't generate. (For instance, for a sample image of number 9, we didn't produce target image of class '9', so the distortion value of 0 shouldn't take part in the distortion computation)


BUB97:
@googlebot I signed it!

PR #1141, opened 2020-01-02
[MRG] CWL2 in PyTorch

michaelshiyu:
Implementing [the Carlini-Wagner attack (L2)](https://arxiv.org/abs/1608.04644) in PyTorch.

All related test cases from the existing TF interface have been implemented in PyTorch in this PR and passed by this implementation of CWL2.

Please review and let me know what needs to be modified.

Thanks :)

### DIFF
--- a/cleverhans/future/torch/attacks/carlini_wagner_l2.py
+++ b/cleverhans/future/torch/attacks/carlini_wagner_l2.py
@@ -123,8 +123,7 @@ def carlini_wagner_l2(model_fn, x, n_classes,
 
   # Prepare some variables
   modifier = torch.zeros_like(x, requires_grad=True)
-  y_onehot = y.new_zeros((len(y), n_classes), dtype=torch.float)
-  y_onehot[range(len(y)), y] = 1.
+  y_onehot = torch.nn.functional.one_hot(y, n_classes).to(torch.float)
 
   # Define loss functions and optimizer
   f_fn = lambda real, other, targeted: torch.max(
--- a/cleverhans/future/torch/attacks/carlini_wagner_l2.py
+++ b/cleverhans/future/torch/attacks/carlini_wagner_l2.py
@@ -25,8 +25,11 @@ def carlini_wagner_l2(model_fn, x, n_classes,
   as this attack is often much slower than others.
 
   :param model_fn: a callable that takes an input tensor and returns
-            the model logits.
-  :param x: input tensor.
+            the model logits. The logits should be a tensor of shape
+            (n_examples, n_classes).
+  :param x: input tensor of shape (n_examples, ...), where ... can
+            be any arbitrary dimension that is compatible with
+            model_fn.
   :param n_classes: the number of classes.
   :param y: (optional) Tensor with true labels. If targeted is true,
             then provide the target label. Otherwise, only provide
@@ -34,7 +37,9 @@ def carlini_wagner_l2(model_fn, x, n_classes,
             crafting adversarial samples. Otherwise, model predictions
             are used as labels to avoid the "label leaking" effect
             (explained in this paper:
-            https://arxiv.org/abs/1611.01236). Default is None.
+            https://arxiv.org/abs/1611.01236). If provide y, it
+            should be a 1D tensor of shape (n_examples, ).
+            Default is None.
   :param targeted: (optional) bool. Is the attack targeted or
             untargeted? Untargeted, the default, will try to make the
             label incorrect. Targeted will instead try to move in the
@@ -83,13 +88,37 @@ def carlini_wagner_l2(model_fn, x, n_classes,
       pred_copy[label] += confidence * (-1) ** int(targeted)
       pred = torch.argmax(pred_copy)
 
-    return pred == label if targeted else pred !=label
+    return pred == label if targeted else pred != label
 
 
   if y is None:
     # Using model predictions as ground truth to avoid label leaking
     pred = model_fn(x)
-    _, y = torch.max(pred, 1)
+    y = torch.argmax(pred, 1)
+
+  # Initialize some values needed for binary search on const
+  lower_bound = [0.] * len(x)
+  upper_bound = [1e10] * len(x)
+  const = x.new_ones(len(x), 1) * initial_const
+
+  o_bestl2 = [1e10] * len(x)
+  o_bestscore = [-1.] * len(x)
+  x = torch.clamp(x, clip_min, clip_max)
+  ox = x.clone().detach() # save the original x
+  o_bestattack = x.clone().detach()
+
+  # Map images into the tanh-space
+  # TODO as of 01/06/2020, PyTorch does not natively support
+  # arctanh (see, e.g.,
+  # https://github.com/pytorch/pytorch/issues/10324).
+  # This particular implementation here is not numerically
+  # stable and should be substituted w/ PyTorch's native
+  # implementation when it comes out in the future
+  arctanh = lambda x: 0.5 * torch.log((1 + x) / (1 - x))
+  x = (x - clip_min) / (clip_max - clip_min)
+  x = torch.clamp(x, 0, 1)
+  x = x * 2 - 1
+  x = arctanh(x * .999999)
 
   # Prepare some variables
   modifier = torch.zeros_like(x, requires_grad=True)
@@ -103,15 +132,6 @@ def carlini_wagner_l2(model_fn, x, n_classes,
   )
   l2dist_fn = lambda x, y: torch.pow(x - y, 2).sum(list(range(len(x.size())))[1:])
   optimizer = torch.optim.Adam([modifier], lr=lr)
-
-  # Initialize some values needed for binary search on const
-  lower_bound = [0.] * len(x)
-  upper_bound = [1e10] * len(x)
-  const = x.new_ones(len(x), 1) * initial_const
-
-  o_bestl2 = [1e10] * len(x)
-  o_bestscore = [-1.] * len(x)
-  o_bestattack = x.clone().detach()
 
   # Outer loop performing binary search on const
   for outer_step in range(binary_search_steps):
@@ -131,7 +151,7 @@ def carlini_wagner_l2(model_fn, x, n_classes,
 
       optimizer.zero_grad()
       f = f_fn(real, other, targeted)
-      l2 = l2dist_fn(new_x, x)
+      l2 = l2dist_fn(new_x, ox)
       loss = (const * f + l2).sum()
       loss.backward()
       optimizer.step()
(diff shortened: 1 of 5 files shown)


cleverhans/future/torch/attacks/carlini_wagner_l2.py
```diff
@@ -0,0 +1,204 @@
...
+            class (int).
+    :param label: int. A label to compare against.
+    :param is_logits: (optional) bool. If True, treat pred as an
+            array of logits. Default is False.
+    """
+
+    # Convert logits to predicted class if necessary
+    if is_logits:
+      pred_copy = pred.clone().detach()
+      pred_copy[label] += confidence * (-1) ** int(targeted)
```
carlini: This is a clever trick but it might be more clear to say `confidence if targeted else -confidence`

michaelshiyu: Ah yes, of course. Will revise. Thanks.

cleverhans/future/torch/attacks/carlini_wagner_l2.py
```diff
@@ -0,0 +1,204 @@
...
+  x = arctanh(x * .999999)
+
+  # Prepare some variables
+  modifier = torch.zeros_like(x, requires_grad=True)
+  y_onehot = y.new_zeros((len(y), n_classes), dtype=torch.float)
+  y_onehot[range(len(y)), y] = 1.
+
+  # Define loss functions and optimizer
+  f_fn = lambda real, other, targeted: torch.max(
+    (-1) ** int(targeted) * (real - other) + confidence,
```
carlini: Maybe same change here to make it more clear what's going on.

cleverhans/future/torch/attacks/carlini_wagner_l2.py
```diff
@@ -0,0 +1,204 @@
...
+    torch.tensor(0.)
+  )
+  l2dist_fn = lambda x, y: torch.pow(x - y, 2).sum(list(range(len(x.size())))[1:])
+  optimizer = torch.optim.Adam([modifier], lr=lr)
+
+  # Outer loop performing binary search on const
+  for outer_step in range(binary_search_steps):
+    # Initialize some values needed for the inner loop
+    bestl2 = [1e10] * len(x)
```
carlini: I know this is what my code does, but it would be better to set this to `np.inf`.

cleverhans/future/torch/attacks/carlini_wagner_l2.py
michaelshiyu: Will do. Thanks.

cleverhans/future/torch/attacks/carlini_wagner_l2.py
michaelshiyu: Modified ```1e10``` to ```float('inf')``` instead since ```numpy``` was not imported anywhere in this module (so importing it for this one-time use might be unjustifiably expensive if this is the first time ```numpy``` is imported anywhere in the running process) and ```float('inf')``` does essentially the same thing here. Thanks.

michaelshiyu:
@carlini Integrated the proposed changes. Please review at your earliest convenience. Thanks :)

cleverhans/future/torch/attacks/carlini_wagner_l2.py
```diff
@@ -0,0 +1,205 @@
...
+  # implementation when it comes out in the future
+  arctanh = lambda x: 0.5 * torch.log((1 + x) / (1 - x))
+  x = (x - clip_min) / (clip_max - clip_min)
+  x = torch.clamp(x, 0, 1)
+  x = x * 2 - 1
+  x = arctanh(x * .999999)
+
+  # Prepare some variables
+  modifier = torch.zeros_like(x, requires_grad=True)
+  y_onehot = y.new_zeros((len(y), n_classes), dtype=torch.float)
```
iamgroot42, line 126: Using `torch.nn.functional.one_hot(y, n_classes)` might be a more elegant solution

cleverhans/future/torch/attacks/carlini_wagner_l2.py
```diff
@@ -0,0 +1,205 @@
+"""The CarliniWagnerL2 attack."""
+import torch
+
+
+INF = float('inf')
```
iamgroot42, line 5: Looks awkward: maybe use np.inf directly in code ? (I read your thread regarding not using numpy import, but it doesn't matter since other attacks are using it and the __init_.py will lead to it being imported anyway. The only case where it would help is when someone directly imports this attack)

michaelshiyu:
> Is there any particular reason for not including `abort_early` functionality in the Pytorch version of the attack (it's present in the TF version)?

That's a good point. I was aiming for a streamlined implementation with only essential functionalities for max efficiency before this PR gets reviewed. If @npapernot or @carlini think that ```abort_early``` and other logging functionalities from the existing TF version is essential, let me know and either @iamgroot42 or myself can go ahead and implement it in this PyTorch version. Thanks.

cleverhans/future/torch/attacks/carlini_wagner_l2.py
```diff
@@ -0,0 +1,205 @@
...
+      loss = (const * f + l2).sum()
+      loss.backward()
+      optimizer.step()
+
+      # Update best results
+      for n, (l2_n, logits_n, new_x_n) in enumerate(zip(l2, logits, new_x)):
+        y_n = y[n]
+        if l2_n < bestl2[n] and compare(logits_n, y_n, is_logits=True):
+          bestl2[n] = l2_n
+          bestscore[n] = torch.argmax(logits_n)
```
iamgroot42, line 165: Just a suggestion (not sure how much faster it would be):  Maybe precompute and store the predictions corresponding to `logits_n`, instead of doing argmax calls individually for each batch element (since it's presumably on the GPU, it could benefit from speedups)

cleverhans/future/torch/attacks/carlini_wagner_l2.py
michaelshiyu: Good point. Thanks!

cleverhans/future/torch/attacks/carlini_wagner_l2.py
michaelshiyu: Pre-computing ```torch.argmax(logits_n)``` outside this ```for``` loop probably won't help since ```torch.argmax(logits_n)``` may or may not be called for each ```n``` so pre-computing for all ```n``` might be a waste.

On the other hand, since ```l2_n < o_bestl2[n]``` being true implies ```l2_n < bestl2[n]``` being true, it makes sense to streamline the implementation there. A new commit has been submitted. Please let me know if it LGTY. Thanks.

cleverhans/future/torch/attacks/carlini_wagner_l2.py
michaelshiyu: "Looking awkward" is probably a subjective thing ;) Also I don't think importing all attacks would necessarily import ```numpy``` unless an attack that uses ```numpy``` has been executed.

```
Python 3.7.5 (default, Oct 25 2019, 10:52:18)
[Clang 4.0.1 (tags/RELEASE_401/final)] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import cleverhans.future.torch.attacks as attacks
>>> np
  File "<stdin>", line 1, in <module>
NameError: name 'np' is not defined
>>> attacks.projected_gradient_descent
<function projected_gradient_descent at 0x10eb40440>
```

But as I said in a separate thread. This is more of a coding style issue instead of anything that matters in terms of performance. I will let @npapernot or @carlini decide :)

cleverhans/future/torch/attacks/carlini_wagner_l2.py
```diff
@@ -0,0 +1,205 @@
+"""The CarliniWagnerL2 attack."""
+import torch
+
+
+INF = float('inf')
+def carlini_wagner_l2(model_fn, x, n_classes,
```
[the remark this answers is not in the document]
michaelshiyu: This is a good point. I was abiding by the Zen of Python and trying to be explicit rather than implicit. But if @npapernot or @carlini think it's better to infer the number of classes from logits, I'm happy to make the change. Thanks.

cleverhans/future/torch/attacks/carlini_wagner_l2.py
iamgroot42: Fair enough (although the attached way to check if one library imports the other is incorrect)

michaelshiyu: Ah you're right. Updated my reply above. Thanks.

cleverhans/future/torch/attacks/carlini_wagner_l2.py
iamgroot42: Yeah, that's a fair point!

iamgroot42:
@michaelshiyu it looks like the `__init__.py` file inside the `torch/attacks/` folder does not contain a reference to `carlini_wagner_l2` (like the rest of attacks do)?

cleverhans/future/torch/attacks/carlini_wagner_l2.py
```diff
@@ -0,0 +1,209 @@
...
+  x = arctanh(x * .999999)
+
+  # Prepare some variables
+  modifier = torch.zeros_like(x, requires_grad=True)
+  y_onehot = torch.nn.functional.one_hot(y, n_classes).to(torch.float)
+
+  # Define loss functions and optimizer
+  f_fn = lambda real, other, targeted: torch.max(
+    ((other - real) if targeted else (real - other)) + confidence,
+    torch.tensor(0.)
```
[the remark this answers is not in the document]
michaelshiyu: Good catch! Fixed with d6417be. Thanks.

cleverhans/future/torch/attacks/carlini_wagner_l2.py
```diff
@@ -0,0 +1,209 @@
...
+      else:
+        # Failure, either multiply by 10 if no solution found yet
+        # or do binary search with the known upper bound
+        lower_bound[n] = max(lower_bound[n], const[n])
+        if upper_bound[n] < 1e9:
+          const[n] = (lower_bound[n] + upper_bound[n]) / 2
+        else:
+          const[n] *= 10
+
+  return o_bestattack
```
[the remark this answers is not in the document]
michaelshiyu: The ```requires_grad``` flag is not so much of an issue since it does not really affect anything but I do agree that the returned tensor carrying with it its entire backward graph might cause memory leaks. Fixed with d6417be. Thanks.

jonasguan:
Lgtm, thanks all for the hard work!

michaelshiyu:
@jonasguan Thank you for the review!

PR #1146, opened 2020-02-08
Request to Add Pixel Attack and Threshold Attack

shashankkotyan:
I would like to add an implementation of Few Pixel Attack (https://arxiv.org/pdf/1906.06026.pdf) which is a generalisation of One Pixel Attack (https://arxiv.org/pdf/1710.08864.pdf) and Threshold Attack (https://arxiv.org/pdf/1906.06026.pdf).

- Added Implementation of Pixel Attack and Threshold Attack. 
- Updated init file.
- Requires pycma (https://github.com/CMA-ES/pycma) library.

### DIFF
--- a/cleverhans/attacks/__init__.py
+++ b/cleverhans/attacks/__init__.py
@@ -37,5 +37,8 @@ from cleverhans.compat import softmax_cross_entropy_with_logits
 from cleverhans.utils_tf import clip_eta
 from cleverhans import utils_tf
 
+from cleverhans.attacks.pixel_threshold import PixelAttack
+from cleverhans.attacks.pixel_threshold import ThresholdAttack
+
 _logger = utils.create_logger("cleverhans.attacks")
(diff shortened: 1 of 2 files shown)


shashankkotyan:
@googlebot I signed it!

jonasguan:
Hi @shashankkotyan , thanks for the PR. Unfortunately we're undergoing some major updates and have dropped support for the current version using TF1 in favor for TF2, PyTorch and Jax, so I'll have to close this for now. However, if you would like to open a new PR for the upcoming new library version, we'd much welcome your contribution.

A few comments/questions, in case you would like to resubmit: Is the paper for the Few Pixel Attack peer-reviewed anywhere? Is it possible to implement (without too much trouble) without adding the pycma library as a dependency?
Also, we'd prefer it if you can fill in the TODO comments, since it's usually best to have the developer who wrote the code add the comments to ensure correctness.

PR #1148, opened 2020-02-26
Implement HopSkipJump in PyTorch 

iamgroot42:
Tested attack locally (in both modes) with some models of my own as well.
Ported tests from main TF version as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,7 +297,7 @@ The following authors contributed 100 lines or more (ordered according to the Gi
 * Aashish Kumar (HARMAN X)
 * Aurko Roy (Google Brain)
 * Alexander Matyasko (Nanyang Technological University)
-* Anshuman Suri (Microsoft)
+* Anshuman Suri (University of Virginia)
 * Yen-Chen Lin (MIT)
 * Vahid Behzadan (Kansas State)
 * Jonathan Uesato (DeepMind)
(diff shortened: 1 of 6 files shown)


Jianbo-Lab:
Great thanks for taking your time to implement HopSkipJumpAttack in PyTorch, @iamgroot42 ! 
I have read through the script and there remains one question to this line: Line 287: _, idx = torch.min(dists, 0)
Suppose in the extreme case when several values in dists are equal. Then idx will become a k-hot vector, with k boolean Trues. That will cause some unexpected behaviors in Line 290: out_image = out_images[idx].unsqueeze(0). 
Of course this only affects grid search which does not the default choice of HSJA.

iamgroot42:
@Jianbo-Lab thanks for your feedback! I see what you mean here, but I think Pytorch handles equal values in torch.min() the same way numpy does. So even if some values are the same, it won't be a k-hot vector. You can try it yourself in Pytorch and see:

image

Jianbo-Lab:
@iamgroot42 Thanks for the clarification! BTW, feel free to @ me when future changes are made to HSJA, so that I can review it without delay. 

I have gone through the code and think it is ready to merge @npapernot .

iamgroot42:
Messed up the commits in here. Opening another PR for this attack

PR #1149, opened 2020-03-17
[TF2] PGD attack

CNOCycle:
1. support L-2 and L-inf in PGD attack
2. support custom loss
3. fix for rand_minmax issue
4. fix for default argument value

### DIFF
--- a/cleverhans/future/tf2/attacks/fast_gradient_method.py
+++ b/cleverhans/future/tf2/attacks/fast_gradient_method.py
@@ -110,7 +110,12 @@ def optimize_linear(grad, eps, norm=np.inf):
   axis = list(range(1, len(grad.get_shape())))
   avoid_zero_div = 1e-12
   if norm == np.inf:
+    # Take sign of gradient
     optimal_perturbation = tf.sign(grad)
+    # The following line should not change the numerical results. It applies only because
+    # `optimal_perturbation` is the output of a `sign` op, which has zero derivative anyway.
+    # It should not be applied for the other norms, where the perturbation has a non-zero derivative.
+    optimal_perturbation = tf.stop_gradient(optimal_perturbation)
   elif norm == 1:
     abs_grad = tf.abs(grad)
     sign = tf.sign(grad)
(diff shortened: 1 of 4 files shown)


cleverhans/future/tf2/attacks/fast_gradient_method.py
```diff
@@ -42,7 +47,7 @@ def fast_gradient_method(model_fn, x, eps, norm, clip_min=None, clip_max=None, y
     # Using model predictions as ground truth to avoid label leaking
     y = tf.argmax(model_fn(x), 1)
 
-  grad = compute_gradient(model_fn, x, y, targeted)
+  grad = compute_gradient(model_fn, loss_fn, x, y, targeted)
```
alkaet, line 50: Loss function should come as an optional parameter at the end to avoid breaking code calling compute_gradient directly

CNOCycle: In my opinion, `loss_fn` is essential argument for private function `compute_gradient`. and `loss_fn` is optional for public function `fast_gradient_method`. 

we have selected a default `loss_fn` at line 34-35 and put `loss_fn` to function compute_gradient.

cleverhans/future/tf2/utils_tf.py
```diff
@@ -31,3 +31,80 @@ def clip_eta(eta, norm, eps):
...
+  
+  shape: shape of the sampled tensor.
+  :rate: (optional) rate parameter of the exponential distribution, defaults to 1.0.
+  :dtype: (optional) data type of the sempled tensor, defaults to tf.float32.
+  :seed: (optional) custom seed to be used for sampling.
+  """
+  return tf.random.gamma(shape, alpha=1, beta=1. / rate, dtype=dtype, seed=seed)
+
+
+def random_laplace(shape, loc=0.0, scale=1.0, dtype=tf.float32, seed=None):
```
dhalf, line 49: I've fixed the implementation of the Laplace distribution: it should be the difference of two iid. random exponential vectors such that `rate = 1/scale`, shifted by the mean `loc`

CNOCycle: The original code is from here:
https://github.com/cleverhans-lab/cleverhans/blob/8c095c9a65b4d2dacffff6676a565ee5ec269875/cleverhans/utils_tf.py#L451-L458

Since the most popular attacks are on L-inf norm, I did not check this function's the correctness on L-1 and L-2 norm.

But I think that you're right.

PR #1163, opened 2020-05-14
[MRG] Sparse L1 Descent for PyTorch

iamgroot42:
Implementing the Sparse L-1 Descent Algorithm in PyTorch.


iamgroot42:
Implementing the Sparse L-1 Descent Algorithm in PyTorch, ported from the TF version.
All tests ported from TF version.

nick-jia:
Hello @iamgroot42, thanks a lot for yout contribution and sorry for the late reply. I read your codes and it looks great! Just a few things I want to discuss with you:

1. Regarding sanity check, do you mind including the assertion at line 51 of `cleverhans/future/torch/attacks/sparse_l1_descent.py` in the asserts list, and set default `sanity_checks` to False?
2. I would suggest add `grad_sparsity = torch.tensor(grad_sparsity)` after line 61, since `grad_sparsity` may not have attribute 'shape'
3. Comparing to the tf version (see line 495 of `cleverhans/utils_tf.py`), the random initialization at line 88 of `cleverhans/future/torch/attacks/sparse_l1_descent.py` seems to miss the scaling step.

Thank you again!

PR #1165, opened 2020-05-26
Fixed calculation of alternating sign of the jacobian augmentation update step

gilfidel:
This fixes a bug in the calculation introduced in [4774fe5324742ffa915ddf676992a2f246e6ed26](https://github.com/tensorflow/cleverhans/commit/4774fe5324742ffa915ddf676992a2f246e6ed26)
that caused the sign to be negative during the first 3 jbda epochs and positive afterwards instead of alternating every 3 epochs.


gilfidel:
@googlebot I signed it!

jonasguan:
@npapernot Can you take a quick look at this?

npapernot (approved):

Issue #1167, opened 2020-06-08
[TF2] TF2 implementation of FGM has the loss function hard coded


Joool:
**Describe the bug**
In contrast to the tf 1 version: 
https://github.com/tensorflow/cleverhans/blob/0bea6b83bb0c55e0843d95f49bf234b96c0bacb1/cleverhans/attacks/fast_gradient_method.py#L65-L75

The implementation of FGM in TF 2 has the loss function fixed in the source code:
https://github.com/tensorflow/cleverhans/blob/0bea6b83bb0c55e0843d95f49bf234b96c0bacb1/cleverhans/future/tf2/attacks/fast_gradient_method.py#L76-L78

Thus experimenting with differnt loss functions/configurations requires you to fork the library.

madarax64:
Hello @Joool ,
This looks to be fixed in the current release?

Joool:
Hi @madarax64,
Yes this is fixed in the current version, I'm closing this and the related pull request.

PR #1168, opened 2020-06-08
[TF2] Change FGM implementation to take different loss functions as an argument.

Joool:
This pull request addresses #1167. It changes the FGM implementation as well as all other attacks which are based on it to take different loss functions as optional arguments.


Joool:
Already addressed in the current verison.

PR #1175, opened 2020-06-25
Fix grad normalisation for PGD-L2 attack

aam-at:
When the gradient is small, we need to safeguard against division by 0. The previous implementation fails to perform the normalisation correctly when the gradient norm is less than 10^-6. This commit fixes it.

Example of the attack results for "plain" model from [mnist_challenge](https://github.com/MadryLab/mnist_challenge):
Attack parameters: PGD-L2 with 100 steps, epsilon 2.0, epsilon step 0.2:
Before this commit, robust accuracy at 2.0: 26.1%
After this commit, robust accuracy at 2.0: 1.7%


jonasguan:
Hi @aam-at , it looks like you've included some irrelevant commits in your PR; did you mean to only submit your last commit? And regarding your last commit, it looks to me that the code is quite similar, could you explain in more detail how this fixes grad normalization?

I'm going to close this PR for now since we're discontinuing support for TF1 and the current version of the library, but we're looking to launch a new version with support for TF2, PyTorch and Jax soon. If you still see this same issue in the new version, we very much welcome you to submit a new PR.

PR #1176, opened 2020-07-09
Update spsa.py

gamaleldin:
Change the dependency of spsa.py on the deprecated tf.contrib


mchan133:
Hi, the addition of tensorflow_addons (TF2 only) seems to conflict with CleverHans 3.1 being supported on TF1.

Please correct me if I'm missing something.

PR #1181, opened 2020-10-12
Update spsa.py

rsandler00:
Currently, although SPSA attack theoretically works with both differentiable and non-differentiable classifiers, its use with the latter is limited since it expects a TF tensor output. Thus it cannot be tested w/ non-TF classifiers (e.g. random forests, scipy, etc...) which output Numpy arrays instead of TF tensors. This is because line 49 uses logits.get_shape().

To fix this, I propose to sub logits.get_shape to logits.shape since this would generalize to both Numpy arrays and TF tensors.


rsandler00:
@googlebot I signed it!

PR #1187, opened 2021-01-13
Fix adversarial patch notebook

amrzv:
Hi!
Updated AdversarialPatch.ipynb to python3. Also
- removed unnecessary and duplicate imports;
- specified tf version
- replaced `math.` to `np.`
- replaced `xrange` to `range`
- used `pickle` to save and load obj's and patch's
- added colab badge and link to notebook from repo to readme
- prettified a little

P.S: tested in colab (run all)
P.S.S: all outputs left the same

### DIFF
--- a/examples/adversarial_patch/AdversarialPatch.ipynb
+++ b/examples/adversarial_patch/AdversarialPatch.ipynb
@@ -16,8 +16,8 @@
       ]
     },
     "kernelspec": {
-      "name": "python2",
-      "display_name": "Python 2"
+      "name": "python3",
+      "display_name": "Python 3"
     },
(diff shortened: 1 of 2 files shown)


jonasguan:
Hi @amrzv , thanks for the PR. Unfortunately we're currently undergoing some major updates, and have discontinued support for our TF1 implementations and the current examples that use them in the `examples/` directory. As such, we currently don't have the bandwidth to review changes to older code, and I'll have to close to PR for now. But if you would like to contribute an example after we launch CleverHans v4 in a few days, we would very much welcome it.

amrzv:
Okay, will be staying tuned.

PR #1194, opened 2021-03-01
Changed code formatting to Black

jonasguan:
Don't think I broke anything...


tejuafonja (approved):
Looks like we're in great shape ⬛  ;)

jonasguan:
Thanks for the review @tejuafonja !

PR #1196, opened 2021-03-03
Black formatter

alkaet:
Resolved merge conflict and applied black to the two functions were it was not applied


jonasguan:
Thanks @alkaet :)

PR #1207, opened 2021-04-02
add resampling audio input transformation defense in torch

saumya0303:
Discarding samples from a waveform during downsampling could remove a significant portion of the adversarial perturbation, thereby prevents an adversarial attack.

### DIFF
--- a/defenses/torch/audio/input_tranformation/resampling.py
+++ b/defenses/torch/audio/input_tranformation/resampling.py
@@ -1,18 +1,27 @@
 import torchaudio
 import librosa
 
+
 # resampling reference https://core.ac.uk/download/pdf/228298313.pdf
 # resampling input transformation defense for audio
 
 T = torchaudio.transforms
-audio_data = librosa.load(files, sr=16000)[0][-19456:]  # Read audio file
+
+# Read audio file
+audio_data = librosa.load(files, sr=16000)[0][-19456:]
+
 audio_data = torch.tensor(audio_data).float().to(device)
-sample = T.Resample(
-    16000, 8000, resampling_method="sinc_interpolation"
-)  # resample the audio files to 8kHz from 16kHz
+
+# Discarding samples from a waveform during downsampling could remove a significant portion of the adversarial perturbation,
+# thereby prevents an adversarial attack.
+
+# resample the audio files to 8kHz from 16kHz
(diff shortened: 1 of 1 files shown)


alkaet:
Hi
Thanks for the pull request!
Could you comment a little on this pull request purpose?
As far as I know such downsampling followed by an upsampling either introduces aliasing or deletes information from the signal in which case the upsampling is not really needed. Maybe I am missing something?
@dhalf An opinion on this topic?

saumya0303 (commented):
Discarding samples from a waveform during down-sampling
could remove a significant portion of the adversarial perturbation, thereby prevents an adversarial attack.

alkaet:
Thanks for the response.
I guess I'm still a bit puzzled by this defence from a signal perspective. To put it bluntly, what is the point of upsampling back since you already lost the extra signal quality of wideband by downsampling to narrowband? Why would you not directly sample the signal at 8kHz as the defender?

saumya0303:
I agree with your doubt and will be thankful if someone helps us out with the reason. I have followed what was mentioned in the paper [https://core.ac.uk/download/pdf/228298313.pdf](url) for re-sampling defense and implemented the same.

alkaet:
Thanks!
I was not aware of this specific work but have seen resampling in other papers. Afaik these other works did not justify resampling against the points I made above either.

I'll check with other maintainers, get consensus and come back to you?

alkaet:
I checked with the other maintainers.
We'll accept the pull request but it would be nice to have a comment outlining the potential signal limitations of the defense.
Can you format the code to black and add a comment to that effect?
Thanks a lot for the contribution

saumya0303:
I have formatted the code to black using this [link](github.com/psf/black). Please check if it is the appropriate way to write the code in this format. I have added the comment to elaborate why using this defence.  Sorry I couldn't get what comment you are expecting to add. Can you please elaborate it further?

alkaet:
Thanks!
Can you add in the warnings that the defence is disputed from a signal point of view?

alkaet:
I ment the fact that this defence may yield aliasing and that it is not obvious why you would downsample then upsample rather than just use the narrowband sample rate.

alkaet:
Thanks for the added limitation.
These issues are linked to the Shannon sampling theorem and the fact that lost information cannot be recreated.

PR #1212, opened 2021-05-31
use torch builtin arctanh method

redwrasse:
use torch builtin arctanh method in torch implementation of carlini-wagner_l2


jonasguan:
Lgtm, thanks @redwrasse !

PR #1222, opened 2021-11-17
Make loss function a parameter in torch implementation

gkysaad:
Implemented fix to this issue: https://github.com/cleverhans-lab/cleverhans/issues/1211.
The `loss_fn` used must only take 2 parameters (output and target).


kylematoba:
This seems great. It'd help me if you could also make the same change in this call: https://github.com/cleverhans-lab/cleverhans/blob/master/cleverhans/torch/attacks/projected_gradient_descent.py#L123.

